PR #11, opened 2020-01-07
Fix non standard input for Cvv

ioan-ghisoi-cko:
## Proposed changes

Change "textContentType" to ".creditCardNumber" to avoid non-standar number keyboard input for the CVV field.

## Types of changes

### DIFF
--- a/Source/CvvInputView.swift
+++ b/Source/CvvInputView.swift
@@ -30,7 +30,7 @@ import UIKit
 
     private func setup() {
         textField.keyboardType = .numberPad
-        textField.textContentType = nil
+        textField.textContentType = .creditCardNumber
         textField.delegate = self
         textField.addTarget(self, action: #selector(textFieldDidChange), for: UIControl.Event.editingChanged)
     }


Source/CvvInputView.swift
```diff
@@ -30,7 +30,7 @@ import UIKit
 
     private func setup() {
         textField.keyboardType = .numberPad
-        textField.textContentType = nil
+        textField.textContentType = .creditCardNumber
```
deepesh-vasthimal-cko: @ioan-ghisoi-cko , Just one quick question do these bug fixes directly go inside master ? do we not have release branches ? or we use tags for different SDK releases ?

daven-gomes-cko:
The `.creditCardNumber` type will prompt iOS to suggest the user's saved credit card number for the CVV field - this could be changed to `.asciiCapableNumberPad`.  A new PR was created here with the fix https://github.com/checkout/frames-ios/pull/12

deepesh-vasthimal-cko:
@ioan-ghisoi-cko Closing this PR as a new PR : https://github.com/checkout/frames-ios/pull/12 is now merged.

PR #21, opened 2020-09-08
Fixed issues when 'createCardToken' and 'createApplePayToken' request…

christianvershkov:
Fixed issues when 'createCardToken' and 'createApplePayToken' requests could return nothing if an unexpected error occurs.
Also, there is an ability now to expand NetworkError enum in order to handle more cases


daven-gomes-cko:
Thanks for your contribution @christianvershkov 

Could you please keep the original method and add a new method with the new `NetworkError` type? Something like:

```
public func createCardToken(card: CkoCardTokenRequest,
                                successHandler: @escaping (CkoCardTokenResponse) -> Void,
                                errorHandler: @escaping (ErrorResponse) -> Void) {
// original implementation
}

public func createCardToken(card: CkoCardTokenRequest,
                                successHandler: @escaping (CkoCardTokenResponse) -> Void,
                                errorHandler: @escaping (NetworkError) -> Void) {
// same implementation as above, but using the new `NetworkError` type
}
```

This will mean there is no breaking change when we release this, and when we release a new major version of the SDK, we will remove the old method.

Also, there seems to be a [failing test](https://travis-ci.org/github/checkout/frames-ios/builds/725269149) on this PR:
```
Value of type 'NetworkError' has no member 'errorCodes'
	Testing cancelled because the build failed.
```

daven-gomes-cko:
Perhaps you could even mark the old method as deprecated:
`@available(*, deprecated, message: "Use createCardToken with a NetworkError instead.")`

daven-gomes-cko:
Hi @christianvershkov, may I ask if you are still working on this PR? 🙂

christianvershkov:
@daven-gomes-cko  sorry for the late reply. Thank you for bringing it up. I will have finished it by tomorrow evening (I will do my best to have it done today)

christianvershkov:
@daven-gomes-cko  I little bit updated my approach and added usage of Swift Result type (looks really better)

daven-gomes-cko (approved):

daven-gomes-cko:
Thanks @christianvershkov for your contribution! We've merged this into `develop` and it will be included in the next release.

christianvershkov:
Hi, @daven-gomes-cko 
Are there any updates on releasing this functionality?

deepesh-vasthimal-cko:
Hi, @christianvershkov 
we are not planning to make a release this month. However the fix is merged into develop branch if you like to use it as a workaround. 
Thanks

Issue #47, opened 2021-03-23
CkoPhoneNumber coding key query


adamwaite:
Hi.

Just wanted to query something I've noticed in the API contracts for registering a card in the payment flows (`POST https://api.checkout.com/tokens`).

The `CkoAddress` type (and other associated types, like `CkoCardTokenRequest`) is encoded using snake case keys (https://github.com/checkout/frames-ios/blob/master/Source/CkoAddress.swift), as can be seen in the `Encodable` implementation. `addressLine1` becomes `address_line1` for example.

However `CkoPhoneNumber` is not encoded with any custom `CodingKeys` (https://github.com/checkout/frames-ios/blob/master/Source/CkoPhoneNumber.swift) and will therefore take the standard swift case convention e.g. `countryCode` remains as `countryCode`.

Is this a known quirk on the API? Or maybe a bug in this SDK?

deepesh-vasthimal-cko:
@adamwaite Thanks for raising this with us. We are aware of this now and we will plan to make a fix in the SDK with our next release. Thanks

deepesh-vasthimal-cko:
@adamwaite , This is now fixed as part of our latest release 3.4.0. Please feel free to re-open this if you have any further questions.

adamwaite:
Hey CkoAddress still implements `CodingKeys`, is that an edge case?

adamwaite:
Also `CkoCardTokenRequest`

daven-gomes-cko:
Hey @adamwaite, the update was a quick fix that just addressed the Codable objects that were causing errors. We'll work on improving the whole decoding/encoding logic for a future update. 🙂

PR #48, opened 2021-03-24
Feature/pimob 580 coding key query

daven-gomes-cko:
## Proposed changes

Uses the snake case decoding strategy for JSONDecoder. Addresses #47 

## Types of changes

What types of changes does your code introduce to Frames Ios? Addresses #47.

## Checklist

### DIFF
--- a/Source/CheckoutAPIClient.swift
+++ b/Source/CheckoutAPIClient.swift
@@ -18,16 +18,18 @@ public class CheckoutAPIClient {
                 "Content-Type": "application/json"]
     }
 
-    private var jsonEncoder: JSONEncoder {
-        let jsonEncoder = JSONEncoder()
-        jsonEncoder.keyEncodingStrategy = .convertToSnakeCase
-        return jsonEncoder
-    }
+    private let jsonEncoder: JSONEncoder
+    private let jsonDecoder: JSONDecoder
 
-    private var jsonDecoder: JSONDecoder {
-        let jsonDecoder = JSONDecoder()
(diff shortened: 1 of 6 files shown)


deepesh-vasthimal-cko (approved):

Source/CheckoutAPIClient.swift
```diff
@@ -134,7 +152,13 @@ public class CheckoutAPIClient {
...
 
         request(urlRequest)
-            .validate().responseJSON { response in
+            .validate().responseJSON { [weak self] response in
+
+                guard let strongSelf = self else {
+
+                    completion(.failure(.objectDeallocatedUnexpectedly))
+                    return
+                }
```
harry-brown-cko, line 161: just want to double check, the places where we're using weak and strong self are weak wherever possible right?

harry-brown-cko (approved):
one comment, otherwise looking good 👍

Source/CheckoutAPIClient.swift
daven-gomes-cko: Yep, I updated the "new" methods but not the deprecated methods as there is no way to exit/return a `NetworkError` (only `ErrorResponse`), e.g. 

```
@available(*, deprecated, message: "Use createApplePayToken with a Swift Result type for completion handler.")
    public func createApplePayToken(paymentData: Data,
                                    successHandler: @escaping (CkoCardTokenResponse) -> Void,
                                    errorHandler: @escaping (ErrorResponse) -> Void)
```

PR #51, opened 2021-03-26
PIMOB-572: use frames user agent

harry-brown-cko:
## Proposed changes

- Use `checkout-sdk-frames-ios/\(framesVersion)` as `User-Agent` header

## Types of changes

## Checklist

## Further comments

- None

### DIFF
--- a/Tests/Core/CheckoutAPIClientTests.swift
+++ b/Tests/Core/CheckoutAPIClientTests.swift
@@ -9,6 +9,11 @@ class CheckoutAPIClientTests: XCTestCase {
         publicKey: "pk_test_6ff46046-30af-41d9-bf58-929022d2cd14",
         environment: .sandbox)
 
+    let everythingWithCorrectUserAgent: (_ request: URLRequest) -> Bool = { request in
+        let framesVersion = Bundle(for: CheckoutAPIClient.self).infoDictionary?["CFBundleShortVersionString"] as? String ?? "unknown"
+        return request.allHTTPHeaderFields?["User-Agent"] == "checkout-sdk-frames-ios/\(framesVersion)"
+    }
+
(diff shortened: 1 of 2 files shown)


Tests/Core/CheckoutAPIClientTests.swift
```diff
@@ -9,6 +9,11 @@ class CheckoutAPIClientTests: XCTestCase {
         publicKey: "pk_test_6ff46046-30af-41d9-bf58-929022d2cd14",
         environment: .sandbox)
 
+    let everythingWithCorrectUserAgent: (_ request: URLRequest) -> Bool = { request in
+        let framesVersion = Bundle(for: CheckoutAPIClient.self).infoDictionary?["CFBundleShortVersionString"] as? String ?? "unknown"
```
deepesh-vasthimal-cko, line 13: This line ideally in test should have been picked up from CheckoutAPIClient.swift line 18 and not duplicated. Can you some how use the headers set in CheckoutAPIClient.swift to include in your unit tests ?

harry-brown-cko: I was trying to not test the internals of `CheckoutAPIClient` here and instead test the headers on the `URLRequest`s, and I don't think making the tests dependent on `CheckoutAPIClient` internals would be ideal. I could hardcode the version string here but this would require changing with every version. What do you think is the best approach here?

deepesh-vasthimal-cko: Ok makes sense for now, Probably we can refactor this bit as we go.

deepesh-vasthimal-cko (approved):

daven-gomes-cko (approved):

PR #54, opened 2021-03-30
fix: SPM resources and example project setup

harry-brown-cko:
## Proposed changes

- Resources now loading correctly through SPM
- Example SPM project now uses local version of Frames rather than version on Github
- Use version number from constants rather than from `Bundle`
    - This is because the version number is not available on the bundle with SPM

## Types of changes

## Checklist

## Further comments

None

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,8 +1,9 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.3
 import PackageDescription
 
 let package = Package(
     name: "Frames",
+    defaultLocalization: "en",
     platforms: [
         .macOS(.v10_12),
         .iOS(.v10)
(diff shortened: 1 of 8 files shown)


Package.swift
```diff
@@ -1,8 +1,9 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.3
```
harry-brown-cko, line 1: had to bump the version here to enable explicit resource declarations

Package.swift
```diff
@@ -21,13 +22,59 @@ let package = Package(
...
+                .process("Resources/arrows/keyboard-down.png"),
+                .process("Resources/schemes/icon-dinersclub@2x.png")
+            ]
         ),
-        .testTarget(
-            name: "Frames-Tests",
-            dependencies: ["Frames"],
-            path: "Tests"
-        )
+// Tests are currently broken as Mockingjay doesn't support SPM
```
harry-brown-cko, line 63: currently blocked from running tests against our SPM package as Mockingjay does not support SPM

Source/Extensions/StringExtensions.swift
```diff
@@ -4,8 +4,12 @@ import UIKit
 extension String {
 
     private func getBundle(forClass: AnyClass) -> Bundle {
+        #if SWIFT_PACKAGE
+        let baseBundle = Bundle.module
+        #else
```
harry-brown-cko, line 9: recommended bundle access in SPM

Source/UI/Views/SchemeIconsStackView.swift
```diff
@@ -17,7 +17,7 @@ class SchemeIconsStackView: UIStackView {
     func addSchemeIcon(scheme: CardScheme) {
         let imageView = UIImageView()
         imageView.setContentHuggingPriority(.defaultHigh, for: .horizontal)
-        let image = "schemes/icon-\(scheme.rawValue)".image(forClass: SchemeIconsStackView.self)
+        let image = "icon-\(scheme.rawValue)".image(forClass: SchemeIconsStackView.self)
```
harry-brown-cko, line 20: this is the fix for the images

deepesh-vasthimal-cko (approved):

PR #100, opened 2021-09-17
PI-1174: Update documentation

miles-williams-cko:


docs/Classes/AddressView.html
```diff
@@ -42,7 +42,7 @@
     </header>
 
     <p class="breadcrumbs">
-      <a class="breadcrumb" href="../index.html">FramesIos Reference</a>
```
deepesh-vasthimal-cko, line 45: Can you share the jazzy command which worked for you ? just for future reference.

miles-williams-cko: Everything is configured in `.jazzy.yaml`, so you just need to run `jazzy` 😄

harry-brown-cko (approved):

daven-gomes-cko (approved):

deepesh-vasthimal-cko (approved):

PR #101, opened 2021-09-20
PI-1285: Extract token from 3DS flow

harry-brown-cko:
## Proposed changes

- Added new callbacks for 3DS completion to pass token back to merchants
- Removed force unwrapping optionals in `ThreedsWebViewController`
- Added tests for 3DS completion callbacks

## Types of changes

## Checklist

## Further comments

Documentation will be updated as part of [feature/PI-1174-update-documentation](https://github.com/checkout/frames-ios/tree/feature/PI-1174-update-documentation)

### DIFF
--- a/Source/UI/Controllers/ThreedsWebViewController.swift
+++ b/Source/UI/Controllers/ThreedsWebViewController.swift
@@ -7,8 +7,8 @@ public class ThreedsWebViewController: UIViewController {
     // MARK: - Properties
 
     var webView: WKWebView!
-    let successUrl: String
-    let failUrl: String
+    let successUrl: URL?
+    let failUrl: URL?
 
     /// Delegate
     public weak var delegate: ThreedsWebViewControllerDelegate?
@@ -19,7 +19,15 @@ public class ThreedsWebViewController: UIViewController {
     // MARK: - Initialization
 
     /// Initializes a web view controller adapted to handle 3dsecure.
+    @available(*, deprecated, renamed: "init(successUrl:failUrl:)")
     public init(successUrl: String, failUrl: String) {
+        self.successUrl = URL(string: successUrl)
+        self.failUrl = URL(string: failUrl)
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    /// Initializes a web view controller adapted to handle 3dsecure.
+    public init(successUrl: URL, failUrl: URL) {
         self.successUrl = successUrl
         self.failUrl = failUrl
         super.init(nibName: nil, bundle: nil)
@@ -27,15 +35,15 @@ public class ThreedsWebViewController: UIViewController {
 
     /// Returns a newly initialized view controller with the nib file in the specified bundle.
     public override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Foundation.Bundle?) {
-        successUrl = ""
-        failUrl = ""
+        successUrl = nil
+        failUrl = nil
         super.init(nibName: nibNameOrNil, bundle: nibBundleOrNil)
     }
 
     /// Returns an object initialized from data in a given unarchiver.
     required public init?(coder aDecoder: NSCoder) {
-        successUrl = ""
-        failUrl = ""
+        successUrl = nil
+        failUrl = nil
         super.init(coder: aDecoder)
     }
 
@@ -89,7 +97,7 @@ extension ThreedsWebViewController: WKNavigationDelegate {
 
     private func shouldDismiss(absoluteUrl: URL) {
         // get URL conforming to RFC 1808 without the query
-        let url = absoluteUrl.withoutQuery?.absoluteString
+        let url = absoluteUrl.withoutQuery
 
         if url == successUrl {
             // success url, dismissing the page with the payment token
--- a/Source/UI/Controllers/ThreedsWebViewController.swift
+++ b/Source/UI/Controllers/ThreedsWebViewController.swift
@@ -81,14 +81,25 @@ public class ThreedsWebViewController: UIViewController,
         if url == successUrl {
             // success url, dismissing the page with the payment token
             self.dismiss(animated: true) {
+                let token = self.extractToken(from: absoluteUrl)
+                self.delegate?.threeDSWebViewControllerAuthenticationDidSucceed(self, token: token)
                 self.delegate?.onSuccess3D()
             }
         } else if url == failUrl {
             // fail url, dismissing the page
             self.dismiss(animated: true) {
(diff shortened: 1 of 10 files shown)


Source/UI/Delegates/ThreedsWebViewControllerDelegate.swift
```diff
@@ -4,8 +4,23 @@ import Foundation
...
+    /// - Parameters:
+    ///   - threeDSWebViewController: The `ThreedsWebViewController` instance calling this method.
+    ///   - token: The token extracted from the success URL.
+    func threeDSWebViewControllerAuthenticationDidSucceed(_ threeDSWebViewController: ThreedsWebViewController,
+                                                          token: String?)
+
+    /// Called upon unsuccessful 3D Secure authentication.
+    /// - Parameters:
+    ///   - threeDSWebViewController: The `ThreedsWebViewController` instance calling this method.
+    func threeDSWebViewControllerAuthenticationDidFail(_ threeDSWebViewController: ThreedsWebViewController)
```
harry-brown-cko, line 24: Bit unsure on why we're returning the threeDSWebViewController here. Should we keep it or remove it?

Source/UI/Controllers/ThreedsWebViewController.swift
```diff
@@ -53,42 +52,67 @@ public class ThreedsWebViewController: UIViewController,
...
-    // MARK: - WKNavigationDelegate
+// MARK: - WKNavigationDelegate
+extension ThreedsWebViewController: WKNavigationDelegate {
 
     /// Called when the web view begins to receive web content.
     public func webView(_ webView: WKWebView, didCommit navigation: WKNavigation!) {
-        shouldDismiss(absoluteUrl: webView.url!)
+
+        guard let absoluteUrl = webView.url else {
+            return
```
deepesh-vasthimal-cko, line 73: Is it safe to do silent returns here ?

Source/UI/Delegates/ThreedsWebViewControllerDelegate.swift
deepesh-vasthimal-cko: Good question if we are already on threeDSWebViewController we would already have its instance here. If we have a example using threeDSWebViewController then we may find a answer to this.

deepesh-vasthimal-cko (approved):
Are you running jazzy on the PR ? to get updated docs.

Source/UI/Controllers/ThreedsWebViewController.swift
harry-brown-cko: I think that a silent return is better than an app crash, which was what we had before. We're also limited here on return type as this is a WKNavigationDelegate protocol function.

Source/UI/Controllers/ThreedsWebViewController.swift
```diff
@@ -53,42 +52,67 @@ public class ThreedsWebViewController: UIViewController,
...
     /// Called when a web view receives a server redirect.
     public func webView(_ webView: WKWebView,
                         didReceiveServerRedirectForProvisionalNavigation navigation: WKNavigation!) {
         // stop the redirection
-        shouldDismiss(absoluteUrl: webView.url!)
+        guard let absoluteUrl = webView.url else {
+            return
+        }
+
+        shouldDismiss(absoluteUrl: absoluteUrl)
```
daven-gomes-cko, line 87: Do you know if this method and `func webView(_ webView: WKWebView, didCommit navigation: WKNavigation!)` can both be called for a single navigation action? E.g. if the webview tried to load a URL that falls under both delegate methods

As `shouldDismiss` would then be called twice. It's unlikely iOS would call both WK delegate methods but it just popped in my mind.

harry-brown-cko:
> Are you running jazzy on the PR ? to get updated docs.

Plan is to merge this before the documentation PR and update the docs there, as the changes to `.jazzy.yml` in that branch are required to update the docs.

Source/UI/Delegates/ThreedsWebViewControllerDelegate.swift
daven-gomes-cko: I think he included it to follow other iOS delegate methods by Apple, e.g. `UITableViewDelegate` has `func tableView(_ tableView: UITableView, ...)`, and WK delegates, etc. I think we should keep it to stick with this standard and because the integrator might create an object that conforms to `ThreedsWebViewControllerDelegate`, but doesn't actually hold a reference to `ThreedsWebViewController`. These methods would give them access to that.

Source/UI/Controllers/ThreedsWebViewController.swift
harry-brown-cko: I think they are independent of each other, but the better delegate method here would be `func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void)`. Do we want to make this change or leave it as is (as far as I can tell the behaviour would be identical)

iOS Example Frame/iOS Example Frame/MainViewController.swift
```diff
@@ -9,20 +9,24 @@
 import UIKit
 import Frames
 
-class MainViewController: UIViewController, CardViewControllerDelegate {
-    
+class MainViewController: UIViewController, CardViewControllerDelegate, ThreedsWebViewControllerDelegate {
+
     @IBOutlet weak var goToPaymentPageButton: UIButton!
     @IBOutlet weak var createTokenWithApplePay: UIButton!
+    @IBOutlet weak var threeDSURLTextField: UITextField!
```
deepesh-vasthimal-cko, line 16: 👍

deepesh-vasthimal-cko: Can we add some comments to make sure this is only for demo : threeDSURLTextField is added for demo purpose on how to use `ThreedsWebViewControllerDelegate` and its methods. In reality this url would be seen in the response from https://api-reference.checkout.com/#tag/Payments end point from your server.

Source/UI/Controllers/ThreedsWebViewController.swift
daven-gomes-cko: I guess it's not worth it as the behaviour is the same 👍

deepesh-vasthimal-cko (approved):

gravindragupta90:
Hi Team, 
When can this PR be merged?

deepesh-vasthimal-cko:
> Hi Team,
> When can this PR be merged?

The PR will be merged soon into develop, But the released is planned next week.

daven-gomes-cko (approved):

Issue #111, opened 2021-10-06
Localizable strings of the pod conflict with that of the main app when the pod is not used as a dynamic framework


aleh:
**Describe the bug**

When the Frames pod is not used as a dynamic framework (that is when `use_frameworks!` is absent in the app's `Podfile` or static frameworks are requested via`use_frameworks! :linkage => :static`), then its resources are not wrapped into a bundle and can override the corresponding resources of the app. Some names are generic enough (`checkmark.png`) but the clash is particularly easy to run into with `Localizable.strings` which is the usual name of the strings file in the app. (Actual case in our app.)

**To Reproduce**

1. Install pods for "iOS Example Frame" example (commit `71b39a9`, cocoapods 1.11.2 let say) and open its workspace in Xcode (I used 12.5, but should not matter.) Set up provisioning, so it can run on the device, don't do any other configuration. 
2. Add `Localizable.strings` into the project, click "Localize..." button, and make it available for at least English.
3. Add a single line there: `"Payment" = "Payment EN";`
4. Find `showAlert` method in `MainViewController.swift` and replace the reference to `"Payment"` string there with `NSLocalizedString("Payment", comment: "")`.
5. Run the app and trigger this alert, for example, by tapping "Start 3DS" button. 
6. Observe the title of the alert translated, i.e. being "Payment EN" now.
7. Open the app bundle (e.g. Project navigator > Products > iOS Example Frame.app > Show in Finder, then Show Package Contents on it) and observe that `en.lproj/Localizable.strings` contain the binary representation of the `.strings` file we've just added and there is no `Localizable.strings` in the root of the bundle.
8. Now open the `Podfile` and comment out `use_frameworks!`, so the pod is used as a static library and resources are copied as is into the app's bundle.
9. Install pods and run the app again.
10. Trigger the alert again.
11. Observe that the title is now simply "Payment", i.e. the translation did not work now. 
12. Open the app bundle again and observe that both `Localizable.strings` and `en.lproj/Localizable.strings` contain the strings of the pod, not the strings of the app, i.e. that the app's resource has been completely overridden.

Note that this also happens when the pod is used as a static framework, e.g. by specifying `use_frameworks! :linkage => :static` in the `Podfile` in step 8.

**Expected behavior**

[...]

PR #112, opened 2021-10-06
Fix up the .podspec to always bundle the resources

aleh:
This is to address issue #111.

The code already looks up for resources in `Frames.bundle`, so tweaking the .podspec to generate the resource bundle fixes the issue with resource clashes when the pod is linked statically.


harry-brown-cko:
I've added unit tests and opened a new PR to merge in these changes. Please see #113

PR #123, opened 2021-12-06
Bug/pi 1467 add correlation id for form presented and token flow

deepesh-vasthimal-cko:
## Proposed changes

## Types of changes

* [ x] Bugfix (non-breaking change which fixes an issue)

## Checklist

* [ x] Lint and unit tests pass locally with my changes
* [ x] I have added tests that prove my fix is effective or that my feature works
* [ x] I have added necessary documentation (if appropriate)

## Further comments

### DIFF
--- a/Source/CheckoutAPIClient.swift
+++ b/Source/CheckoutAPIClient.swift
@@ -89,12 +89,13 @@ public class CheckoutAPIClient {
             remoteProcessorMetadata: remoteProcessorMetadata)
         
         let dateProvider = DateProvider()
-        let framesEventLogger = FramesEventLogger(checkoutEventLogger: checkoutEventLogger, dateProvider: dateProvider)
+        let correlationIDGenerator = CorrelationIDGenerator()
+        let framesEventLogger = FramesEventLogger(correlationID: correlationIDGenerator.getCorrelationID(),
+                                                checkoutEventLogger: checkoutEventLogger, dateProvider: dateProvider)
         let networkFlowLoggerFactory = NetworkFlowLoggerFactory(
             framesEventLogger: framesEventLogger,
             publicKey: publicKey)
-        
-        let correlationIDGenerator = CorrelationIDGenerator()
+        print(correlationIDGenerator.getCorrelationID())
         let mainDispatcher = DispatchQueue.main
 
         self.init(publicKey: publicKey,
@@ -165,7 +166,8 @@ public class CheckoutAPIClient {
         card: CkoCardTokenRequest,
         completion: @escaping ((Swift.Result<CkoCardTokenResponse, NetworkError>) -> Void)
     ) {
-        let correlationID = correlationIDGenerator.generateCorrelationID()
+        let correlationID = correlationIDGenerator.getCorrelationID()
+        print(correlationID)
         let networkFlowLogger = networkFlowLoggerProvider.createLogger(
             correlationID: correlationID,
             tokenType: .card)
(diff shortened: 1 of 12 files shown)


Source/UI/Controllers/CardViewController.swift
```diff
@@ -108,10 +109,12 @@ public class CardViewController: UIViewController,
         registerKeyboardHandlers(notificationCenter: notificationCenter,
                                  keyboardWillShow: #selector(keyboardWillShow),
                                  keyboardWillHide: #selector(keyboardWillHide))
-
+        
         if suppressNextLog {
             suppressNextLog = false
         } else {
+            checkoutApiClient?.logger.add(metadata: (checkoutApiClient?.correlationID())!,
```
deepesh-vasthimal-cko, line 116: @daven-gomes-cko @harry-brown-cko I wanted to get rid of the forced unwrap here. But for some reason the checkoutApiClient is option on the view controller itself. Any idea why checkoutApiClient optional when creating the CardViewController.

Tests/Mocks/StubCorrelationIDGenerator.swift
```diff
@@ -1,12 +1,15 @@
...
     
     var generateCorrelationIDReturnValue: String!
     
     func generateCorrelationID() -> String {
         
         return generateCorrelationIDReturnValue
     }
-    
+
+    func destroyCorrelationID() {
```
deepesh-vasthimal-cko, line 12: Any ideas how to use this in unit tests? currently destroyCorrelationID gets only called within the success of token generation so stub object wont even reach that point.

Source/CheckoutAPIClient.swift
```diff
@@ -26,15 +26,15 @@ public class CheckoutAPIClient {
...
+         correlationIDGenerator: CorrelationIDManaging,
          logger: FramesEventLogging,
          mainDispatcher: Dispatching,
          networkFlowLoggerProvider: NetworkFlowLoggerProviding,
          requestExecutor: RequestExecuting) {
 
         self.publicKey = publicKey
         self.environment = environment
-        self.correlationIDGenerator = correlationIDGenerator
+        self.correlationIDManager = correlationIDGenerator
```
harry-brown-cko, line 37: can we change the name of the argument from `correlationIDGenerator` to `correlationIDManager` as well?

deepesh-vasthimal-cko: Makes sense. Will do.

Source/UI/Controllers/CardViewController.swift
harry-brown-cko: Because the initialisers aren't setup to guarantee that it's there 😔

Tests/Mocks/StubCorrelationIDGenerator.swift
harry-brown-cko: The stub does get called as part of `CheckoutAPIClientTests`. Anywhere where `stubRequestExecutor.executeCalledWithCompletionHandler` is called, the completion block with destroy is called. You should be able to add this check into one or all of the existing tests that call this.

daven-gomes-cko: I think the test in `StubCorrelationIDGeneratorTests` is fine, as you test that it regenerates a new one.

Here you'd add `var destroyCorrelationIDCalled = false` and set it to `true` in this method. Then you'd test that `destroyCorrelationIDCalled` is `true` in the dependant class.

deepesh-vasthimal-cko: Updated.

Source/UI/Controllers/CardViewController.swift
deepesh-vasthimal-cko: For now I have unwrapped this optional and doing a return of nil as other places in this class which is safe. Ideally we should be return a error back.

daven-gomes-cko (approved):

harry-brown-cko (approved):

PR #134, opened 2022-03-03
PI-1661: log CheckoutAPIClient init

harry-brown-cko:
## Proposed changes

- Log when `CheckoutAPIClient` is initialised
- Tests for the above

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/Core/Logging/FramesLogEvent.swift
+++ b/Source/Core/Logging/FramesLogEvent.swift
@@ -4,6 +4,7 @@ import CheckoutEventLoggerKit
 enum FramesLogEvent: Equatable {
 
     enum Property: String {
+        case environment
         case errorCodes
         case errorType
         case httpStatusCode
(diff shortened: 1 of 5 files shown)


Source/Core/Logging/FramesLogEvent.swift
```diff
@@ -63,6 +68,9 @@ enum FramesLogEvent: Equatable {
         case .paymentFormPresented,
              .billingFormPresented:
             return [:]
+        case let .checkoutAPIServiceInitialised(environment):
+            let environmentString = environment.rawValue == "live" ? "production" : environment.rawValue
```
deepesh-vasthimal-cko, line 72: Do we know why we have both live and production ? and can we sync this as just being production in future.

deepesh-vasthimal-cko (approved):
Just one comment otherwise looks good.

Source/Core/Logging/FramesLogEvent.swift
harry-brown-cko: There is currently only live in the SDK. This is specifically setting it to production for this log. When we move to 4.0 we should replace all instances of live with production

PR #144, opened 2022-04-26
PI-1738: Billing Form component: Standard

ehab-al-cko:
URL:
https://checkout.atlassian.net/browse/PIMOB-956

## Proposed changes

## Types of changes

* [x ] New feature (non-breaking change which adds functionality)

## Checklist

## Further comments


harry-brown-cko:
Base branch should be `4.0.0_RC`

harry-brown-cko:
Few notes on UI and behaviour:

- prefilled address is not populated
- scrolling form is not working when dragging from text fields (dismisses view instead)
- header is currently part of the scroll view and can be scrolled off of the screen. should the header be static?

Source/UI/NewUI/BillingForm/Style/Default Implmentation/BillingForm/Fields/Postcode/DefaultFormStylePostcodedCell.swift
```diff
@@ -0,0 +1,22 @@
+import UIKit
+
+struct DefaultFormStylePostcodedCell : BillingFormTextFieldCellStyle {
```
harry-brown-cko, line 3: typo, should be `DefaultFormStylePostcodeCell`

Source/UI/NewUI/BillingForm/Style/Cell/BillingFormTextFieldCellStyle.swift
```diff
@@ -0,0 +1,10 @@
+import UIKit
+
+// Field Cell
+public protocol BillingFormTextFieldCellStyle {
+    var backgroundColor: UIColor { get }
+    var title: InputLabelStyle? { get }
+    var hint: InputLabelStyle? { get }
+    var textfield: TextFieldStyle { get set }
+    var error: ErrorInputLabelStyle { get set }
+}
```
harry-brown-cko, line 10: Is there a way to make a field mandatory/optional? From an API point of view all of these fields are optional, Frames currently lets the merchant decide which fields to make mandatory / optional / removed.

Source/UI/NewUI/BillingForm/Style/Default Implmentation/BillingForm/Fields/AddressLine1/DefaultFormStyleAddressLine1Cell.swift
```diff
@@ -0,0 +1,23 @@
...
+    var title: InputLabelStyle?
+    var hint: InputLabelStyle?
+    var textfield: TextFieldStyle
+    var error: ErrorInputLabelStyle
+    
+    init(backgroundColor: UIColor = .white,
+         header: InputLabelStyle = DefaultTitleLabelStyle(text: "Address line 1"),
+         hint: InputLabelStyle? = nil,
+         textfield: TextFieldStyle = DefaultTextField(),
+         error: ErrorInputLabelStyle = DefaultErrorInputLabelStyle(text: "missing Address line 1")) {
```
harry-brown-cko, line 15: Can we move strings like this to Localizable.strings?

Source/UI/NewUI/BillingForm/Style/Protocols/FormButtonStyle.swift
```diff
@@ -0,0 +1,15 @@
...
+public protocol FormButtonStyle {
+    var text: String { get }
+    var font: UIFont { get }
+    var isEnabled: Bool { get set}
+    var activeTitleColor: UIColor { get }
+    var inActiveTitleColor: UIColor { get }
+    var inActiveTintColor: UIColor { get }
+    var activeTintColor: UIColor { get }
+    var backgroundColor: UIColor { get }
+    var hight: Double { get }
```
harry-brown-cko, line 13: typo, should be `height`

Source/UI/NewUI/BillingForm/Style/Default Implmentation/BillingForm/Fields/Fullname/DefaultFormStyleFullNameCell.swift
```diff
@@ -0,0 +1,22 @@
+import UIKit
+
```
harry-brown-cko, line 2: typo in folder name, I'd expect `FullName`

Source/UI/Controllers/CardViewController.swift
```diff
@@ -43,7 +43,7 @@ public class CardViewController: UIViewController,
     var topConstraint: NSLayoutConstraint?
 
     private var loggedForCurrentCorrelationID = false
-
+    public var isNewUI: Bool = false
```
harry-brown-cko, line 46: `: Bool` is not required here. Is this included for readability?

Source/UI/NewUI/BillingForm/Style/Default Implmentation/BillingForm/Fields/Fullname/DefaultFormStyleFullNameCell.swift
harry-brown-cko: also typo on the parent folder, should be `Default Implementation`

Source/UI/NewUI/BillingForm/View/Field/BillFormTextField.swift
```diff
@@ -0,0 +1,13 @@
+import UIKit
+
+final class BillFormTextField: UITextField {
```
harry-brown-cko, line 3: should this be `BillingFormTextField`?

Source/UI/NewUI/BillingForm/ViewModel/BillingFormCellType.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+@frozen enum BillingFormCellType: Int, CaseIterable {
```
harry-brown-cko, line 3: do we need to have a raw type of Int here?

harry-brown-cko (commented):
approach and testing looks good. few comments, mainly typos. a few quirks with behaviour as well

Source/UI/NewUI/BillingForm/Factory/BillingFormFactory.swift
```diff
@@ -0,0 +1,22 @@
+import UIKit
+
+struct BillingFormFactory {
+    // in order
+    static let styles: [BillingFormTextFieldCellStyle] = [
```
alex-ioja-yang-cko, line 5: Could these styles just be an enum case?

Lets imagine:
```
enum BillingFormTextFieldCellStyle {
    case fullName
    case addressLine1
    .....

    func getConfiguration() -> BilingFormTextFieldConfig {
        switch self {
        case .fullName:
            return .init(color1: .... , color2: ...)
        }
    }
}
```

My initial thought is it would reduce a lot of the syntactic sugar you currently have and may increase configuration options as well as extensibility. Additionally, you could completely drop this `styles` property if you make the enum declaration conform to `CaseIterable` and end up with a syntax of `.allCases` at call site instead of the current `styles` reference

Source/UI/NewUI/BillingForm/Factory/BillingFormFactory.swift
```diff
@@ -0,0 +1,22 @@
...
+        DefaultFormStyleAddressLine1Cell(),
+        DefaultFormStyleAddressLine2Cell(),
+        DefaultFormStyleCityCell(),
+        DefaultFormStyleStateCell(),
+        DefaultFormStylePostcodedCell(),
+        DefaultFormStyleCountryCell(),
+        DefaultFormStylePhoneNumberCell()
+    ]
+    
+    static func getBillingFormViewController(delegate: BillingFormViewModelDelegate) -> UIViewController {
```
alex-ioja-yang-cko, line 16: Personally I find the passing of a delegate at initialisation time quite uncommon. Not sure this factory provides any value, besides hiding the view model creation, which I am not sure is a common pattern?

Source/UI/NewUI/BillingForm/Style/Cell/BillingFormHeaderCellStyle.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+// Header Cell
+public protocol BillingFormHeaderCellStyle {
```
alex-ioja-yang-cko, line 4: Not certain if the comment above the protocol provides any value? The type naming should clearly convey it, and if we want the user of the syntax to flag some helper comment, we should at least use the description marker (`///`)

alex-ioja-yang-cko (commented):
A few opinions and some feedback after a limited, initial review. In my opinion there is more to cover but I am also not familiar with the standards expected so don't want my _opinion_ to get in the way too much either.

Source/UI/NewUI/BillingForm/Style/Default Implmentation/BillingForm/Fields/AddressLine1/DefaultFormStyleAddressLine1Cell.swift
```diff
@@ -0,0 +1,23 @@
+import UIKit
+
+//Address line 1
+struct DefaultFormStyleAddressLine1Cell : BillingFormTextFieldCellStyle {
```
alex-ioja-yang-cko, line 4: Do you have some custom formatting applied locally? I am familiar with a space after comment tags (`// comment`), as well as no space after object declaration when inheriting or conforming (`struct MyCell: BillingFormTextFieldCellStyle`), but your code is consistently the other way around 😄.

alex-ioja-yang-cko: This is not technically a cell, its a configuration object that will be applied to a cell. As such, the naming is getting very long very quickly, is the enum option I proposed separately an option?

Source/UI/NewUI/BillingForm/View/Field/BillingFormTextFieldView.swift
```diff
@@ -0,0 +1,171 @@
...
+
+extension BillingFormTextFieldView {
+
+    private func setupViews(){
+        backgroundColor = style.backgroundColor
+        setupHeaderLabel()
+        setupHintLabel()
+        setupTextFieldContainer()
+        setupTextField()
+        setupErrorView()
```
alex-ioja-yang-cko, line 92: I see that across the project this method of breaking down the setup of the UI was avoided. I think there is a good reason for it, as in your current approach, I can come and reorder your setup to a way of my linking. That will crash the app when I try to add a constraint between 2 views that don't share the same view hierarchy (ie:  pin top edge of errorView to bottom edge of text field, if text field is not already added as a subview).

I agree there is value in clearly separating and organising the setup, but that may be done better without using functions:
```
// Header

         addSubview(headerLabel)
         NSLayoutConstraint.activate([
             headerLabel.topAnchor.constraint(equalTo: topAnchor),
             headerLabel.leadingAnchor.constraint(equalTo: leadingAnchor),
             headerLabel.trailingAnchor.constraint(equalTo: trailingAnchor)
         ])

// Hint Label

         addSubview(hintLabel)
         NSLayoutConstraint.activate([
             hintLabel.topAnchor.constraint(equalTo: headerLabel.safeBottomAnchor, constant: 2),
             hintLabel.leadingAnchor.constraint(equalTo: leadingAnchor),
             hintLabel.trailingAnchor.constraint(equalTo: trailingAnchor)
         ])
.....
```

Source/UI/NewUI/BillingForm/View/Field/BillingFormTextFieldView.swift
```diff
@@ -0,0 +1,171 @@
...
+            errorView.heightAnchor.constraint(equalToConstant: style.error.isHidden ? 0 : style.error.height)
+        ])
+    }
+}
+
+extension BillingFormTextFieldView: UITextFieldDelegate {
+    
+    func textFieldDidBeginEditing(_ textField: UITextField) {
+        textFieldContainer.layer.borderColor = style.textfield.focusBorderColor.cgColor
+        self.delegate?.textFieldShouldBeginEditing(textField: textField)
```
alex-ioja-yang-cko, line 154: inconsistent use of `self`

Tests/Mocks/BillingFormTextFieldCellMockDelegate.swift
```diff
@@ -0,0 +1,37 @@
+import UIKit
+@testable import Frames
+
+class BillingFormTextFieldCellMockDelegate: BillingFormTextFieldCellDelegate {
+    
+    var textFieldShouldBeginEditingCalledTimes = 0
+    var textFieldShouldBeginEditingLastCalledWithTextField: UITextField?
```
alex-ioja-yang-cko, line 7: As all software unit tests can grow heavily quickly, so I would propose to join the counter and the tracker property into an array of called properties.

Where you currently record things like:
```
calledTimes += 1
lastCalled = textField
```

you could instead record as:
```
calledTextFields.append(textField)
```

This now allows you to test counter as well as verify every single passed object if you did expect more than one to be passed. And you only need 1 property, not 2.

Source/Validator/StateValidator.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+class StateValidator: Validator {
+    func validate(text: String?) -> Bool {
+        return isEmpty(text: text)
```
alex-ioja-yang-cko, line 5: Implicit return is possible. Is this final version, a StateValidator just checks text exists?

Source/Validator/StateValidator.swift
```diff
@@ -0,0 +1,13 @@
...
+    }
+    
+    private func isEmpty(text: String?) -> Bool {
+        guard let text = text, !text.isEmpty else { return true }
+        return false
```
alex-ioja-yang-cko, line 10: can join this into a simpler line
`text?.isEmpty != false`

Tests/Mocks/AddressViewControllerMockDelegate.swift
```diff
@@ -1,4 +1,4 @@
-import Foundation
+import UIKit
```
alex-ioja-yang-cko, line 1: Not sure this is intended as no other changes were made to the file

Tests/UI/New UI/BillingForm/Validator/AddressLine2ValidatorTests.swift
```diff
@@ -0,0 +1,19 @@
...
+        let text = ""
+        let isValid = expectedType.validator.validate(text: text)
+        XCTAssertEqual(isValid, true)
+    }
+    
+    func testValidationWhenTextIsNonEmptyThenShouldBeFalse() {
+        let expectedType = BillingFormCellType.addressLine2
+        let text = "addressLine2"
+        let isValid = expectedType.validator.validate(text: text)
+        XCTAssertEqual(isValid, false)
```
alex-ioja-yang-cko, line 17: When asserting against a boolean it may be clearer to use the correct assertion API (ie: `XCTAssertTrue`, `XCTAssertFalse`). I see this was missed across on many occasions so addressing it may provide _muscle memory_.

Tests/UI/New UI/BillingForm/Validator/PostCodeValidatorTests.swift
```diff
@@ -0,0 +1,19 @@
...
+class PostCodeValidatorTests: XCTestCase {
+
+    func testValidationWhenTextIsEmptyThenShouldBeTrue() {
+        let expectedType = BillingFormCellType.postcode
+        let text = ""
+        let isValid = expectedType.validator.validate(text: text)
+        XCTAssertEqual(isValid, true)
+    }
+    
+    func testValidationWhenTextIsNonEmptyThenShouldBeFalse() {
```
alex-ioja-yang-cko, line 13: There is no test to validate correct functionality, just the no input and the invalid format.

Source/UI/NewUI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -0,0 +1,119 @@
...
+    func tableView(numberOfRowsInSection section: Int) -> Int {
+        style.fields.count + 1
+    }
+    
+    func tableView(cellForRowAt indexPath: IndexPath, sender: UIViewController) -> UITableViewCell {
+        getCell(for: indexPath.row, delegate: sender)
+    }
+    
+    func doneButtonIsPressed(sender: UIViewController) {
+        let countryCode = "44" //"\(addressView.phoneInputView.phoneNumber?.countryCode ?? 44)"
```
alex-ioja-yang-cko, line 89: I think there is something wrong on this line but can't tell for certain how it should be

Source/UI/NewUI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -0,0 +1,119 @@
...
+        self.editDelegate = cell
+        return cell
+    }
+    
+    private func getFieldCell(for row: Int, delegate: BillingFormTextFieldCellDelegate?) -> UITableViewCell {
+        guard let type = BillingFormCellType(rawValue: row) else { return UITableViewCell()}
+        var style = style.fields[row - 1]
+        updateErrorView(with: &style, type: type)
+        updateTextField(with: &style, type: type)
+        return BillingFormTextFieldCell(type: type, style: style, delegate: delegate)
```
alex-ioja-yang-cko, line 47: I am afraid your current approach is missing cell dequeuing. As a practice this is something I disagree with, but I am not sure I understand sufficiently to assess if you found an edge case where this becomes acceptable? Happy to understand more.

alex-ioja-yang-cko: Your approach to the use of style is very interesting. You start with a default value, then attempt to override a number of times, whilst the methods are called `update....` and yet they don't update what they claim to, they just mutate a property. 

Can we rethink this flow ti a more common way?

alex-ioja-yang-cko (changes requested):
As it stands I think the lack of cell dequeuing is something that absolutely must be documented if intended or fixed. 

[UITableView. dequeueReusableCell](https://developer.apple.com/documentation/uikit/uitableview/1614891-dequeuereusablecell)
> For performance reasons, a table view’s data source should generally reuse [UITableViewCell](https://developer.apple.com/documentation/uikit/uitableviewcell) objects when it assigns cells to rows in its [tableView(_:cellForRowAt:)](https://developer.apple.com/documentation/uikit/uitableviewdatasource/1614861-tableview) method

Source/UI/NewUI/BillingForm/Style/Cell/BillingFormTextFieldCellStyle.swift
ehab-al-cko: So all fields are optionals? can the merchant decide to remove all fields and it will show nothing?

Source/UI/Controllers/CardViewController.swift
```diff
@@ -339,3 +344,16 @@ public class CardViewController: UIViewController,
         }
     }
 }
+
+
+extension CardViewController: BillingFormViewModelDelegate {
+    func onTapDoneButton(address: CkoAddress, phone: CkoPhoneNumber) {
+        billingDetailsAddress = address
+        billingDetailsPhone = phone
+        let value = "\(address.addressLine1 ?? ""), \(address.city ?? "")"
```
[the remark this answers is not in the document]
ehab-al-cko: I agree. Out of scope. This is from old code. you can ask Patric to add a new ticket to refactor this part.

Source/UI/NewUI/BillingForm/ViewController/BillingFormViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+        delegate?.cancelButtonIsPressed(sender: self)
+    }
+}
+
+// ViewModel
+extension BillingFormViewController {
+    private func setupViewModel() {
+        viewModel.updateRow = {
+            DispatchQueue.main.async { [weak self] in
+                guard let self = self, let row = self.viewModel.updatedRow else { return }
```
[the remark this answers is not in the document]
ehab-al-cko: Yes.
https://github.com/apple/swift/pull/15306

Source/UI/NewUI/BillingForm/ViewModel/BillingFormCellType.swift
```diff
@@ -0,0 +1,25 @@
...
+    case fullName = 1
+    case addressLine1 = 2
+    case addressLine2 = 3
+    case city = 4
+    case state = 5
+    case postcode = 6
+    case country = 7
+    case phoneNumber = 8
```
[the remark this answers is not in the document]
ehab-al-cko: I think no need for raw type of Int.

Source/UI/NewUI/BillingForm/View/Field/BillingFormTextFieldView.swift
```diff
@@ -0,0 +1,171 @@
...
+    }
+    
+    private func setupErrorView() {
+        addSubview(errorView)
+        NSLayoutConstraint.activate([
+            errorView.topAnchor.constraint(equalTo: textFieldContainer.bottomAnchor, constant: 10),
+            errorView.leadingAnchor.constraint(equalTo: leadingAnchor),
+            errorView.trailingAnchor.constraint(equalTo: trailingAnchor),
+            errorView.bottomAnchor.constraint(equalTo: bottomAnchor),
+            errorView.heightAnchor.constraint(equalToConstant: style.error.isHidden ? 0 : style.error.height)
```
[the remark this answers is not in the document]
ehab-al-cko: It's actually resizing when reload cell.

Source/UI/NewUI/BillingForm/ViewModel/BillingFormViewModel.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+
+protocol BillingFormViewModelDelegate {
+    /// Executed when an user tap on the done button.
+    ///
+    /// - parameter controller: `AddressViewController`
+    /// - parameter address: Address entered by the user
+    func onTapDoneButton(address: CkoAddress, phone: CkoPhoneNumber)
```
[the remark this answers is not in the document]
ehab-al-cko: Nice catch. I copied some of old logic. I will update documentation .

iOS Example Frame SPM/iOS Example Frame/MainViewController.swift
```diff
@@ -12,8 +12,15 @@ import Frames
...
     
     @IBOutlet weak var goToPaymentPageButton: UIButton!
+    var isNewUI: Bool = false
     
     @IBAction func goToPaymentPage(_ sender: Any) {
+        cardViewController.isNewUI = false
+        navigationController?.pushViewController(cardViewController, animated: true)
+    }
+    
+    @IBAction func goToNewPaymentPage(_ sender: Any) {
```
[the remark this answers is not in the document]
ehab-al-cko: New UI and old UI for testing both logic.

Source/UI/NewUI/BillingForm/Factory/BillingFormFactory.swift
ehab-al-cko: I understand, what are the common patterns?

Tests/Mocks/BillingFormTextFieldCellMockDelegate.swift
ehab-al-cko: I think, i have a better idea. we can start using Sourcery. what do you think?
https://github.com/krzysztofzablocki/Sourcery

Source/Validator/StateValidator.swift
ehab-al-cko: For now, Yes.

Tests/UI/New UI/BillingForm/Validator/PostCodeValidatorTests.swift
ehab-al-cko: please explain more.

Source/UI/NewUI/BillingForm/ViewController/BillingFormViewController.swift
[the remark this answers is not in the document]
alex-ioja-yang-cko: I am not sure I follow the relevance of that post. I am not asking what the syntax `guard let self = self` does, I am asking why do we need to have a strong referenced self in this case?

We can make use of the guard that checks if the `updatedRow` is not nil, to also validate `self` is not nil, without implicitly capturing self in the closure:

```
viewModel.updateRow = {
             DispatchQueue.main.async { [weak self] in
                 guard let row = self?.viewModel.updatedRow else { return }
                 let indexPath = IndexPath(row: row, section: 0)
                 self?.tableView.reloadRows(at: [indexPath], with: .automatic)
             }
         }
```

I may be mistaken as I didn't check inside compiler, but this code should work just the same, whilst continuing to hold a weak self reference.

Happy to get into it further on the benefits of this instead of automatically defaulting to strong self captures every time.

Source/UI/NewUI/BillingForm/Factory/BillingFormFactory.swift
alex-ioja-yang-cko: From my experience, the delegate is commonly exposed as it can occur it may need overriding. But equally its not a requirement to the initialisation of the object, just a requirement for the object to return its updates externally.

Then this factory to create VCs. I don't mind using factories, but I wonder if when you create a VC you **need** to have visibility and control over the requirements at the presentation point. I think the standard flow here when presenting VCs would be

```
let viewModel = ScreenViewModel()
let screen = Screen(with: viewModel)
viewModel.delegate = screen
present(screen)
```

The main benefit for this is the clarity of how the VC and that VM were created and any customisations done at presentation stage. Abstracting externally (don't mind a local method `createScreen`) hides this code that is not really meant to be reused that often (how many places present this screen in the UI flows? )

Source/UI/NewUI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+    
+    lazy var errorFlagOfCellType = [BillingFormCellType: Bool]()
+    lazy var textValueOfCellType = [BillingFormCellType: String]()
+    
+    private var initialCountry: String
+    private var initialRegionCode: String?
+    var editDelegate: BillingFormViewModelEditingDelegate?
+    var delegate: BillingFormViewModelDelegate?
```
alex-ioja-yang-cko, line 18: weak

Source/UI/NewUI/BillingForm/Style/Cell/BillingFormTextFieldCellStyle.swift
harry-brown-cko: In theory yes. If all fields were disabled we should probably disable the billing form

Tests/Mocks/BillingFormTextFieldCellMockDelegate.swift
alex-ioja-yang-cko: I have only seen Sorcery used in 2 projects so far and they were actually teams that were not very well focussed on native, so they tried to use approaches that could be easily used in an analogue project on a different project. Personally I am a fan of learning eco stem well first and using available tools correctly, before trading off native to more cross team.

Having said that, I don't want to dismiss something only based on my personal historic bad experience. I think it could be very valuable to understand how you see it and for us as a team to discuss on it more.

For now, this is unit test, not live code, we could keep as is, but would also like Deepesh or Harry to confirm they see no problem with it. For me, I am always weary when merging some pattern that can immediately be improved on that it establishes the wrong models for others to potentially copy in the future. But at the same time, testing is better than no testing, so 🤷‍♂️

Tests/UI/New UI/BillingForm/Validator/PostCodeValidatorTests.swift
alex-ioja-yang-cko: I looked into the PostCodeValidator now and I can see that it only checks if empty or not, so I suppose I was expecting some type of confirmation the postcode validation works correctly.

I now see that's not the case, but I do need to now ask if creating many validators that do the exact same thing (check value exists) is a good solution. We end up with many identical placeholders, that whilst I can see how they can grow, I also see they are not used and (knowing software development) are probably going to go empty for a long time, potentially forever.

Should we just have a validator `ContentExistsValidator` if we are very keen on following with validators. Then as we find ways to test postcodes, names or other functionality, we replace as and when needed with the correct fit validator for that use case.

Source/UI/NewUI/BillingForm/View/Field/BillingFormPhoneNumberText.swift
```diff
@@ -0,0 +1,86 @@
+import UIKit
+import PhoneNumberKit
+
+final class BillingFormPhoneNumberText: UITextField {
+    var type: BillingFormCellType
+    /// Phone Number Kit
+    let phoneNumberKit = PhoneNumberKit()
+    var partialFormatter: PartialFormatter {
+        return PartialFormatter.init(phoneNumberKit: phoneNumberKit, defaultRegion: "GB", withPrefix: true)
```
alex-ioja-yang-cko, line 9: Formatters are usually quite an expensive operation so creating a new one every time is not the most efficient approach.

I would instead suggest having a lazy loaded formatter that gets created locally:
ie:
```
private lazy var partialFormatter: PartialFormatter = createPhoneNumberFormatter()

private func createPhoneNumberFormatter() -> PartialFormatter {
    PartialFormatter()
}
```

Additionally, do we need to use the provided arguments? I am not sure defaultRegion "GB" is a safe assumption, and the `convenience initialiser` from PartialFormatter already uses the same defaults as we do.

Source/UI/NewUI/BillingForm/View/Field/BillingFormPhoneNumberText.swift
```diff
@@ -0,0 +1,86 @@
...
+        if let startPosition = textField.selectedTextRange?.start {
+            targetCursorPosition = textField.offset(from: textField.beginningOfDocument, to: startPosition)
+        }
+
+        let phoneNumber = textField.text!
+        let formatted = partialFormatter.formatPartial(phoneNumber)
+        textField.text = formatted
+
+        if var targetPosition = textField.position(from: textField.beginningOfDocument, offset: targetCursorPosition) {
+            if targetCursorPosition != 0 {
```
alex-ioja-yang-cko, line 62: I am confused on what this code does (with targetCursor, previousPosition, targetPosition), why do we need it?

Source/UI/NewUI/BillingForm/View/Field/BillingFormTextField.swift
```diff
@@ -0,0 +1,13 @@
+import UIKit
+
+final class BillingFormTextField: UITextField {
+    var type: BillingFormCellType
```
alex-ioja-yang-cko, line 4: Should this be a let, or could we remove the parameter from init? If this is modifiable but also required at init, its a little inconsistent?

Source/UI/NewUI/BillingForm/View/Field/BillingFormTextFieldView.swift
```diff
@@ -0,0 +1,171 @@
...
+        view.text = style.hint?.text
+        view.font = style.hint?.font
+        view.textColor = style.hint?.textColor
+        view.numberOfLines = 0
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.backgroundColor = .white
+        return view
+    }()
+    
+    private(set) lazy var textFieldContainer: UIView = {
```
alex-ioja-yang-cko, line 38: Why are the UI components here private(set)? Tried searching for some but can't find them being used externally, so private seemed appropriate?

Source/UI/NewUI/BillingForm/View/TableViewCell/Header/BillingFormHeaderCell.swift
```diff
@@ -0,0 +1,138 @@
...
+        view.setTitleColor(style.doneButton.activeTitleColor, for: .normal)
+        view.setTitleColor(style.doneButton.inActiveTitleColor, for: .disabled)
+        view.tintColor = view.isEnabled ? style.doneButton.activeTintColor : style.doneButton.inActiveTintColor
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.backgroundColor = style.doneButton.backgroundColor
+        view.addTarget(self, action: #selector(doneAction), for: .touchUpInside)
+        return view
+    }()
+    
+    private lazy var headerLabel: UILabel = {
```
alex-ioja-yang-cko, line 39: I can't see this field used outside of the creation and setup, so I wonder if we need to have this property, or when we create it for the view, we just create, customise, add to view and allow to lose local reference.

Probably other similar UI components are doing the same thing?

Source/UI/Controllers/CardViewController.swift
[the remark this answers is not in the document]
ehab-al-cko: Yea sure, Old means old UI logic code from `AddressViewController`

Source/UI/NewUI/BillingForm/Style/Default Implmentation/BillingForm/Fields/AddressLine1/DefaultFormStyleAddressLine1Cell.swift
ehab-al-cko: I think, no need for cell dequeue for this case. let's discuss more after vacation 
https://www.codementor.io/@warrenburton/do-we-need-to-dequeue-tableview-cells-w8szp55uq

alex-ioja-yang-cko: Sounds great, thank you for taking the time while you're away @ehab-al-cko . 

Not actionable now, but want to make sure I don't forget, that article is perfectly valid and I know I've asked myself many times that question, so getting someone to build a project and compare it really is useful. However, specifically for our use case, I would still be inclined to use the better performance method as we have no control over what that application does. For all intents and purposes, imagine app using our SDK is doing crypto mining... Any reasonable easy to gain performance improvement on our side will allow them to keep more resources available. Not saying we should over optimise early (I am quite against that as well), but if the standard practice carries an optimisation benefit (however big/small), then that standard practice should be the choice for an SDK).

deepesh-vasthimal-cko:
Trying to run this branch locally to review the UI, I see some build issue. Please can you check below and suggest. 
Screenshot 2022-05-06 at 21 24 50

Source/UI/NewUI/BillingForm/View/Field/BillingFormTextFieldView.swift
ehab-al-cko: for testing purpose, 
 XCTAssertEqual(view.textFieldContainer.layer.borderColor, style.textfield.normalBorderColor.cgColor)

Source/UI/NewUI/BillingForm/View/Field/BillingFormPhoneNumberText.swift
ehab-al-cko: This can be refactored in another ticket. This is reused from old code

Source/UI/NewUI/BillingForm/Style/Default Implementation/BillingForm/Header/Cancel/DefaultCancelButtonFormStyle.swift
```diff
@@ -0,0 +1,36 @@
...
+    init( isEnabled: Bool = true,
+          text: String = "Cancel",
+          font: UIFont = UIFont.systemFont(ofSize: 17),
+          activeTitleColor: UIColor = .brandeisBlue,
+          inActiveTitleColor: UIColor = .doveGray,
+          inActiveTintColor: UIColor = .doveGray,
+          activeTintColor: UIColor = .brandeisBlue,
+          backgroundColor: UIColor = .white,
+          height: Double = 44,
+          width: Double = 53) {
```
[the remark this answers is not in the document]
deepesh-vasthimal-cko: Good catch @alex-ioja-yang-cko . We did discuss this over our last refinement. We are waiting for our designer feedback. I think we do need some min height & width for all fields which will be part of a separate ticket and also allow merchants to pass in their own width and height.. Something which we did not discuss this on ACs for this ticket. 

@ehab-al-cko @harry-brown-cko you can keep that in mind so that you are aware of this change coming along your way.

ehab-al-cko:
> Trying to run this branch locally to review the UI, I see some build issue. Please can you check below and suggest. Screenshot 2022-05-06 at 21 24 50

Fixed

Source/UI/NewUI/BillingForm/Style/Protocols/InputLabelStyle.swift
```diff
@@ -0,0 +1,9 @@
+import UIKit
+
+// Label element
+public protocol InputLabelStyle {
```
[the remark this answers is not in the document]
ehab-al-cko: Protocols is used for DI and testing cases

Source/UI/NewUI/BillingForm/Style/Protocols/FormTextFieldStyle.swift
```diff
@@ -0,0 +1,19 @@
+import UIKit
+
+// TextField element
+public protocol TextFieldStyle {
```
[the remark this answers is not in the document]
ehab-al-cko: Protocols is used for DI and testing cases

Source/UI/NewUI/BillingForm/Style/Protocols/FormButtonStyle.swift
```diff
@@ -0,0 +1,15 @@
+import UIKit
+
+//Button element
+public protocol FormButtonStyle {
```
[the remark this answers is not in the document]
deepesh-vasthimal-cko: Agree with @alex-ioja-yang-cko , @ehab-al-cko any idea why this is protocol and not struct may be we missed something here.

Source/UI/NewUI/BillingForm/View/Error/BillingFormTextFieldErrorView.swift
```diff
@@ -0,0 +1,68 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+extension BillingFormTextFieldErrorView {
+
+    private func setupViews(){
+        backgroundColor = style.backgroundColor
+        setupHeaderLabel()
+        setupImageView()
```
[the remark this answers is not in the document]
deepesh-vasthimal-cko: Question : @alex-ioja-yang-cko any idea why it would crash if kept in separate function ? wouldn't the compiler treat that as same. My understanding was the different functions is really there for readability. I am keep to know if something crashes and we can avoid that.

Source/UI/NewUI/BillingForm/View/Field/BillingFormPhoneNumberText.swift
```diff
@@ -0,0 +1,76 @@
...
+                let lastChar = formatted[formatted.index(formatted.startIndex,
+                                                         offsetBy: targetCursorPosition - 1)]
+                if lastChar == " " && previousTextCount < formatted.count && phoneNumber != formatted {
```
deepesh-vasthimal-cko: suggestion : You can create a let constant as string called whitespace and use it elsewhere instead of magic white space 
let whiteSpace = " "

Source/UI/NewUI/BillingForm/View/Field/BillingFormTextFieldView.swift
deepesh-vasthimal-cko: question : @alex-ioja-yang-cko , But even if all the code is in one big function could anyone not change Header and label up and down to break the code ? That will be applicable to entire project. My understanding is such scenarios are better covered off using UI tests where it checks whats expected behaviour and order and breaks the UI test if necessary.

Source/UI/NewUI/BillingForm/View/Error/BillingFormTextFieldErrorView.swift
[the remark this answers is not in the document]
alex-ioja-yang-cko: It doesn't crash if kept in separate functions, but having separate functions can make it appear simpler than it is and hide the complex dependency on the layout side.

If you create a constraint between 2 views that don't belong to the same stack, the application will crash. So keeping the correct layout as it was designed is very important. In this case, abstracting that complexity behind easy to navigate functions may give a false confidence that the code is easy. Of course, this does improve developer experience and can be kept, but we do need to be very careful on the code being committed.

Imagine tomorrow a new dev comes and copies this approach to a new view, see it running and all good, do some method shuffling as part as code review (lets say swap order on those 2 so the trailing edge faces a certain way, or order alphabetically 😜). If there is no test to crash when layout is created, that can sit silently for a while.
Take it further, a test may exist, but doing the setup view on init may be using resources too eagerly. You may later optimise by moving the setup to a lifecycle method that increases likelihood of reaching display (like `willMove(toSuperview`). Most unit tests will not trigger that method, allowing you to _safely refactor the order of the methods_ and have it start crashing. Process can obviously mitigate, but even UITests may not 100% cover all elements.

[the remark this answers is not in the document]
alex-ioja-yang-cko: oversimplified example for crash on constraint creation:
```
let superview = UIView()
let subview1 = UIView()
let subview2 = UIView()

superview.addSubview(subview1)
subview1.pinTrailingEdge(to: subview2) // This crashes as subview 1 and subview 2 don't have the same hierarchy
```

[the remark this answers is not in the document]
alex-ioja-yang-cko: I love keeping things simple, and I'd always suggest we break code down so Patrick can come and read the code like a story. But Layout Constraints are 1 thing I wouldn't want to oversimplify and I'd say keeping it harder encourages dev to really look at what they're doing 👀

Tests/UI/New UI/BillingForm/Validator/PostCodeValidatorTests.swift
ehab-al-cko: I will add more validation for some fields like country, phone number, name, etc in the next ticket

Source/UI/NewUI/BillingForm/View/Field/BillingFormTextFieldView.swift
alex-ioja-yang-cko: Agree, someone can always do that, but moving a block of 6 lines requires more attention than an abstract method name. Also, the big block can make obvious the constraint between 2 views that were added at the same time, as they're in the same viewing range.


Alternatively, we can use an extension method `addSubviews`, in which we add all the equal level subviews at the same time, then break down the constraint creation safely as they will be in the same hierarchy at that point.

```
addSubviews([headerLabel, hintLabel])
setupHeaderLabel()
setupHintLabel()
```

I've only ever previously used UITests for Product Core Journey testing, so I'm not sure you could ever achieve 100% coverage once you start being concerned of the resources the UI test uses.
Snapshot testing may catch it though. Those are like unit tests with full UI lifecycle called.

ehab-al-cko: change naming to `setupViewsInOrder`

Tests/Mocks/BillingFormTextFieldCellMockDelegate.swift
ehab-al-cko: I think, this requires a tech meeting for this update to choose best approach for future work.

Source/UI/NewUI/BillingForm/Style/Protocols/InputLabelStyle.swift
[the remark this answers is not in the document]
alex-ioja-yang-cko: I am not sure DI is a concern when the object facing us is a configuration value type object.

On a flow of
```
Configuration -> View -> Display
```
If the Display property matches our intended Configuration property, we know View mapped them correctly. Not sure why Configuration would need to conform to something 🤔

Source/UI/NewUI/BillingForm/Style/Protocols/FormButtonStyle.swift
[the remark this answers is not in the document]
ehab-al-cko: adding defaultCancel and defaultDone are not following on SRP

Tests/Mocks/BillingFormTextFieldCellMockDelegate.swift
alex-ioja-yang-cko: I have also [posted on Slack](https://checkout.slack.com/archives/C038WRNDKTM/p1651661700347499) a link to an article describing exactly the flow you're going through. The creator starts with your exact approach, then improves it to what I suggested and I think explains quite well the transition.

Would really like your input on that, regardless of the changes (/ lack of) in this particular case 😉

Source/UI/NewUI/BillingForm/View/Error/BillingFormTextFieldErrorView.swift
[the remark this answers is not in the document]
ehab-al-cko: Renaming method to `setupViewsInOrder`

Source/UI/NewUI/BillingForm/View/Field/BillingFormPhoneNumberText.swift
ehab-al-cko: This part is from old code, it will be refactored next country ticket.

Source/UI/NewUI/BillingForm/Style/Default Implementation/BillingForm/Header/Cancel/DefaultCancelButtonFormStyle.swift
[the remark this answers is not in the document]
ehab-al-cko: Yes, it's safe

Source/UI/NewUI/BillingForm/View/Field/BillingFormTextFieldView.swift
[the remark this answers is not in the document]
ehab-al-cko: Project is fixed. You may run the project now

Source/UI/NewUI/BillingForm/Style/Cell/BillingFormTextFieldCellStyle.swift
ehab-al-cko: Hint and error are optionals. The merchant can decide which field should show.

Source/UI/NewUI/BillingForm/Style/Protocols/FormButtonStyle.swift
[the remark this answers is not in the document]
alex-ioja-yang-cko: I think we may be mixing things a little. 

As per [Wikipedia definition](https://en.wikipedia.org/wiki/Single-responsibility_principle)
> The single-responsibility principle (SRP) is a computer-programming principle that states that every [module](https://en.wikipedia.org/wiki/Modular_programming), [class](https://en.wikipedia.org/wiki/Class_(computer_programming)) or [function](https://en.wikipedia.org/wiki/Function_(computer_programming)) in a [computer program](https://en.wikipedia.org/wiki/Computer_program) should have responsibility over a single part of that program's [functionality](https://en.wikipedia.org/wiki/Software_feature), and it should [encapsulate](https://en.wikipedia.org/wiki/Encapsulation_(object-oriented_programming)) that part. All of that module, class or function's [services](https://en.wikipedia.org/wiki/Service_(systems_architecture)) should be narrowly aligned with that responsibility.

We are looking at an object that provides state description for another object. I think that acting as a factory for those states is narrowly aligned with that responsibility. As examples from the Apple ecosystem, I already checked and can say that `UILayoutPriority` and `UIColor` behave the same. If we dig more we will find this is a very common pattern in the ecosystem and also doesn't appear to breach SRP. Not sure if you've used SwiftUI yet, that would not work without this approach 😄

Source/UI/NewUI/BillingForm/Style/Cell/BillingFormHeaderCellStyle.swift
alex-ioja-yang-cko: Not going to lie, going ahead with these comments above the class/protocol declaration makes our publicly exposed code look unfinished.

It would be nice if our exposed code would make CKO appear knowledgeable and professional, not showing useless comments. This occurs in a few places, happy to flag them if we decide to do something about it.

Source/UI/NewUI/BillingForm/Style/Protocols/InputLabelStyle.swift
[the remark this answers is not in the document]
alex-ioja-yang-cko: Before resolving, would you mind providing some form of closure to my argument? I would even settle for a follow up somewhere else.

All of this code is public facing, and I'd like to hope many dev teams would use it. I would like to know that when the likes of `NewDay` or other fintechs with large iOS teams dig into our exposed code, they will be excited and trust us. Not make us look like we have a weak grasp of Swift / iOS functionality and question the dependency in the first place.

Source/UI/NewUI/BillingForm/Style/Cell/BillingFormHeaderCellStyle.swift
ehab-al-cko: Fixed

alex-ioja-yang-cko (commented):
Tried running locally. Not sure what state the functionality progress is in, but I got this.

Some opinions on this screenshot alone:
- the * in the label name is not matching the expectation of being compulsory or not
- Country appears twice, once compulsory, once with lower case letter
- It is not possible to dismiss keyboard by tapping background
- screen is dismissible by simply scrolling down the list too far by mistake

Simulator Screen Shot - iPhone 13 Pro Max - 2022-05-10 at 12 57 13
 
 Additionally, there is a memory leak. In screenshot you can see Billing not presented, but 6 VCs in memory and a number of other entries.
Screenshot 2022-05-10 at 13 02 25

ehab-al-cko:
> Tried running locally. Not sure what state the functionality progress is in, but I got this.
> 
> Some opinions on this screenshot alone:
> 
> * the * in the label name is not matching the expectation of being compulsory or not  
( This is localised string, I will ask Patrick for it) 
> * Country appears twice, once compulsory, once with lower case letter
( Country is not implemented Yet, Next ticket is for Country logic)
> * It is not possible to dismiss keyboard by tapping background
( This is can be done with next ticket)
> * screen is dismissible by simply scrolling down the list too far by mistake
> Need more details for it.

> Simulator Screen Shot - iPhone 13 Pro Max - 2022-05-10 at 12 57 13
> 
> Additionally, there is a memory leak. In screenshot you can see Billing not presented, but 6 VCs in memory and a number of other entries. Screenshot 2022-05-10 at 13 02 25
**[ Out of Scope]**

Source/UI/NewUI/BillingForm/Factory/BillingFormFactory.swift
deepesh-vasthimal-cko: @ehab-al-cko to Create Tech investment ticket to address this in new PR

Source/UI/NewUI/BillingForm/Style/Protocols/FormButtonStyle.swift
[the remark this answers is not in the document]
deepesh-vasthimal-cko: We kind of follow this approach of using protocols in frames SDK.

Source/UI/NewUI/BillingForm/Style/Protocols/InputLabelStyle.swift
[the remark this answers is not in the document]
deepesh-vasthimal-cko: Agree between devs : We kind of follow this approach of using protocols in frames SDK.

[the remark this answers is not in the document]
deepesh-vasthimal-cko: public facing interface to have below convention for jazzy to pick proper doc output 
`/// InputLabelStyle for text fields components `

Source/UI/NewUI/BillingForm/View/TableViewCell/Header/BillingFormHeaderCell.swift
ehab-al-cko: Resolved on slack

Cartfile.resolved
```diff
@@ -1 +1 @@
-github "marmelroy/PhoneNumberKit" "3.3.3"
+github "marmelroy/PhoneNumberKit" "3.3.7"
```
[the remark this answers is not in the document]
ehab-al-cko: Should be reverted to 3.3.3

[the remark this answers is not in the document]
ehab-al-cko: Done

Source/Extensions/StringExtensions.swift
```diff
@@ -22,4 +22,23 @@ extension String {
         let bundle = getBundle(forClass: forClass)
         return UIImage(named: self, in: bundle, compatibleWith: nil) ?? UIImage()
     }
+ 
+    //https://www.hackingwithswift.com/example-code/core-graphics/how-to-render-a-pdf-to-an-image
+    func vectorPDFImage(forClass: AnyClass) -> UIImage? {
+        let bundle = getBundle(forClass: forClass)
+        guard let urlPath = bundle.url(forResource: self, withExtension: "pdf") else { return nil }
+        guard let document = CGPDFDocument(urlPath as CFURL) else { return nil }
+        guard let page = document.page(at: 1) else { return nil }
```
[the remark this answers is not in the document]
ehab-al-cko: TODO: a new ticket to refactor and test this change

deepesh-vasthimal-cko (approved):
Approved but do not me merge as the base branch needs to be `origin/release/4.0.0_RC`

PR #146, opened 2022-04-28
PI-1630: fix logging with checkout

harry-brown-cko:
## Proposed changes

## Types of changes

## Checklist

## Further comments


Source/UI/Controllers/CardViewController.swift
```diff
@@ -124,13 +129,7 @@ public class CardViewController: UIViewController,
         if suppressNextLog {
             suppressNextLog = false
         } else {
-            guard let checkoutAPIService = checkoutAPIService else {
-                return
-            }
-// Potential Task : add correlationID and logger to checkoutAPIService or Frames
-//          checkoutAPIService.logger.add(metadata: checkoutAPIService.correlationID(),
-//                                          forKey: .correlationID)
-//          checkoutAPIService.logger.log(.paymentFormPresented)
```
ehab-al-cko, line 133: no need for this line 
`//          checkoutAPIService.logger.log(.paymentFormPresented)
`

Source/UI/Controllers/CardViewController.swift
```diff
@@ -179,8 +178,7 @@ public class CardViewController: UIViewController,
 
     @objc func onTapAddressView() {
         navigationController?.pushViewController(addressViewController, animated: true)
-        // add  let logger: FramesEventLogging to checkoutAPIService
-        //checkoutApiClient?.logger.log(.billingFormPresented)
```
ehab-al-cko, line 183: no need for this line. 
` //checkoutApiClient?.logger.log(.billingFormPresented)`

Tests/Mocks/StubCheckoutAPIService.swift
```diff
@@ -9,15 +9,22 @@
 import Checkout
 
 final class StubCheckoutAPIService: Frames.CheckoutAPIProtocol {
-  private(set) var createTokenCalledWith: (paymentSource: PaymentSource, completion: (Result<TokenDetails, TokenisationError.TokenRequest>) -> Void)?
-  func createToken(_ paymentSource: PaymentSource, completion: @escaping (Result<TokenDetails, TokenisationError.TokenRequest>) -> Void) {
-    createTokenCalledWith = (paymentSource, completion)
-  }
+    private(set) var createTokenCalledWith: (paymentSource: PaymentSource, completion: (Result<TokenDetails, TokenisationError.TokenRequest>) -> Void)?
+    func createToken(_ paymentSource: PaymentSource, completion: @escaping (Result<TokenDetails, TokenisationError.TokenRequest>) -> Void) {
```
ehab-al-cko, line 13: i think, functions should e after variables

harry-brown-cko: e after variables?

Source/Core/CheckoutAPIService.swift
```diff
@@ -5,31 +5,72 @@
...
         self.checkoutAPIService = checkoutAPIService
         self.cardValidator = cardValidator
+        self.logger = logger
     }
 
     public func createToken(_ paymentSource: PaymentSource, completion: @escaping (Result<TokenDetails, TokenisationError.TokenRequest>) -> Void) {
         checkoutAPIService.createToken(paymentSource, completion: completion)
     }
+
+    static func buildFramesEventLogger(environment: Environment, getCorrelationID: @escaping () -> String) -> FramesEventLogger {
```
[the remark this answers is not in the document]
alex-ioja-yang-cko: Agree with @ehab-al-cko. I would plus that it may be . At this stage cleanest syntax would be if the logic was inside the initialiser for `FramesEventLogger`, so the service initialiser ends up looking like.
```
let logger = FramesEventLogger(environment: environment, getCorrelationID: { checkoutAPIService.correlationID })
```

This can also help us remove the import of UIKit into a file that is called service. Whilst not a huge problem in itself, that import should be an indication that separation may be improved :) .

Source/Core/Logging/FramesEventLogger.swift
```diff
@@ -14,24 +14,27 @@ protocol FramesEventLogging {
...
       self.dateProvider = dateProvider
     }
 
     // MARK: - FramesEventLogging
     
     func log(_ framesLogEvent: FramesLogEvent) {
-        
+
+        // by setting correlationID for every log, we always keep in sync with Checkout SDK.
+        add(metadata: getCorrelationID(), forKey: .correlationID)
```
alex-ioja-yang-cko, line 36: Do we expect that correlation ID to change since initialised over the lifetime of the SDK?

If so, would it be an option to change the type on the source of the variable (from what I see `Checkout. checkoutAPIService .correlationID`), and have it be of type `NSMutableString`?

Whilst `String` is a value type, and `NSString` is a reference type binded to a value (behaves like value type), `NSMutableString` can actually act as a mutable reference type. This can save the passing a closure from place to place and simplifying the chain. Biggest drawback though, the source of the information should have its value set via `.setString` instead of `=`, for the behaviour to work.

Example:
```
var str: NSMutableString = NSMutableString(string: "OLD")
var newStr = str
newStr.setString("NEW")
print(str, newStr) // New, New

str = "Other change"
print(str, newStr) // Other change, New

```

alex-ioja-yang-cko (commented):
A few suggestions. Happy to hear your thoughts

Source/Core/Logging/FramesEventLogger.swift
harry-brown-cko: This would be a good feature add in Checkout SDK, then it can be implemented here

Source/Core/Logging/RemoteProcessorMetadata+extensions.swift
```diff
@@ -0,0 +1,27 @@
...
+         appPackageVersion: String,
+         uiDevice: UIDevice) {
+
+        self.init(productIdentifier: Constants.productName,
+                  productVersion: Constants.version,
+                  environment: environment.rawValue,
+                  appPackageName: appPackageName,
+                  appPackageVersion: appPackageVersion,
+                  deviceName: uiDevice.modelName,
+                  platform: "iOS",
```
alex-ioja-yang-cko, line 24: I think we could also obtain this string from UIDevice, under `systemName`. [Apple Documentation](https://developer.apple.com/documentation/uikit/uidevice/1620054-systemname) doesn't directly clarify possible values, some [SO discussion](https://stackoverflow.com/questions/12038425/possible-values-for-uidevice-currentdevice-systemname) points that it comes as `iOS` from iOS 10+, `iPadOS` from iOS 15+ , as well as `tvOS` and `watchOS` where appropriate.

Source/Core/Logging/FramesEventLogger.swift
```diff
@@ -14,24 +14,46 @@ protocol FramesEventLogging {
...
     private let checkoutEventLogger: CheckoutEventLogging
     private let dateProvider: DateProviding
     
     // MARK: - Init
+
+    convenience init(environment: Environment, getCorrelationID: @escaping () -> String) {
+        let checkoutEventLogger = CheckoutEventLogger(productName: Constants.productName)
+        let appBundle = Foundation.Bundle.main
+        let appPackageName = appBundle.bundleIdentifier ?? "unavailableAppPackageName"
+        let appPackageVersion = appBundle.infoDictionary?["CFBundleShortVersionString"] as? String ?? "unavailableAppPackageVersion"
```
alex-ioja-yang-cko, line 27: What would you think of having the default hardcoded values inside an enum, whether the existing `Constants` or a different one. Something like `defaultBundleIdentifier` and `defaultBundleVersion`.

Don't mind your current approach, just thinking as a practice, any hardcoded value, even a fallback, may be handy to be given sufficient visibility and grouped.

Source/UI/Controllers/CardViewController.swift
alex-ioja-yang-cko: Being marked with red, looks like this was an old comment, and its now removed, not added.

Green lines are for code added, red lines for code being removed.

Tests/Core/Logging/FramesEventLoggerTests.swift
```diff
@@ -19,7 +19,7 @@ final class FramesEventLoggerTests: XCTestCase {
         stubCheckoutEventLogger = StubCheckoutEventLogger()
         stubDateProvider = StubDateProvider()
         subject = FramesEventLogger(
-            correlationID: stubcorrelationID,
+            getCorrelationID: { [stubcorrelationID] in stubcorrelationID },
```
alex-ioja-yang-cko, line 22: Would this alternative compile `getCorrelationID: { $0 },`?

I got stuck few seconds looking at that syntax, thinking how that comes as an array and returns an array when it wanted a String. 😕

Source/Core/Logging/FramesEventLogger.swift
```diff
@@ -14,24 +14,46 @@ protocol FramesEventLogging {
...
+
+        let uiDevice = UIKit.UIDevice.current
```
alex-ioja-yang-cko, line 29: Unnecessary source framework mentioned, as well as another with `Foundation` for `appBundle`. I'm curious how these ones come to be as I don't think autocomplete normally adds them (although recently it does the auto import)

alex-ioja-yang-cko (commented):
Could we also add some unit tests for:
- `CheckoutAPIService`
    - ensure the associated logger is configured as expected for the environment provided to the service

I think that's the main unit test missing, as the VC tests were already there and I see you uncommented them (assuming they now work if pipeline passes).

Otherwise, I think it looks good 👍

Source/UI/Controllers/CardViewController.swift
alex-ioja-yang-cko: Being marked with red, looks like this was an old comment, and its now removed, not added.

Green lines are for code added, red lines for code being removed.

harry-brown-cko: Yeah I assume this and a few of the following comments were added by accident

Source/Core/Logging/FramesEventLogger.swift
deepesh-vasthimal-cko: @harry-brown-cko can you create a tech investment ticket for this which can go into checkout SDK.

deepesh-vasthimal-cko (approved):

PR #155, opened 2022-05-18
PIMOB-1141: eager card checks in checkout and update parts missed in merge

harry-brown-cko:
## Proposed changes

2 sets of changes

1. update checkout sdk to later version (older version was committed previously by accident)
1. add eager card checks pr and into checkout sdk
    - note that eager card checks are not setup with frames currently. this will be added in a follow up PR

## Types of changes

## Checklist


Checkout/Checkout/Source/3DSHandling/ThreeDSError.swift
```diff
@@ -0,0 +1,22 @@
+//
+//  ThreeDSError.swift
+//  Checkout
+//
+//  Created by Daven.Gomes on 02/02/2022.
```
alex-ioja-yang-cko, line 5: File header show something else, is this file imported from somewhere else?

Checkout/Sample/Source/3DS/WebViewController.swift
```diff
@@ -0,0 +1,90 @@
...
+    let onDismiss: ((_ webViewResult: Result<String, ThreeDSError>) -> Void)?
+    let webViewConfig: WebViewConfig
+
+    enum WebViewConfig {
+      case url(challengeURL: URL, successURL: URL, failureURL: URL)
+      case demo
+    }
+  }
+
+  var viewModel: ViewModel? {
```
alex-ioja-yang-cko, line 23: Not sure this is a genuine viewModel. I think it would be more intuitive if the user of the VC was kickstarting things like:
```
let vc = WebViewController()
let webConfig = WebViewConfig(.....) // WebViewConfig.demo
vc.load(config: webConfig) {
    // Completion handler
}
```

As such I would drop the ViewModel altogether and maybe use a typealias on `((_ webViewResult: Result<String, ThreeDSError>) -> Void)?` to have it more presentable.

Issue with injecting completion handlers via properties is how easy it is to create retain cycles.
- you don't use delegate which offers 1 point of failure to make all implementations safe from retain cycles
- you don't need `@escaping` which is enforced by compiler when it detects async code

This is similar with Reactive patterns but those usually come with reference collections to cleanup :).

alex-ioja-yang-cko: Also, small thing, but at call site assigning a property wouldn't be clear that will have consequences.

```
// Assigning
vc.model = FlashyModel()

// Clearly using a method and as such reasonably assuming things may happen - especially as its called start :)
vc.start(with: FlashyModel()
```

Checkout/Sample/Source/3DS/ThreeDSViewController.xib
```diff
@@ -0,0 +1,86 @@
...
+                        </textField>
+                        <textField opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="left" contentVerticalAlignment="center" borderStyle="roundedRect" placeholder="Failure URL" textAlignment="natural" minimumFontSize="17" translatesAutoresizingMaskIntoConstraints="NO" id="oY9-KH-e6T">
+                            <rect key="frame" x="0.0" y="123" width="314" height="34"/>
+                            <fontDescription key="fontDescription" type="system" pointSize="14"/>
+                            <textInputTraits key="textInputTraits"/>
+                        </textField>
+                        <button opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="system" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="Gw7-Pd-By3">
+                            <rect key="frame" x="0.0" y="164" width="314" height="34"/>
+                            <state key="normal" title="Button"/>
+                            <buttonConfiguration key="configuration" style="plain" title="Load challenge from URL"/>
```
alex-ioja-yang-cko, line 54: Always need to be careful when setting labels in the nib on whether they are updated. Most projects don't bother to localise nibs, as it just adds more files to send for translation, so text content becomes obvious placeholders to ensure at any testing stage it gets noticed (ie: `START CHALLENGE` - if product / qa / anyone else saw that would probably feel encouraged to ask why).

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
```diff
@@ -83,6 +83,15 @@ public class CardValidator: CardValidating {
     return cardNumberValidator.validate(cardNumber: cardNumber)
   }
 
+  /// Checks whether a given card number is at least a partial match for any of the supported schemes.
+  /// - Parameters:
+  ///   - cardNumber: The card number to validate.
+  /// - Returns: The card's scheme if successful, else an error.
+  public func eagerValidate(cardNumber: String) -> Result<Card.Scheme, ValidationError.CardNumber> {
+    logManager.queue(event: .validateCardNumber)
```
alex-ioja-yang-cko, line 91: I hope our logger has a large enough buffer to handle logging on each user input.

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
```diff
@@ -178,63 +203,69 @@ public class CardValidator: CardValidating {
...
-      from: DateComponents(
-        year: fourDigitYear,
-        month: month
-      )
-    ), providedDate >= currentDate else {
+    guard
+      let cardExpiryDate = calendar.date(from: cardExpiryComponents),
+      let nextMonth = calendar.date(byAdding: .month, value: 1, to: cardExpiryDate),
+      nextMonth > currentDate
+    else {
```
alex-ioja-yang-cko, line 225: I think having the first guard statement inlined with guard, as well as the else should be pretty common layout for the code. If you're concerned of too long lines, SwiftLint has some rule for maximum number of characters in a line.

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
```diff
@@ -178,63 +203,69 @@ public class CardValidator: CardValidating {
...
+      return .failure(.invalidYear)
+    }
+
+    guard expiryYear < 100 else {
+      // 4 digit year
+      return .success(expiryYear)
+    }
+
+    // 2 digit year
+    return .success(2000 + expiryYear)
```
alex-ioja-yang-cko, line 269: This won't age very well in 80 years. Although there's a good chance we won't care 😆

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
```diff
@@ -178,63 +203,69 @@ public class CardValidator: CardValidating {
...
+        return .failure(.invalidYearString)
+      }
+
+      return fourDigitYear(from: year)
     default:
       return .failure(.invalidYear)
     }
   }
+
+  private func fourDigitYear(from expiryYear: Int) -> Result<Int, ValidationError.ExpiryDate> {
```
alex-ioja-yang-cko, line 249: Just a side note that it feels weird to support 2 digit years. If you think while typing:
"2" = Invalid
"20" = "2020"
"203" = Invalid
"2034" = "2034"

If we're ok with this possible behaviour on the UX, no problem, but may just be worth being aware of possible uses of an expectation that 2 digits is safe.

Checkout/Checkout/Test/Stubs/StubCardValidator.swift
```diff
@@ -9,6 +9,14 @@ import Foundation
 @testable import Checkout
 
 final class StubCardValidator: CardValidating {
+  var eagerValidateCardNumberToReturn: Result<Card.Scheme, ValidationError.CardNumber> = .success(.visa)
+  private(set) var eagerValidateCardNumberCalledWith: String?
```
alex-ioja-yang-cko, line 13: I wouldn't generally worry about access control in stubs. It would be clearly visible at use site if it was not handled correctly.

alex-ioja-yang-cko (commented):
Haven't had time to see it all, but nothing major so far imo

Checkout/Checkout/Integration/CardValidatorTests.swift
```diff
@@ -0,0 +1,78 @@
...
+//  Created by Harry Brown on 22/02/2022.
+//
+
+import XCTest
+import Checkout
+
+final class CardValidatorTests: XCTestCase {
+  private let subject = CardValidator(environment: .sandbox)
+  private let calendar = Calendar(identifier: .gregorian)
+
```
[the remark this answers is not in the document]
alex-ioja-yang-cko: I may have seen at some point some tests for that logic already, there may be some duplication / need for separate test cases

Checkout/Checkout/Source/3DSHandling/ThreeDSError.swift
deepesh-vasthimal-cko: Yes this one is just moved from previous repo, which was missed to add on previous PRs.

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
deepesh-vasthimal-cko: @harry-brown-cko is there required to log each time we call eagerValidate ? , This will be called on every character changed in text field right ?

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
deepesh-vasthimal-cko: 2 or 4 digits are both valid for expiry year. This is expected behaviour. Do you think in any case this will have issues ?

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
deepesh-vasthimal-cko: we did discuss this . 80 years is good enough for now.

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
alex-ioja-yang-cko: If we have it all under control and implemented, no problem. But if we are going to update the UI to read this result on every input change and we are using it incorrectly it could lead to the chain I mentioned in previous message.

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
alex-ioja-yang-cko: Thought so, all this code will probably be replaced by AI generated Metaverses 🤣

Checkout/Checkout/Test/Stubs/StubCardValidator.swift
alex-ioja-yang-cko: Also, you guys may have discussed this before, and I know its a very pedantic detail, but these stuck are [technically mocks](https://stackoverflow.com/a/5180286) 🙈.

Checkout/Sample/Source/3DS/ThreeDSViewController.swift
```diff
@@ -0,0 +1,105 @@
...
+    case .success(let token):
+      return ("Success", token)
+    case .failure(let error):
+      return ("Failure", "ThreeDSError \(error.code)")
+    }
+  }
+
+  private func presentWebView(webViewConfig: WebViewController.ViewModel.WebViewConfig) {
+    let webViewController = WebViewController()
+    webViewController.viewModel = WebViewController.ViewModel(onDismiss: onWebViewDismiss, webViewConfig: webViewConfig)
```
alex-ioja-yang-cko, line 99: As mentioned separately about the risk of passing completion handlers as properties, I think this `onWebViewDismiss` implicitly captures self and actually has the potential to create a retain cycle.

Checkout/Sample/Source/3DS/WebViewController.swift
```diff
@@ -0,0 +1,90 @@
...
+      <a href="https://www.example.com/">Failure</a>
+      <a href="https://www.google.com/">Other</a>
+    </div>
+  </html>
+  """.trimmingCharacters(in: .whitespaces)
+}
+
+extension WebViewController: ThreeDSWKNavigationHelperDelegate {
+  func threeDSWKNavigationHelperDelegate(didReceiveResult result: Result<String, ThreeDSError>) {
+    navigationController?.popViewController(animated: true)
```
alex-ioja-yang-cko, line 87: Is there any chance this could be called from a background thread? Given the flow is relying on redirects, I think those redirects might be dispatched on threads other than main at some point, leading to the whole chain causing runtime warnings.

Personally I prefer enforcing main thread at the UI layer, when data is presented (here), but I have worked in teams where they'd want to enforce it at origin of the data (web view delegate updates?).

Checkout/Sample/Source/Card Tokenisation/Card Details/CardDetailsTableViewCell.swift
```diff
@@ -7,30 +7,129 @@
...
+  @IBOutlet weak var expiryDatePickerField: UITextField!
+
+  struct ViewModel {
+    var onCardNumberChange: ((String?) -> Void)?
+    var onCVVChange: ((String?) -> Void)?
+
+    var onExpiryMonthStringChange: ((String?) -> Void)?
+    var onExpiryYearStringChange: ((String?) -> Void)?
+
+    var onExpiryIntChange: ((_ month: Int, _ year: Int) -> Void)?
```
alex-ioja-yang-cko, line 29: Is this really a ViewModel? Being a collection of callbacks it feels like it wants to be a delegate. Also, same thing about hidden retain cycles applies again.

Checkout/Sample/Source/Card Tokenisation/Card Details/CardDetailsTableViewCell.swift
```diff
@@ -7,30 +7,129 @@
...
+    return dateComponents.date
+  }
+
+  weak var delegate: CardDetailsTableViewCellDelegate? {
+    didSet {
+      guard let delegate = delegate else {
+        return
+      }
+
+      cardNumberTextField.delegate = delegate.cardNumberDelegate
```
alex-ioja-yang-cko, line 90: It would be nice to keep code a little organised and move this property around other properties, potentially grouping by access control for quick discovery.

Also, if delegate is set to nil, shouldn't that also be passed to the `cardNumberTextField`? I'd think the guard is not required.

But good use of the property observer, I'd say this is the best way to use it, to pass a state change along to other dependencies, not to trigger logic flows.

Checkout/CocoapodsSample/CheckoutCocoapodsSample.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
+	<true/>
```
alex-ioja-yang-cko, line 6: Might want to add this file to gitignore.

Checkout/Checkout/Integration/CardValidatorTests.swift
```diff
@@ -0,0 +1,78 @@
...
+      XCTFail("expected success, received error: \(error)")
+    }
+  }
+
+  func test_expiryDate_thisMonth() {
+    let currentDate = Date()
+    let month = calendar.component(.month, from: currentDate)
+    let year = calendar.component(.year, from: currentDate)
+
+    let result = subject.validate(expiryMonth: month, expiryYear: year)
```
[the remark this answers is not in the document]
harry-brown-cko: we can add this as part of https://checkout.atlassian.net/browse/PIMOB-1080

Checkout/SPMSample/CheckoutSPMSample.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
+	<true/>
```
alex-ioja-yang-cko, line 6: Git ignore?

Checkout/Checkout/Integration/CardValidatorTests.swift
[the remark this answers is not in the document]
harry-brown-cko: we are already planning to do this as part of https://checkout.atlassian.net/browse/PIMOB-1080

Checkout/Sample/Source/3DS/WebViewController.swift
harry-brown-cko: Already exists, was just moving things. We can handle this in a follow up task

Checkout/Sample/Source/Card Tokenisation/Card Details/CardDetailsTableViewCell.swift
```diff
@@ -7,30 +7,129 @@
...
-    onCVVChange?(cvvTextField.text)
+    viewModel?.onCVVChange?(cvvTextField.text)
+  }
+
+  private enum ExpiryDateSegment: Int {
+    case text = 0
+    case dropdown = 1
+  }
+
+  @IBAction func expiryTypeChanged(_ sender: UISegmentedControl) {
```
alex-ioja-yang-cko, line 115:
```suggestion
  @IBAction private func expiryTypeChanged(_ sender: UISegmentedControl) {
```

Checkout/Sample/Source/Card Tokenisation/Card Details/CardDetailsTableViewCell.swift
```diff
@@ -7,30 +7,129 @@
...
   }
 
   @IBAction func cvvTextFieldEditingChanged(_ sender: Any) {
-    onCVVChange?(cvvTextField.text)
+    viewModel?.onCVVChange?(cvvTextField.text)
+  }
+
+  private enum ExpiryDateSegment: Int {
+    case text = 0
+    case dropdown = 1
```
alex-ioja-yang-cko, line 112: Move this to the top of the file so it doesn't get mixed with logic?

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
harry-brown-cko: Log is only called once, check the log manager

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
harry-brown-cko: Was copied over, lets handle this in a future task

Checkout/Sample/Source/Card Tokenisation/Card Details/ExpirationDatePicker.swift
```diff
@@ -0,0 +1,155 @@
...
+//  ExpirationDatePicker.swift
+//
+//
+//  Created by Harry Brown on 08/02/2022.
+//
+
+import UIKit
+
+class ExpirationDatePicker: UIPickerView {
+  private static var secondsInYear: TimeInterval = 31536000
```
alex-ioja-yang-cko, line 11: When doing something like this it may be easier to add the complete source of the number:
```suggestion
private static var secondsInYear: TimeInterval = 60 * 60 * 24 * 365
```

Checkout/Sample/Source/Card Tokenisation/Card Details/ExpirationDatePicker.swift
```diff
@@ -0,0 +1,155 @@
...
+
+  struct ViewModel {
+    var onDateChanged: ((_ month: Int, _ year: Int) -> Void)?
+  }
+
+  var viewModel: ViewModel?
+
+  private var months: [String] = []
+  private var years: [Int] = []
```
alex-ioja-yang-cko, line 20: These feel like lets? 

For months, you can use something like 
```
private let months = 1...12.map { formatter.string(from: 0) } // If you use a numberFormatter with formatter.minimumIntegerDigits = 2
private let months = 1...12.map { String(format: "%02d", $0) }
```

Similarly for year, I am not sure you need to create dates from minimum to maximum, then break them down and fetch the years. Something like this may also work and save you some properties
```
let years = 0...20.map { currentYear + $0 }
```

Checkout/Sample/Source/UIColor+CKO.swift
```diff
@@ -0,0 +1,31 @@
...
+  )
+
+  static let ckoLightBlue = UIColor(
+    red: 41 / 255,
+    green: 212 / 255,
+    blue: 219 / 255,
+    alpha: 1
+  )
+
+  static let ckoLightYellow = UIColor(
```
alex-ioja-yang-cko, line 25: I think there were some changes introduced by @ehab-al-cko , could we reuse the fame file?

Checkout/Checkout/Source/3DSHandling/URLHelper.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+
+import Foundation
+
+protocol URLHelping {
+  func urlsMatch(redirectUrl: URL, matchingUrl: URL) -> Bool
+  func extractToken(from url: URL) -> String?
+}
+
+final class URLHelper: URLHelping {
```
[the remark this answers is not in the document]
harry-brown-cko: looks like they were missed when copied over. just added them now

Checkout/Checkout/Source/3DSHandling/URLHelper.swift
```diff
@@ -0,0 +1,47 @@
...
+        guard queryItemIsPresent else {
+          return false
+        }
+      }
+    }
+
+    return redirectURLComponents.scheme == matchingURLComponents.scheme
+    && redirectURLComponents.host == matchingURLComponents.host
+    && redirectURLComponents.path == matchingURLComponents.path
+    && redirectURLComponents.fragment == matchingURLComponents.fragment
```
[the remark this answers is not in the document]
harry-brown-cko: Yes. On the success case of our 3DS the query value will have the session id and/or the token id appended. The absolute value will have changed but the URL is still what we want to intercept with

Checkout/Sample/Source/3DS/ThreeDSViewController.swift
```diff
@@ -0,0 +1,105 @@
...
+  @IBOutlet weak var urlChallengeButton: UIButton!
+  @IBOutlet weak var demoChallengeButton: UIButton!
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    setUpContent()
+  }
+
+  private func setUpContent() {
+    title = "3DS"
```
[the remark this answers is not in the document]
harry-brown-cko: we aren't currently localising this sample app

Checkout/Sample/Source/3DS/ThreeDSViewController.swift
```diff
@@ -0,0 +1,105 @@
...
+    let (title, message) = threeDSResultAlertInfo(from: result)
+    presentAlert(title: title, message: message)
+  }
+
+  private func threeDSResultAlertInfo(from result: Result<String, ThreeDSError>) -> (title: String, message: String) {
+    switch result {
+    case .success(let token):
+      return ("Success", token)
+    case .failure(let error):
+      return ("Failure", "ThreeDSError \(error.code)")
```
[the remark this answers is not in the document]
harry-brown-cko: we aren't currently localising this sample app

Checkout/Sample/Source/3DS/ThreeDSViewController.swift
```diff
@@ -0,0 +1,105 @@
...
+    return url
+  }
+
+  private func presentAlert(title: String, message: String) {
+    let alertController = UIAlertController(
+      title: title,
+      message: message,
+      preferredStyle: .alert
+    )
+    let action = UIAlertAction(title: "OK", style: .default, handler: nil)
```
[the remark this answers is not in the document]
harry-brown-cko: we aren't currently localising this sample app

Checkout/Sample/Source/3DS/ThreeDSViewController.swift
```diff
@@ -0,0 +1,105 @@
...
+
+  @IBAction func loadDemoChallenge(_ sender: Any) {
+    presentWebView(webViewConfig: .demo)
+  }
+
+  private func url(from textField: UITextField, name: String) -> URL? {
+    let url = textField.text.flatMap { URL(string: $0) }
+
+    if url == nil {
+      presentAlert(title: "Invalid value", message: "\(name) URL")
```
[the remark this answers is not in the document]
harry-brown-cko: we aren't currently localising this sample app

Checkout/Sample/Source/Card Tokenisation/Card Details/CardDetailsTableViewCell.xib
```diff
@@ -21,26 +21,36 @@
...
                                 <textInputTraits key="textInputTraits"/>
                                 <connections>
                                     <action selector="cardNumberTextFieldEditingChanged:" destination="KGk-i7-Jjw" eventType="editingChanged" id="eQI-jF-FJD"/>
                                 </connections>
                             </textField>
+                            <segmentedControl opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="left" contentVerticalAlignment="top" segmentControlStyle="plain" selectedSegmentIndex="0" translatesAutoresizingMaskIntoConstraints="NO" id="pLN-0h-2zN">
+                                <rect key="frame" x="0.0" y="55.5" width="288" height="48.5"/>
+                                <segments>
+                                    <segment title="Text Expiry"/>
+                                    <segment title="Dropdown Expiry"/>
```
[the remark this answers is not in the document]
harry-brown-cko: we aren't currently localising this sample app

Checkout/SPMSample/CheckoutSPMSample.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,16 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "CheckoutEventLoggerKit",
+        "repositoryURL": "https://github.com/checkout/checkout-event-logger-ios-framework.git",
+        "state": {
+          "branch": null,
+          "revision": "2ef34662f1525526aeb5681c6af5f69572ab36c9",
+          "version": "1.2.0"
```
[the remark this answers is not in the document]
harry-brown-cko: Disagree here. The Package.swift can include minimal info on versions and allow auto updates if they aren't locked. We should commit this in the same way that we commit our Podfile.lock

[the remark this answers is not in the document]
alex-ioja-yang-cko: I would say that this is just a reflection of the latest update from the package dependencies.

If your Package.swift dependency declaration says use latest release, this will update when a new release was made available.

The impact of it will be:
- no effect on restricting the version you use, that comes only from `Package.swift` dependencies declaration
- getting committed constantly without relevance. You may be changing localisations, but the fact that your SPM is resolving the dependencies will update the source controlled file and carry it into your PR if anything was changed, although you haven't technically looked into any differences / impact
- pretty unlikely at this team size, but could lead to merge conflicts. If we have an internal package that has internal dependencies and uses `main` branch (lets say on the SecureDisplay repo I showed yesterday), branches working asynchronously can have different revisions resolved. This can then lead to open PRs that have no overlap but will still create a merge conflict after first PR is merged
- ➕ it does offer at repository level a clear way to view what versions were resolved

This is my interpretation of it, and I would say that it differs from `Podfile.lock` as the pod lock gets updated only when you run a pod install / update, not part of Xcode system build. Packages on the other hand are updated every time you open Xcode and occasionally throughout your use of it. I don't think you can prevent Xcode from resolving Package.swift file as it often times needs to clone dependencies locally and check if they're still in sync (unlike cocoa pods)

Checkout/Sample/Source/3DS/ThreeDSViewController.swift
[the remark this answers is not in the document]
alex-ioja-yang-cko: I won't mind in the future if you dismiss all these comments at the same time. I now see its inside the `Sample` folder, but didn't notice it originally. 

Sorry for the spam :)

Checkout/Sample/Source/Card Tokenisation/Card Details/ExpirationDatePicker.swift
harry-brown-cko: Yeah makes sense, I've done a bit of cleanup here. the ExpirationDatePicker was originally from the old Frames UI. I originally put minimal work into refactoring because it would only be used in the sample app

Checkout/Sample/Source/UIColor+CKO.swift
harry-brown-cko: Wouldn't be available in this context, those colours are in the Frames SDK

Checkout/Checkout/Test/Stubs/StubWKNavigationAction.swift
```diff
@@ -0,0 +1,14 @@
...
+//  StubWKNavigationAction.swift
+//
+//
+//  Created by Harry Brown on 21/02/2022.
+//
+
+import WebKit
+
+final class StubWKNavigationAction: WKNavigationAction {
+  // swiftlint:disable:next force_unwrapping
```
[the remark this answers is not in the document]
deepesh-vasthimal-cko: we did discuss we still want SwiftLint to have basic formatting and not go out of control when compare test to our  existing code base. For some places we can use ignore when we have force_unwrapping which is used in tests.

Checkout/Checkout/Test/Validation/CardNumberValidatorTests.swift
```diff
@@ -83,7 +83,85 @@ final class CardNumberValidatorTests: XCTestCase {
...
+        actualResult,
+        expectedResult,
+        "expected \(expectedResult) for card number \(cardNumber), received \(actualResult)"
+      )
+
+      XCTAssertEqual(stubLuhnChecker.luhnCheckCalledWith, nil)
+    }
+  }
+
+  func test_eagerValidateCardNumber_multipleValidLengths() {
```
deepesh-vasthimal-cko, line 97: Can we have invalid values which are copy pasted garbage entries like below as one more test ?  

 "-42": .success(.visa),
      "$%424": .success(.visa),
      "werewr4242": .success(.visa),
      "42424-": .success(.visa),
      "424242asdf": .success(.visa),

Checkout/Checkout/Integration/CardValidatorTests.swift
```diff
@@ -0,0 +1,78 @@
...
+//
+
+import XCTest
+import Checkout
+
+final class CardValidatorTests: XCTestCase {
+  private let subject = CardValidator(environment: .sandbox)
+  private let calendar = Calendar(identifier: .gregorian)
+
+  func test_expiryDate_2040() {
```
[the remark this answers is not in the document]
harry-brown-cko: I might just do this now. Would be helpful to have it sorted for other tickets

deepesh-vasthimal-cko (commented):
few minor comments otherwise LGTM. 

You still have comments from @alex-ioja-yang-cko please have a look at them.

Checkout/Sample/Source/Card Tokenisation/Card Details/ExpirationDatePicker.swift
```diff
@@ -0,0 +1,155 @@
...
+//
+
+import UIKit
+
+class ExpirationDatePicker: UIPickerView {
+  private static var secondsInYear: TimeInterval = 31536000
+
+  struct ViewModel {
+    var onDateChanged: ((_ month: Int, _ year: Int) -> Void)?
+  }
```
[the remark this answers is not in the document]
harry-brown-cko: This comes from copied code. I think this is an issue across the sample app that can be sorted at a later date

Checkout/Checkout/Source/Model/Card+Scheme.swift
```diff
@@ -21,33 +21,33 @@ extension Card {
...
     case maestro
-    case amex
+    case americanExpress
     case discover
-    case diners
+    case dinersClub
     case jcb
 
     public init?(rawValue: String) {
       switch rawValue.lowercased() {
```
ehab-al-cko, line 34:
```suggestion
      switch rawValue
                   .lowercased()
                   .trimmingCharacters(in: .whitespacesAndNewlines) {
```

ehab-al-cko, line 34:
```suggestion
      switch rawValue
                   .lowercased()
                   .trimmingCharacters(in: .whitespacesAndNewlines) {
```

Checkout/Checkout/Source/Model/Card+Scheme.swift
```diff
@@ -21,33 +21,33 @@ extension Card {
...
       case "unknown":
         self = .unknown
+      case "mada", "visa - mada", "mastercard - mada":
```
ehab-al-cko, line 37: after trimming white spaces
```suggestion
      case "mada", "visa-mada", "mastercard-mada":
```

ehab-al-cko, line 37: after trimming white spaces
```suggestion
      case "mada", "visa-mada", "mastercard-mada":
```

Checkout/Checkout/Source/Model/Card+Scheme.swift
```diff
@@ -111,19 +115,40 @@ extension Card {
...
+        .visa,
+        .mastercard:
+        return 16
+      case .americanExpress:
+        return 15
+      case .maestro,
+        .discover,
+        .dinersClub,
+        .jcb:
+        return 19
```
ehab-al-cko, line 148: The following code, if `scheme.maxCardLength` is `nil` then use `Card.Scheme.maxCardLengthAllSchemes` 

```  
private func withinCardLengthLimit(_ cardNumber: String, for scheme: Card.Scheme) -> Bool {
    let maxCardLength = scheme.maxCardLength ?? Card.Scheme.maxCardLengthAllSchemes
    return cardNumber.count <= maxCardLength
 }
```
so i suggest, the max is always 19 

```suggestion
        return maxCardLengthAllSchemes
```

ehab-al-cko, line 148:
```suggestion
        return maxCardLengthAllSchemes
```

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
```diff
@@ -32,9 +33,32 @@ final class CardNumberValidator: CardNumberValidating {
       return .success(cardScheme)
     case .none:
 
-      let cardScheme = cardTypeMatch(cardNumber, \.eagerCardNumberRegex) ?? .unknown
-      return .success(cardScheme)
+      return .success(.unknown)
```
ehab-al-cko, line 36: I am wondering, can validation be successful with `unkown` schema?

ehab-al-cko, line 36: I am wondering, can validation be successful with `unkown` schema?

Checkout/Checkout/Source/Model/Card+Scheme.swift
```diff
@@ -111,19 +115,40 @@ extension Card {
...
       }
     }
 
+    /// maximum card length for any scheme
+    static let maxCardLengthAllSchemes = 19
+
+    var maxCardLength: Int? {
+      switch self {
+      case .unknown:
+        return nil
```
ehab-al-cko, line 137:
```suggestion
        return maxCardLengthAllSchemes
```

ehab-al-cko, line 137:
```suggestion
        return maxCardLengthAllSchemes
```

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
```diff
@@ -32,9 +33,32 @@ final class CardNumberValidator: CardNumberValidating {
...
+    case .some(let cardScheme):
+      return withinCardLengthLimit(cardNumber, for: cardScheme) ? .success(cardScheme) : .failure(.tooLong)
+    case .none:
+      return shouldHaveFoundScheme(cardNumber) ? .failure(.invalidScheme) : .success(.unknown)
+    }
+  }
+
+  private func withinCardLengthLimit(_ cardNumber: String, for scheme: Card.Scheme) -> Bool {
+    let maxCardLength = scheme.maxCardLength ?? Card.Scheme.maxCardLengthAllSchemes
+    return cardNumber.count <= maxCardLength
```
ehab-al-cko, line 57: This is a follow-up comment on the previous related comment. 
```suggestion
    cardNumber.count <= scheme.maxCardLength
```

ehab-al-cko, line 57: This is a follow-up comment on the previous related comment. 
```suggestion
    cardNumber.count <= scheme.maxCardLength
```

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
```diff
@@ -95,10 +104,20 @@ public class CardValidator: CardValidating {
...
+
+    switch fourDigitYear(from: expiryYear) {
+    case .success(let expiryYear):
+      return validateExpiry(
+        expiryMonth: expiryMonth,
+        expiryYear: expiryYear
+      )
+    case .failure(let error):
+      return .failure(error)
+    }
```
ehab-al-cko, line 120: Is the same code logic like the block code from line 131 to 146?

ehab-al-cko, line 120: Is the same code logic like the block code from line 131 to 146?

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
```diff
@@ -178,63 +203,69 @@ public class CardValidator: CardValidating {
...
-  ) -> Result<Int, ValidationError.ExpiryDate> {
+  private func fourDigitYear(from expiryYear: String) -> Result<Int, ValidationError.ExpiryDate> {
     switch expiryYear.count {
-    case 0, 1, 3:
-      return .failure(.incompleteYear)
-    case 2:
-      return .success(2000 + year)
-    case 4:
-      return .success(year)
+    case 0...4:
```
ehab-al-cko, line 238: please add a comment for this line with example.

ehab-al-cko, line 238: please add a comment for this line with example.

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
```diff
@@ -178,63 +203,69 @@ public class CardValidator: CardValidating {
...
     default:
       return .failure(.invalidYear)
     }
   }
+
+  private func fourDigitYear(from expiryYear: Int) -> Result<Int, ValidationError.ExpiryDate> {
+    // year must be 2 or 4+ digits, otherwise year is incomplete
+    // 0 digits - no year has been entered yet
+    // 1 or 3 digits - we are waiting on an extra digit
+    guard expiryYear > 1000 || (expiryYear > 9 && expiryYear < 100) else {
```
ehab-al-cko, line 253: please add example for this `expiryYear > 9 && expiryYear < 100` and why starts after `9`?

ehab-al-cko, line 253: please add example for this `expiryYear > 9 && expiryYear < 100` and why starts AFTER `9`?

ehab-al-cko (commented):
Good work :) 
some questions and minor changes.
If you have time, UI test will be nice to cover normal scenarios.

Checkout/Checkout/Source/3DSHandling/ThreeDSWKNavigationHelper.swift
```diff
@@ -0,0 +1,69 @@
...
+//  ThreeDSWKNavigationHelper.swift
+//  Checkout
+//
+//  Created by Daven.Gomes on 02/02/2022.
+//
+
+import UIKit
+import WebKit
+
+public class ThreeDSWKNavigationHelper: NSObject {
```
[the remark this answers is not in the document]
harry-brown-cko: resolving as duplicate

Checkout/Checkout/Source/3DSHandling/URLHelper.swift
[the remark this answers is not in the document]
harry-brown-cko: query is not guaranteed to be identical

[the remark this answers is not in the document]
harry-brown-cko: will refactor to make this more obvious

Checkout/Checkout/Source/3DSHandling/ThreeDSWKNavigationHelper.swift
```diff
@@ -0,0 +1,69 @@
...
+      let successURL = successURL,
+      urlHelper.urlsMatch(redirectUrl: redirectUrl, matchingUrl: successURL)
+    {
+      guard let token = urlHelper.extractToken(from: redirectUrl) else {
+        delegate.threeDSWKNavigationHelperDelegate(didReceiveResult: .failure(.couldNotExtractToken))
+        return .cancel
+      }
+
+      delegate.threeDSWKNavigationHelperDelegate(didReceiveResult: .success(token))
+      return .cancel
```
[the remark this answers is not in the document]
harry-brown-cko: We should be cancelling here. We've got the success or failure URL so there's no point in allowing the webview to continue loading

Checkout/Checkout/Source/3DSHandling/ThreeDSWKNavigationHelper.swift
```diff
@@ -0,0 +1,69 @@
...
+extension ThreeDSWKNavigationHelper: WKNavigationDelegate {
+  public func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
+    let navigationDecision = navigationAction.request.url.map { decision(from: $0) } ?? .allow
+
+    decisionHandler(navigationDecision)
+  }
+
+  private func decision(from redirectUrl: URL) -> WKNavigationActionPolicy {
+    guard let delegate = delegate else {
+      return .allow
```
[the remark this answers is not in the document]
harry-brown-cko: Depends what we want the fallback behaviour to be. `.cancel` would stop the webview from working without the delegate, `.allow` lets a user complete the challenge but the intercept would be broken

Checkout/Checkout/Source/Model/Card+Scheme.swift
harry-brown-cko: `trimmingCharacters(in: .whitespacesAndNewlines)` only removes whitespace at the start and end. `"visa - mada"` would still be `"visa - mada"`

Checkout/Checkout/Source/Model/Card+Scheme.swift
harry-brown-cko: our API always returns the scheme without leading or trailing whitespace, did you instead mean `.filter { !.$0.isWhitespace }`

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
harry-brown-cko: Yeah I think this needs to be discussed again with the team. `.unknown` is returned when a string of digits has no regex match, which is confusing for merchants. This should be an error

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
harry-brown-cko: yeah looks duplicated to some degree. have cleaned up

ehab-al-cko (approved):

alex-ioja-yang-cko (approved):
I see no changes to git ignore, so I guess you just deleted the files?

harry-brown-cko:
> I see no changes to git ignore, so I guess you just deleted the files?

Yes. I think they are already ignored and they were added whilst trying to fix the Carthage stuff

PR #164, opened 2022-06-01
Add Optional label to Billing form cell

ehab-al-cko:
## Proposed changes

 Add Optional label to billing form cell

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/Core/Constants+LocalizationKeys.swift
+++ b/Source/Core/Constants+LocalizationKeys.swift
@@ -4,6 +4,10 @@ extension Constants {
 
         struct BillingForm {
             
+            struct Cell {
+                static let isOptional = "isOptional".localized
+            }
+
             struct AddressLine1 {
                 static let title = "addressLine1".localized
                 static let error = "missingBillingFormAddressLine1".localized
--- a/Source/Core/Constants+Style.swift
+++ b/Source/Core/Constants+Style.swift
@@ -40,6 +40,10 @@ extension Constants {
                 case fontSize = 15
             }
 
+            enum InputISOptionalLabel: Double {
+                case fontSize = 13
+            }
+
             enum InputCountryButton: Double {
                 case height = 56
                 case width = 0
--- a/Source/UI/NewUI/CommonUI/View/TextFieldView.swift
+++ b/Source/UI/NewUI/CommonUI/View/TextFieldView.swift
@@ -26,7 +26,15 @@ class TextFieldView: UIView {
         view.backgroundColor = .white
         return view
     }()
-    
+
+    private(set) lazy var isOptionalLabel: UILabel? = {
+        let view = UILabel()
+        view.numberOfLines = 1
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.backgroundColor = .white
+        return view
+    }()
+
     private(set) lazy var hintLabel: UILabel? = {
         let view = UILabel()
         view.numberOfLines = 0
@@ -69,6 +77,7 @@ class TextFieldView: UIView {
         self.style = style
         setupViewsInOrder()
         updateHeaderLabel(style: style)
+        updateIsOptionalLabel(style: style)
         updateHintLabel(style: style)
         updateTextFieldContainer(style: style)
         updateTextField(style: style, textFieldValue: textFieldValue)
@@ -79,6 +88,13 @@ class TextFieldView: UIView {
         headerLabel?.text = style.title?.text
         headerLabel?.font = style.title?.font
         headerLabel?.textColor = style.title?.textColor
+    }
+
+    private func updateIsOptionalLabel(style: CellTextFieldStyle) {
+        isOptionalLabel?.isHidden = !style.isOptional
+        isOptionalLabel?.text = Constants.LocalizationKeys.BillingForm.Cell.isOptional
+        isOptionalLabel?.font = UIFont(graphikStyle: .regular, size:  Constants.Style.BillingForm.InputISOptionalLabel.fontSize.rawValue)
+        isOptionalLabel?.textColor = .doveGray
     }
 
     private func updateHintLabel(style: CellTextFieldStyle) {
@@ -125,6 +141,7 @@ extension TextFieldView {
     private func setupViewsInOrder(){
         backgroundColor = style?.backgroundColor
         setupHeaderLabel()
+        setupIsOptionalLabel()
         setupHintLabel()
         setupTextFieldContainer()
         setupTextField()
@@ -137,7 +154,17 @@ extension TextFieldView {
         NSLayoutConstraint.activate([
             headerLabel.topAnchor.constraint(equalTo: topAnchor),
             headerLabel.leadingAnchor.constraint(equalTo: leadingAnchor),
-            headerLabel.trailingAnchor.constraint(equalTo: trailingAnchor)
+        ])
+    }
+
+    private func setupIsOptionalLabel() {
+        guard let isOptionalLabel = isOptionalLabel else { return }
+        guard let headerLabel = headerLabel else { return }
+        addSubview(isOptionalLabel)
+        NSLayoutConstraint.activate([
+            isOptionalLabel.topAnchor.constraint(equalTo: topAnchor),
+            isOptionalLabel.trailingAnchor.constraint(equalTo: trailingAnchor),
+            isOptionalLabel.leadingAnchor.constraint(greaterThanOrEqualTo: headerLabel.trailingAnchor)
         ])
     }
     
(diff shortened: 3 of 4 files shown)


Source/Core/Constants+Style.swift
```diff
@@ -40,6 +40,10 @@ extension Constants {
                 case fontSize = 15
             }
 
+            enum InputISOptionalLabel: Double {
```
alex-ioja-yang-cko, line 43: Whenever there's a verb, sounds like stuff going on. We could rename to `InputOptionalLabel` or `OptionalInputLabel` and remove that fake expectation.

I also suggested separately to look at content not in terms of _is it optional_, but rather _is it compulsory_. Personally I find that optional has many connotations (including in Swift), so I would avoid the ambiguity by checking if its required / compulsory (maybe more personal preference so feel free to dismiss this last part if disagree).

Source/Core/Constants+LocalizationKeys.swift
```diff
@@ -4,6 +4,10 @@ extension Constants {
 
         struct BillingForm {
             
+            struct Cell {
+                static let isOptional = "isOptional".localized
```
alex-ioja-yang-cko, line 8: Is this key already present? I can't see it on my version anywhere else, so I think its new?

Source/UI/NewUI/CommonUI/View/TextFieldView.swift
```diff
@@ -26,7 +26,15 @@ class TextFieldView: UIView {
         view.backgroundColor = .white
         return view
     }()
-    
+
+    private(set) lazy var isOptionalLabel: UILabel? = {
+        let view = UILabel()
+        view.numberOfLines = 1
```
alex-ioja-yang-cko, line 32: This property is normally 1 by default which usually causes a lot of issues. Might want to set it to 0 instead, as we don't know how the text might scale in different localisations & device/font sizes.

Source/UI/NewUI/CommonUI/View/TextFieldView.swift
```diff
@@ -81,6 +90,13 @@ class TextFieldView: UIView {
         headerLabel?.textColor = style.title?.textColor
     }
 
+    private func updateIsOptionalLabel(style: CellTextFieldStyle) {
+        isOptionalLabel?.isHidden = !style.isOptional
+        isOptionalLabel?.text = Constants.LocalizationKeys.BillingForm.Cell.isOptional
+        isOptionalLabel?.font = UIFont(graphikStyle: .regular, size:  Constants.Style.BillingForm.InputISOptionalLabel.fontSize.rawValue)
+        isOptionalLabel?.textColor = .doveGray
```
alex-ioja-yang-cko, line 97: It seems to me the only `update` is the hidden state. Everything else should be a `setup`?

Source/UI/NewUI/CommonUI/View/TextFieldView.swift
```diff
@@ -26,7 +26,15 @@ class TextFieldView: UIView {
         view.backgroundColor = .white
         return view
     }()
-    
+
+    private(set) lazy var isOptionalLabel: UILabel? = {
```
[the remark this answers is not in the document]
alex-ioja-yang-cko: Also as separately mentioned (and again feel free to disagree and reject), thinking in terms of `compulsory` may be clearer.

alex-ioja-yang-cko (commented):
I think it would be useful to have a snapshot of the design spec for this in the PR description. Just as I can't stop thinking in some way that a NSAttributedString may also work for the requirement, but the UI layout code would tell me it wouldn't. Having that image attached there would remove the ambiguity.

Source/UI/NewUI/CommonUI/View/TextFieldView.swift
```diff
@@ -81,6 +90,13 @@ class TextFieldView: UIView {
         headerLabel?.textColor = style.title?.textColor
     }
 
+    private func setupMandatoryLabel(style: CellTextFieldStyle) {
+        mandatoryLabel?.isHidden = style.isMandatory
+        mandatoryLabel?.text = Constants.LocalizationKeys.BillingForm.Cell.optionalInput
+        mandatoryLabel?.font = UIFont(graphikStyle: .regular, size:  Constants.Style.BillingForm.InputOptionalLabel.fontSize.rawValue)
+        mandatoryLabel?.textColor = .doveGray
+    }
```
alex-ioja-yang-cko, line 98: Sorry, I think you misunderstood my previous suggestion. I was not saying the method should be renamed from `update` to `setup`, I was suggesting that 3 out of these 4 lines should be moved into the existing `setupMandatoryLabel`, and only the visibility should be updated (which may even remove the need for this method altogether.

Also I wouldn't want to have 2 methods named identically, once taking a parameter, one without. `function overloading` definitely is a useful feature of Swift, just not sure this is a good use case for it.

Tests/UI/New UI/BillingForm/View/BillingFormButtonViewTests.swift
```diff
@@ -29,12 +29,6 @@ class BillingFormButtonViewTests: XCTestCase {
         XCTAssertEqual(view.image?.tintColor, style.button.disabledTintColor)
     }
 
-    func testButtonStyle() {
-        XCTAssertEqual(view.button?.isEnabled, style.button.isEnabled)
-        XCTAssertEqual(view.button?.tintColor, .clear)
-        XCTAssertEqual(view.button?.backgroundColor, .clear)
-    }
```
alex-ioja-yang-cko, line 36: Do we intentionally want to remove unit tests?

Source/UI/NewUI/CommonUI/View/TextFieldView.swift
```diff
@@ -27,6 +27,14 @@ class TextFieldView: UIView {
         return view
     }()
 
+    private(set) lazy var mandatoryLabel: UILabel? = {
+        let view = UILabel()
+        view.numberOfLines = 0
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.backgroundColor = .white
```
alex-ioja-yang-cko, line 34: I would avoid hardcoding these colours, especially as you will need them customised and presumably the label doesn't require a separate background colour for itself only (haven't seen a design so may as well be).

alex-ioja-yang-cko (approved):
Are you working on multiple machines? I find it confusing that the code styling can change so often between different commits / PRs, is there anything we can help you with to make this easier?

Also, I think there are a few small details to review, but nothing groundbreaking imo (would encourage extending testing, not removing it though ! ).

Source/UI/NewUI/CommonUI/View/TextFieldView.swift
ehab-al-cko: Fixed in this [PR](https://github.com/checkout/frames-ios/pull/168)

Tests/UI/New UI/BillingForm/View/BillingFormButtonViewTests.swift
ehab-al-cko: Duplicate test cases

Source/Resources/en.lproj/Localizable.strings
```diff
@@ -9,6 +9,7 @@
 "billingFormPhoneNumberHint" = "Phone Number";
 "billingAddressTitle" = "Billing address";
 "BillingFormCountryButtonTitle" = "Please select a country";
+"BillingFormOptional" = "Optional";
```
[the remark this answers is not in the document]
deepesh-vasthimal-cko: so for now these strings are for only english as we have got them from Alex ( designer ). we will then add more languages support, if we don't provide localisation for other languages but try to run the app in other language or region they are defaulted to english, Nothing would break we just dont support, but with New Design changes we don't need to release a new SDK all the time to support languages , merchants can inject the localised string themselves by using our customisation options.

Source/UI/NewUI/CommonUI/View/TextFieldView.swift
```diff
@@ -137,7 +154,17 @@ extension TextFieldView {
...
-            headerLabel.trailingAnchor.constraint(equalTo: trailingAnchor)
+        ])
+    }
+
+    private func setupMandatoryLabel() {
+        guard let mandatoryLabel = mandatoryLabel else { return }
+        guard let headerLabel = headerLabel else { return }
+        addSubview(mandatoryLabel)
+        NSLayoutConstraint.activate([
+            mandatoryLabel.topAnchor.constraint(equalTo: topAnchor),
```
deepesh-vasthimal-cko, line 165: I am assuming leadingAnchor is not required here as the super view does that ?. I am concerned if this mandatoryLabel will support RTL.

ehab-al-cko: Tested and working fine
Screenshot 2022-06-09 at 2 31 33 pm

PR #166, opened 2022-06-08
Frames SDK API specs comments

ehab-al-cko:
## Proposed changes

Screenshot 2022-06-08 at 6 41 37 pm

### DIFF
--- a/Source/UI/NewUI/BillingForm/Factory/BillingFormFactory.swift
+++ b/Source/UI/NewUI/BillingForm/Factory/BillingFormFactory.swift
@@ -18,7 +18,7 @@ public struct BillingFormFactory {
          .phoneNumber(DefaultBillingFormPhoneNumberCellStyle())]
     }
 
-    static func getBillingFormViewController(style: BillingFormStyle?, data: BillingForm?, delegate: BillingFormViewModelDelegate) -> (BillingFormViewModelDelegate?, UINavigationController)? {
+    static func getBillingFormViewController(style: BillingFormStyle?, data: BillingForm?, delegate: BillingFormViewModelDelegate) -> (deleagte: BillingFormViewModelDelegate?, navigationController: UINavigationController)? {
 
         guard let style = style, !style.cells.isEmpty else { return nil }
 
--- a/Source/UI/NewUI/BillingForm/Factory/BillingFormFactory.swift
+++ b/Source/UI/NewUI/BillingForm/Factory/BillingFormFactory.swift
@@ -18,7 +18,7 @@ public struct BillingFormFactory {
          .phoneNumber(DefaultBillingFormPhoneNumberCellStyle())]
     }
 
-    static func getBillingFormViewController(style: BillingFormStyle?, data: BillingForm?, delegate: BillingFormViewModelDelegate) -> (deleagte: BillingFormViewModelDelegate?, navigationController: UINavigationController)? {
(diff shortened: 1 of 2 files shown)


Source/UI/NewUI/BillingForm/Factory/BillingFormFactory.swift
```diff
@@ -18,7 +18,7 @@ public struct BillingFormFactory {
          .phoneNumber(DefaultBillingFormPhoneNumberCellStyle())]
     }
 
-    static func getBillingFormViewController(style: BillingFormStyle?, data: BillingForm?, delegate: BillingFormViewModelDelegate) -> (BillingFormViewModelDelegate?, UINavigationController)? {
+    static func getBillingFormViewController(style: BillingFormStyle?, data: BillingForm?, delegate: BillingFormViewModelDelegate) -> (deleagte: BillingFormViewModelDelegate?, navigationController: UINavigationController)? {
```
alex-ioja-yang-cko, line 21:
```suggestion
    static func getBillingFormViewController(style: BillingFormStyle?, data: BillingForm?, delegate: BillingFormViewModelDelegate) -> (delegate: BillingFormViewModelDelegate?, navigationController: UINavigationController)? {
```

alex-ioja-yang-cko: I'm late for this, but why are we passing in a delegate and then also giving it back out. Doesn't the caller already own it since it provided it in the first place?

deepesh-vasthimal-cko: Agree with @alex-ioja-yang-cko , looking at the code i see getBillingFormViewController takes in a delegate which is of the view controller where the billingVC is presented. It is then passed to DefaultBillingFormViewModel where it has the delegate methods and sets this delegate. 

But from the sample app when we call getBillingFormViewController as merchant it is never used inside the VC itself ( when returned ) and we can get rid of the returned delegate if we are not using it.

deepesh-vasthimal-cko (approved):
LGTM one comment about the delegate which is returned never used for getBillingFormViewController, Please have a look.

Source/UI/NewUI/BillingForm/Factory/BillingFormFactory.swift
harry-brown-cko: we should address this and/or fix the typo

harry-brown-cko (commented):
one comment

Source/UI/NewUI/BillingForm/Factory/BillingFormFactory.swift
ehab-al-cko: It was a comment on old PR to expose [delegate](https://github.com/checkout/frames-ios/pull/144#discussion_r860643734)

alex-ioja-yang-cko: I think you misinterpreted what that comment asked for. It wasn't for a delegate to be returned alongside the VC, it wouldn't make sense to if you have no need for it.

I was saying that the common syntax is to assign the delegate **after** the object has been created, not during the initialisation. The common way to see delegate being assigned is in patterns of:
```swift
let object = MyObject()
object.delegate = myDelegate
```

rather than

```swift
let object = MyObject(delegate: myDelegate)
```

There are some nuances to why that is, which are not applicable here, but make it a more regular occurance when navigating the code. I would have rather you rejected this idea than add this code change :).

alex-ioja-yang-cko (approved):
I do think that excess complication in the tuple being returned should get removed, but don't mind if done separately, up to you.

harry-brown-cko (approved):

PR #168, opened 2022-06-08
Fix: Change the background colour of the whole page

ehab-al-cko:
[Jira Ticket ](https://checkout.atlassian.net/browse/PIMOB-1240)

## Proposed changes

Change the background colour of the whole page
Be able to change the background colour of the whole billing form page in one go,  ( Currently  BillingFormStyle’s mainBackground only changes a part of the background as the table view, cells, components background overlaps and does not allow to change the background, By default the table view, cells, components background opacity should be 0. If a merchant wishes to change this should be able to change it via individual components which overrides the mainBackground color of BillingFormStyle.

## Types of changes

## Checklist

* [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/Source/UI/NewUI/BillingForm/View/BillingFormTextField.swift
+++ b/Source/UI/NewUI/BillingForm/View/BillingFormTextField.swift
@@ -6,6 +6,7 @@ class BillingFormTextField: UITextField {
     init(type: BillingFormCell?, tag: Int) {
         self.type = type
         super.init(frame: .zero)
+        backgroundColor = .clear
         self.tag = tag
     }
 
--- a/Source/UI/NewUI/CommonUI/View/TableViewCell/CellButton.swift
+++ b/Source/UI/NewUI/CommonUI/View/TableViewCell/CellButton.swift
@@ -18,6 +18,7 @@ final class CellButton: UITableViewCell {
 
     override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
         super.init(style: style, reuseIdentifier: reuseIdentifier)
+        backgroundColor = .clear
         self.setupViewsInOrder()
     }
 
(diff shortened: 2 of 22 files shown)


Source/UI/NewUI/CommonUI/View/TableViewCell/CellButton.swift
```diff
@@ -18,6 +18,7 @@ final class CellButton: UITableViewCell {
 
     override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
         super.init(style: style, reuseIdentifier: reuseIdentifier)
+        backgroundColor = .clear
```
alex-ioja-yang-cko, line 21: I am not sure this is needed?

Source/UI/NewUI/BillingForm/View/BillingFormTextField.swift
```diff
@@ -6,6 +6,7 @@ class BillingFormTextField: UITextField {
     init(type: BillingFormCell?, tag: Int) {
         self.type = type
         super.init(frame: .zero)
+        backgroundColor = .clear
```
alex-ioja-yang-cko, line 9: This is nil by default, so I am not sure we ever need to set it to clear?

alex-ioja-yang-cko (commented):
Most of the changes you did are not actually required. By default most of those components start with `nil` backgroundColor (famously not the tableView). I think a better solution would be to remove those lines of code instead of providing a value we don't actually need.

Source/UI/NewUI/BillingForm/View/BillingFormTextField.swift
deepesh-vasthimal-cko: does `nil` means same as `clear` ? if then makes sense, otherwise can stay `clear` does not harm. I think the overall output will be same provided one line of extra code which can be removed.

harry-brown-cko (commented):
would prefer to either not set backgroundColor or set it to nil as default for implementations. Should keep `.clear` for style structs so that the type can stay as `UIColor`

Source/UI/NewUI/BillingForm/View/BillingFormTextField.swift
ehab-al-cko: I have tested removing this line then background is `.white`. For now, Gonna keep it `.clear`

Source/UI/NewUI/CommonUI/View/TableViewCell/CellButton.swift
ehab-al-cko: I have tested removing this line then background is .white. For now, Gonna keep it .clear

harry-brown-cko (approved):

PR #170, opened 2022-06-09
PIMOB-1240: Fix billing form cells order

ehab-al-cko:
[Jira ticket ](https://checkout.atlassian.net/browse/PIMOB-1240)
## Proposed changes

Add country into the correct field when the fields passed in BillingFormCell are changes

Issue: 
When I add a country in, it should always go in the country field even if I change how many fields are passed in BillingFormCell
BillingFormFactory.defaultBillingFormStyle contains 8 BillingFormCell injected, if this is changed to 7 the editing of each text field value is seen to be impacted and interchanged which is not expected. 

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/UI/Controllers/AddressViewController.swift
+++ b/Source/UI/Controllers/AddressViewController.swift
@@ -181,8 +181,8 @@ public class AddressViewController: UIViewController,
 
     // MARK: - CountrySelectionViewControllerDelegate
 
-    /// Executed when an user select a country.
-    public func onCountrySelected(country: Country) {
+    /// Executed when a user select a country.
+    public func onCountrySelected(country: Country, tag: Int) {
         setCountrySelected(country: country)
     }
 
--- a/Source/UI/Controllers/AddressViewController.swift
+++ b/Source/UI/Controllers/AddressViewController.swift
@@ -182,7 +182,7 @@ public class AddressViewController: UIViewController,
     // MARK: - CountrySelectionViewControllerDelegate
 
     /// Executed when a user select a country.
-    public func onCountrySelected(country: Country, tag: Int) {
+    public func onCountrySelected(country: Country) {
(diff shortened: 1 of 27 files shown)


Source/UI/Controllers/AddressViewController.swift
```diff
@@ -181,8 +181,8 @@ public class AddressViewController: UIViewController,
 
     // MARK: - CountrySelectionViewControllerDelegate
 
-    /// Executed when an user select a country.
-    public func onCountrySelected(country: Country) {
+    /// Executed when a user select a country.
+    public func onCountrySelected(country: Country, tag: Int) {
```
deepesh-vasthimal-cko, line 185: does this needs to be public ?

deepesh-vasthimal-cko (approved):

Source/UI/Controllers/AddressViewController.swift
ehab-al-cko: This is old billing form UI. I think, it will be removed after approving new billing form UI

Source/UI/Delegates/CountrySelectionViewControllerDelegate.swift
```diff
@@ -7,5 +7,5 @@ public protocol CountrySelectionViewControllerDelegate: AnyObject {
     ///
     /// - parameter country: Country selected
     /// - parameter regionCode: Region code of the selected country
-    func onCountrySelected(country: Country)
+    func onCountrySelected(country: Country, tag: Int)
```
harry-brown-cko, line 10: we have the ordered list of cells from `DefaultBillingFormViewModel.init`. Does it make more sense to pass around the updated row index or can we rely on the order from there instead?

harry-brown-cko (approved):

PR #176, opened 2022-06-15
Fix country region to be device local region

ehab-al-cko:
## Proposed changes

When `iso3166Alpha2` is not set by the merchant then use local device region 

## Types of changes

## Checklist

* [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/Source/UI/NewUI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
+++ b/Source/UI/NewUI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
@@ -18,8 +18,8 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
     var errorFlagOfCellType = [Int: Bool]()
     var textValueOfCellType = [Int: String]()
 
-    private var country : Country?
     private var countryRow: Int?
+    private(set) var country : Country?
     private(set) var style: BillingFormStyle
     private(set) var data: BillingForm?
     private(set) var updatedRow: Int? {
@@ -73,6 +73,9 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
             country = data?.address.country
             textValueOfCellType[type.index] = hasValue ? value : (!isMandatory ? "" : nil)
             errorFlagOfCellType[type.index] = hasValue ? false : isMandatory
+        }
+        if let country = country, country.iso3166Alpha2.isEmpty {
+            self.country = Country(iso3166Alpha2: Locale.current.regionCode ?? "", dialingCode: nil)
         }
     }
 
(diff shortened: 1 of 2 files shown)


Source/UI/NewUI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -18,8 +18,8 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
     var errorFlagOfCellType = [Int: Bool]()
     var textValueOfCellType = [Int: String]()
 
-    private var country : Country?
     private var countryRow: Int?
+    private(set) var country : Country?
```
alex-ioja-yang-cko, line 22: Why the access control change?

Source/UI/NewUI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -74,6 +74,9 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
             textValueOfCellType[type.index] = hasValue ? value : (!isMandatory ? "" : nil)
             errorFlagOfCellType[type.index] = hasValue ? false : isMandatory
         }
+        if let country = country, country.iso3166Alpha2.isEmpty {
+            self.country = Country(iso3166Alpha2: Locale.current.regionCode ?? "", dialingCode: nil)
```
alex-ioja-yang-cko, line 78: feels like a complicated way of doing it. Couldn't we just target the initialiser directly?

```
    public init(iso3166Alpha2: String, dialingCode: String?) {
        self.iso3166Alpha2 = iso3166Alpha2.isEmpty ? Locale.current.regionCode : iso3166Alpha2
        self.dialingCode = dialingCode
    }
```

We can then just test this simply with Country initialiser unit test.

alex-ioja-yang-cko: The reason I say this feels complicated is that it feels a little hacky, after setup to check if something exists, then if it exists but lacks something, to override.

Source/UI/NewUI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
ehab-al-cko: For unit testing

Source/UI/NewUI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
deepesh-vasthimal-cko: Ok makes sense for now we should ideally be making these changes in checkout iOS SDK if needed. but for now I am fine with this.

deepesh-vasthimal-cko (approved):

alex-ioja-yang-cko (approved):

PR #186, opened 2022-06-27
Add `final` and `@frozen`

ghost:
## Proposed changes

Add `final` and `@frozen`

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/CheckoutFont.swift
+++ b/Source/CheckoutFont.swift
@@ -13,7 +13,7 @@ extension UIFont {
 
 extension UIFont {
 
-    @frozen enum GraphikStyle: String, CaseIterable {
+    enum GraphikStyle: String, CaseIterable {
         case black = "Black"
         case blackItalic = "BlackItalic"
 
--- a/Source/Core/Constants+Style.swift
+++ b/Source/Core/Constants+Style.swift
@@ -1,9 +1,9 @@
 extension Constants {
 
-    @frozen enum Style {
+    enum Style {
 
-        @frozen enum PaymentForm {
-            @frozen enum InputBillingFormButton: Double {
+        enum PaymentForm {
+            enum InputBillingFormButton: Double {
                 case height = 56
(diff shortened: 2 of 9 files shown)


Source/CheckoutFont.swift
```diff
@@ -13,7 +13,7 @@ extension UIFont {
 
 extension UIFont {
 
-    enum GraphikStyle: String, CaseIterable {
+    @frozen enum GraphikStyle: String, CaseIterable {
```
alex-ioja-yang-cko, line 16: It doesn't make sense to add @frozen to something that is not public

alex-ioja-yang-cko (changes requested):
As mentioned in comment, @frozen should not be applied to something that's not public.

Source/CheckoutFont.swift
ehab-al-cko: frozen : Apply this attribute to a structure or enumeration declaration to restrict the kinds of changes you can make to the type. This attribute is allowed only when compiling in library evolution mode. When the compiler isn’t in library evolution mode, all structures and enumerations are implicitly frozen, and this attribute is ignored.

alex-ioja-yang-cko (approved):
I'd remove the final from descriptions as I think its not providing much value, and I'd like to find a way and avoid introducing unintended white space changes, as they generally just cause noise and distract from the work being done.

Otherwise, absolutely making classes final should help to improve compile times to consumer.

deepesh-vasthimal-cko (approved):

Source/Core/InputState.swift
```diff
@@ -5,7 +5,7 @@ import Foundation
 /// - normal
 /// - required
 /// - hidden
-public enum InputState {
+@frozen public  enum InputState {
```
harry-brown-cko, line 8:
```suggestion
@frozen public enum InputState {
```

PR #191, opened 2022-07-03
Feature/PIMOB-1274: Expiry date validation logic

ehab-al-cko:
[Jira ticket](https://checkout.atlassian.net/browse/PIMOB-1274?atlOrigin=eyJpIjoiMDM3ODgxNTJlYTU3NDJlNWJlMGViNzlkNTRmYjY3NjMiLCJwIjoiaiJ9)
## Proposed changes

An invalidMonth would be anything not between [01 - 12]
So if a user inputs 21 as the month, that is of course an invalid month. 
We will handle this with smart month logic, whereby 
if user enters a 0, let them enter any digit between 1-9
if user enters a 1, let them enter any digit between 0-2
if user enters a number between 2-9, then move to the year selector and append a 0 before the digit 

Smart year logic:
User must enter a number between 2-9 to start with 
Happens with a negative year or unexpected value. This would never occur with the keypad

## Types of changes

## Checklist

* [X ] Lint and unit tests pass locally with my changes

## Further comments

### DIFF
--- a/Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
+++ b/Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
@@ -73,6 +73,62 @@ public final class ExpiryDateView: UIView {
         updateErrorView(isHidden: false, text: newDate)
     }
   }
+
+  /*
+   Input validation from location 0 to 4 (MM/yy):
+   ===============
+
+   if location 0 (first month digit)
+      if user enters a number between 2-9,
+          then move to the year selector and append a 0 before the digit
+
+   if location 1 (2nd month digit)
+      if user enters a 0,
+          then enter any digit between 1-9
+      if user enters a 1,
+          then enter any digit between 0-2
+
+   if location 3 (backslash '/')
+      then user must enter a number between 2-9 to start with
+
+   if location 4 (first year digit)
+      then user must enter a number between 2-9 to start with
+
+   if location 5 (2nd year digit)
+      then validate full Expiry Date
+   */
+
+  private func validateInput(_ textField: UITextField, currentDigit: Int, location: Int, replacementText: String) -> Bool {
+    switch location {
+
+      case 0 where 2...9 ~= currentDigit:
+        textField.text = "0"
+
+      case 1:
+        guard let originalText = textField.text, let previousDigit = Int(originalText) else { return false }
+
+        switch previousDigit {
+          case  0 where 1...9 ~= currentDigit,
+                1 where 0...2 ~= currentDigit: break
+          default: return false
+        }
+
+      case 2:
+        textField.text?.append("/")
+        guard 2...9 ~= currentDigit else { return false }
+
+      case 3:
+        guard 2...9 ~= currentDigit else { return false }
+
+      case 4:
+        updateExpiryDate(to: (textField.text ?? "") + replacementText)
+        return false
+
+      default: break
+    }
+
+    return true
+  }
 }
 
 extension ExpiryDateView: TextFieldViewDelegate {
@@ -116,58 +172,6 @@ extension ExpiryDateView: TextFieldViewDelegate {
     guard CharacterSet(charactersIn: replacementText).isSubset(of: .decimalDigits) else { return false }
     guard let currentDigit = Int(replacementText) else { return false }
 
-    /*
-     Input validation from location 0 to 4 (MM/yy):
-     ===============
-
-     if location 0 (first month digit)
-        if user enters a number between 2-9,
-            then move to the year selector and append a 0 before the digit
-
-     if location 1 (2nd month digit)
-        if user enters a 0,
-            then enter any digit between 1-9
-        if user enters a 1,
-            then enter any digit between 0-2
-
-     if location 3 (backslash '/')
-        then user must enter a number between 2-9 to start with
-
-     if location 4 (first year digit)
-        then user must enter a number between 2-9 to start with
-
-     if location 5 (2nd year digit)
-        then validate full Expiry Date
-     */
-
-    switch range.location {
-
-      case 0 where 2...9 ~= currentDigit:
-        textField.text = "0"
-
-      case 1:
-        guard let originalText = textField.text, let previousDigit = Int(originalText) else { return false }
-
-        switch previousDigit {
-          case  0 where 1...9 ~= currentDigit,
-                1 where 0...2 ~= currentDigit: break
-          default: return false
-        }
-
-      case 2:
-        textField.text?.append("/")
-        guard 2...9 ~= currentDigit else { return false }
-
-      case 3:
-        guard 2...9 ~= currentDigit else { return false }
-
-      case 4:
-        updateExpiryDate(to: (textField.text ?? "") + replacementText)
-        return false
-
-      default: break
-    }
-
-    return true
+    return validateInput(textField, currentDigit: currentDigit, location: range.location, replacementText: replacementText)
   }
 }
--- a/Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
+++ b/Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
@@ -110,22 +110,64 @@ extension ExpiryDateView: TextFieldViewDelegate {
     //check for max length including added spacers which all equal to 5
     guard !string.isEmpty else { return false }
     
-    var originalText = textField.text
     let replacementText = string.replacingOccurrences(of: " ", with: "")
     
     //verify entered text is a numeric value
-    if !CharacterSet(charactersIn: replacementText).isSubset(of: .decimalDigits) { return false }
-    
-    //put `/` space after 2 digit
-    if range.location == 2 {
-      originalText?.append("/")
-      textField.text = originalText
+    guard CharacterSet(charactersIn: replacementText).isSubset(of: .decimalDigits) else { return false }
+    guard let currentDigit = Int(replacementText) else { return false }
+
+    /*
+     Input validation from location 0 to 4 (MM/yy):
+     ===============
+
+     if location 0 (first month digit)
+        if user enters a number between 2-9,
+            then move to the year selector and append a 0 before the digit
+
+     if location 1 (2nd month digit)
+        if user enters a 0,
+            then enter any digit between 1-9
+        if user enters a 1,
+            then enter any digit between 0-2
+
+     if location 3 (backslash '/')
+        then user must enter a number between 2-9 to start with
+
+     if location 4 (first year digit)
+        then user must enter a number between 2-9 to start with
+
+     if location 5 (2nd year digit)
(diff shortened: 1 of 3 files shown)


Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
```diff
@@ -73,6 +73,62 @@ public final class ExpiryDateView: UIView {
         updateErrorView(isHidden: false, text: newDate)
     }
   }
+
+  /*
+   Input validation from location 0 to 4 (MM/yy):
```
deepesh-vasthimal-cko, line 78: 👍

Tests/UI/New UI/BillingForm/View/ExpiryDateViewTests.swift
```diff
@@ -89,4 +89,139 @@ class ExpiryDateViewTests: XCTestCase {
         view.updateExpiryDate(to: "Test")
         XCTAssertFalse(view.style?.error?.isHidden ?? true)
     }
+
+  func testValidFirstDigitInputWith0() {
```
deepesh-vasthimal-cko, line 93: Can you please also add few test scenarios with garbage or invalid values and strings : for example ```
textField.text = "adbadfa", textField.text = "99/99" textField.text = "00/00" textField.text = "999999999/999999999" etc 
textField.text = "-*/@@"
```

**Review by deepesh-vasthimal-cko — APPROVED** [review]
LGTM just one comment around invalid tests scenarios.

File: Tests/UI/New UI/BillingForm/View/ExpiryDateViewTests.swift [hunk 3]
```diff
@@ -89,4 +89,139 @@ class ExpiryDateViewTests: XCTestCase {
...
+    let input = "0"
+    let textField = UITextField()
+    textField.text = ""
+
+    let shouldContinueAdding = view.textField(textField, shouldChangeCharactersIn: NSRange(location: 0, length: 0), replacementString: input)
+    if shouldContinueAdding {
+      textField.text?.append(input)
```
deepesh-vasthimal-cko, line 100: can we avoid any logic inside the unit test to only test what is in the code. For example can we just test the function `validateInput` , This and other places below where if appends.

deepesh-vasthimal-cko (changes requested):
Few minor comments.

Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
```diff
@@ -110,22 +166,12 @@ extension ExpiryDateView: TextFieldViewDelegate {
...
-    
-    if range.location == dateFormatTextCount - 1 {
-      updateExpiryDate(to: (originalText ?? "") + replacementText)
-      return false
-    }
-    return true
+    guard CharacterSet(charactersIn: replacementText).isSubset(of: .decimalDigits) else { return false }
+    guard let currentDigit = Int(replacementText) else { return false }
+
+    return validateInput(textField, currentDigit: currentDigit, location: range.location, replacementText: replacementText)
```
alex-ioja-yang-cko, line 175: Could we use an SDK configuration property _smart input validation_ to enable / disable this feature?

I think its definitely well intended, but from UX perspective I'd have to question how many times I write a 5 and I get an `05/` showing on the screen from it. Good user experience should be about behaving in the way user expects, even if sometimes that may be _less smart way_. If we'd be able to add some consumer configurable property for this, then merchant can have full control over this and can even A/B test it if they so wish.

alex-ioja-yang-cko: Having tried on simulator, I do mind it less and less the more I try, but the other UX question is:
> If I want to type month 00, or 13...19, why does my keyboard feel like its not working?

I fully get the _smart_ bit, but from user point of view, I feel like keyboard is broken. I get feedback of the press and yet nothing happens... Instead of disabling input, should we just toggle error message visibility on each input?

ehab-al-cko: This is the requirement. @patrick-hoban-cko please advice

patrick-hoban-cko: @alex-ioja-yang-cko has a very fair point, I have to admit I did find it a bit strange that we actually limit what the customer of a merchant is able to type in (even when we know they shouldn't be typing 13 or 19). 

Let's go with Alex's suggestion, and allow someone to type in 13, or 19, but then just show the error message right away when they do it. Otherwise if they hit the 3 or 9 after typing in 1, they might think something is broken.

Tests/UI/New UI/BillingForm/View/ExpiryDateViewTests.swift
ehab-al-cko: `shouldChangeCharactersIn` is returning `Bool` which allow us to add new change to text field. UITextfield should be more tested using UITest

Tests/UI/New UI/BillingForm/View/ExpiryDateViewTests.swift
ehab-al-cko: Update

Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
ehab-al-cko: @alex-ioja-yang-cko please try other branch `feature/PIMOB-1274_expiery_date_error_messages` and let me know you feedback.

ehab-al-cko: @patrick-hoban-cko may i make this update another PR or all 4 PRs will be blocked?

patrick-hoban-cko: Okay, are you saying that in https://github.com/checkout/frames-ios/pull/192 Alex's suggestion to allow a user to type in 13 or 19 is fixed?

harry-brown-cko: @patrick-hoban-cko this is about making our web and mobile SDKs' behaviour more cohesive. If we no longer want this why did we discuss it at length last week?

patrick-hoban-cko: Let's get in a call to discuss @harry-brown-cko @ehab-al-cko @alex-ioja-yang-cko

ehab-al-cko: @patrick-hoban-cko https://github.com/checkout/frames-ios/pull/192 is not supporting Alex's suggesting but it's showing error message when there's any invalid input which we have agreed on last week

patrick-hoban-cko: @alex-ioja-yang-cko we're going to merge this and discuss fine tuning the logic in the office together. Logic stays as is for now!

alex-ioja-yang-cko (approved):
I think we can improve the writing of the unit tests to improve their readability. Currently the unit tests include paths that we know will not be executed so that's pretty confusing imo.

Otherwise, unit tests are better than no unit tests 👍 (also to me it looks like you've covered the logic paths well) and this would be a good thing to reflect on.

Tests/UI/New UI/BillingForm/View/ExpiryDateViewTests.swift
```diff
@@ -89,4 +89,139 @@ class ExpiryDateViewTests: XCTestCase {
...
+    textField.text = ""
+    let input = "3"
+
+    let shouldContinueAdding = view.textField(textField, shouldChangeCharactersIn: NSRange(location: 0, length: 0), replacementString: input)
+    if shouldContinueAdding {
+      textField.text?.append(input)
+    }
+
+    XCTAssertTrue(shouldContinueAdding)
+    XCTAssertEqual(textField.text, "03")
```
alex-ioja-yang-cko, line 132:
```suggestion
    let textField = UITextField()
    textField.text = ""
    let input = "3"

    let shouldContinueAdding = view.textField(textField, shouldChangeCharactersIn: NSRange(location: 0, length: 0), replacementString: input)
    XCTAssertTrue(shouldContinueAdding)
    
    textField.text?.append(input)
    XCTAssertEqual(textField.text, "03")
```

Follow up from previous comment as this test has slightly different behaviour. In this case the reading would be
**Given** I have this expected input and prepared textField

**When** I try to add the input to the textField
**Then** I am allowed to

**When** I do add it
**Then** input is as expected

Tests/UI/New UI/BillingForm/View/ExpiryDateViewTests.swift
```diff
@@ -89,4 +89,139 @@ class ExpiryDateViewTests: XCTestCase {
...
+    let textField = UITextField()
+    textField.text = "0"
+
+    let shouldContinueAdding = view.textField(textField, shouldChangeCharactersIn: NSRange(location: 1, length: 0), replacementString: input)
+    if shouldContinueAdding {
+      textField.text?.append(input)
+    }
+
+    XCTAssertFalse(shouldContinueAdding)
+    XCTAssertEqual(textField.text, "0")
```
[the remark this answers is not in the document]
ehab-al-cko: `shouldChangeCharactersIn` is boolean to know if the new input can be added or not. after this if the return  is true then append to text field. then check in the end the what is the expected.

Tests/UI/New UI/BillingForm/View/ExpiryDateViewTests.swift
ehab-al-cko: if `shouldContinueAdding` is `false` then `textField.text` must not append the input.

alex-ioja-yang-cko: Yes, but this is a unit test, not live code. This is expectation of behaviour, not "if this, then that, otherwise something else".

Tests/UI/New UI/BillingForm/View/ExpiryDateViewTests.swift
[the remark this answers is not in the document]
alex-ioja-yang-cko: Similar to other comment, this is an unit test. The purpose of unit tests is to setup a controlled environment where given the setup, when we run an action, then the output is as expected.

If we're adding logic on the output, we missed the point of unit tests.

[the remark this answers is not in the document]
ehab-al-cko: Updated

Tests/UI/New UI/BillingForm/View/ExpiryDateViewTests.swift
ehab-al-cko: Updated

harry-brown-cko (approved):

PR #192, opened 2022-07-03
Feature/PIMOB-1274: Expiry date error messages

ehab-al-cko:
[Jira ticket](https://checkout.atlassian.net/browse/PIMOB-1274?atlOrigin=eyJpIjoiMDM3ODgxNTJlYTU3NDJlNWJlMGViNzlkNTRmYjY3NjMiLCJwIjoiaiJ9)

please review this PR firstly https://github.com/checkout/frames-ios/pull/191
## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/Resources/en.lproj/Localizable.strings
+++ b/Source/Resources/en.lproj/Localizable.strings
@@ -15,7 +15,9 @@
 "ExpiryDate" = "ExpiryDate";
 "ExpiryDateFormat" = "Format is MM/YY";
 "ExpiryDatePlaceholder" = "MM/YY";
-"ExpiryDateErrorMessage" = "Missing Expiry Date";
+"ExpiryDateErrorMessageInvalid" = "Invalid Expiry Date";
+"ExpiryDateErrorMessageMissing" = "Please enter a valid Expiry Date";
+"ExpiryDateErrorMessageInvalidInPast" = "Expiry Date is in the past";
 "CardNumber" = "Card Number";
 "CardNumberErrorMessage" = "Missing Card Number";
 "city" = "City";
--- a/Source/Resources/en.lproj/Localizable.strings
+++ b/Source/Resources/en.lproj/Localizable.strings
@@ -15,8 +15,7 @@
 "ExpiryDate" = "ExpiryDate";
 "ExpiryDateFormat" = "Format is MM/YY";
 "ExpiryDatePlaceholder" = "MM/YY";
-"ExpiryDateErrorMessageInvalid" = "Invalid Expiry Date";
-"ExpiryDateErrorMessageMissing" = "Please enter a valid Expiry Date";
+"ExpiryDateErrorMessageInvalid" = "Please enter a valid Expiry Date";
 "ExpiryDateErrorMessageInvalidInPast" = "Expiry Date is in the past";
 "CardNumber" = "Card Number";
 "CardNumberErrorMessage" = "Missing Card Number";
--- a/Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
+++ b/Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
@@ -47,21 +47,10 @@ public final class ExpiryDateView: UIView {
       style?.textfield.text = text ?? ""
       expiryDateView.update(style: style)
     }
-
     guard let error = error else { return }
-    var errorText = ""
-    switch error {
-      case .invalidMonthString,
-          .invalidYearString,
-          .invalidMonth,
-          .invalidYear:
-        errorText = Constants.LocalizationKeys.PaymentForm.ExpiryDate.Error.invalid
-      case .incompleteMonth,
-          .incompleteYear:
-        errorText = Constants.LocalizationKeys.PaymentForm.ExpiryDate.Error.missing
-      case .inThePast:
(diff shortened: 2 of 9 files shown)


Source/Resources/en.lproj/Localizable.strings
```diff
@@ -15,7 +15,9 @@
 "ExpiryDate" = "ExpiryDate";
 "ExpiryDateFormat" = "Format is MM/YY";
 "ExpiryDatePlaceholder" = "MM/YY";
-"ExpiryDateErrorMessage" = "Missing Expiry Date";
+"ExpiryDateErrorMessageInvalid" = "Invalid Expiry Date";
+"ExpiryDateErrorMessageMissing" = "Please enter a valid Expiry Date";
```
deepesh-vasthimal-cko, line 19: ExpiryDateErrorMessageMissing and ExpiryDateErrorMessageInvalid both mean same, Can we club them into one ? . Or is there a reason for it to be 2 different strings ?

ehab-al-cko: @patrick-hoban-cko please advice

patrick-hoban-cko: As per https://checkout.atlassian.net/browse/PIMOB-1274, _Please enter a valid Expiry Date_ appears when the user has not entirely filled out the month or year, for example - if they put 06/2

_Invalid Expiry Date_ should never appear in reality, as it's only a placeholder for error scenarios that should be handled by either the keypad, or smart month/year logic.

Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
```diff
@@ -41,15 +41,22 @@ public final class ExpiryDateView: UIView {
...
+    defer {
+      style?.error?.isHidden = isHidden
+      style?.textfield.text = text ?? ""
+      expiryDateView.update(style: style)
+    }
+    guard let error = error else { return }
+    let errorText = error == .inThePast ?
+    Constants.LocalizationKeys.PaymentForm.ExpiryDate.Error.past :
+    Constants.LocalizationKeys.PaymentForm.ExpiryDate.Error.invalid
+    style?.error?.text = errorText
```
alex-ioja-yang-cko, line 54: Looking at this method it feels like the text is ignored when error exists. This would make me think 
> why do we need so many parameters if we purposely ignore some?

I wonder if it would be simpler and clearer if this method was still only receiving a text. I would even remove `isHidden`, as I think its a fair assumption that:
- you wouldn't show an error without text
- you wouldn't send text for error and not expect error to be shown

Code level, this would now be:
```swift
private func updateErrorView(text: String?) {
    let hideError = text?.isEmpty != false
    style?.error?.isHidden = hideError
    style?.textfield.text = text // Should be able to be optional ?
    expiryDateView.update(style: style)
}
```

Then to consumers where error was passed, they would now need to own converting an error to a message
```swift
let errorMessage = Constants.LocalizationKeys.PaymentForm.ExpiryDate.Error.past
updateErrorView(text: errorMessage)
```

More impact from this could be removing error cases (invalidYear, invalidMonth seem to have the same result for example)

alex-ioja-yang-cko (commented):
I wonder if we could simplify the current logic by reducing object abstractions that would not be needed and function argument?

Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
alex-ioja-yang-cko: I now see that style has an errorText and a textField text, all update whilst updating error? When we _updateErrorView_, why the side effect of updating a textfield?

ehab-al-cko: `text` is in defer, it's not ignored

alex-ioja-yang-cko: I was getting confused on the `error.text` and `textField.text`. That to me is confusing as well. I still think we can improve on the code simplicity, but your unit tests do seem to show the behaviour works, so as far as strictly getting the work done, this seems to be ✅

alex-ioja-yang-cko (approved):
I genuinely appreciate you took time in PRs to try and simplify code where possible. As my other comment, I do wonder if `updateErrorView(isHidden` could still be simplified here and reduce its side effects, but overall your extensive test coverage has allowed us to mitigate for potential developer error in the future 👍

alex-ioja-yang-cko (approved):

Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
```diff
@@ -66,43 +52,45 @@ public final class ExpiryDateView: UIView {
...
-    
+
     switch cardValidator.validate(expiryMonth: monthDigit, expiryYear: yearDigit) {
       case .success:
         let expiryDate = ExpiryDate(month: monthDigit, year: yearDigit)
         delegate?.update(expiryDate: expiryDate)
-        updateErrorView(isHidden: true, text: newDate)
+        updateErrorViewStyle(isHidden: true, textfieldText: newDate)
       case .failure(let error):
         print(error)
```
alex-ioja-yang-cko, line 65: This print should be removed, even if not now, but in a different PR.

PR #200, opened 2022-07-07
PIMOB-1204: card number validation and icon changing logic

harry-brown-cko:
## Proposed changes

- Add card validation
- Add scheme change logic
- Add new icons and resizing

## Types of changes

## Checklist

## Further comments

Customisation is not covered as part of this ticket


Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -0,0 +1,151 @@
...
+import Checkout
+
+protocol CardNumberViewDelegate: AnyObject {
+  func update(cardNumber: String)
+}
+
+final class CardNumberView: UIView {
+  weak var delegate: CardNumberViewDelegate?
+
+  private let cardUtils = CardUtils()
```
deepesh-vasthimal-cko, line 19: @harry-brown-cko to put something in backlog to move cardUtils to checkout iOS SDK as CardFormatter class to have a format method in it as a tech investment story. Reason the CardUtils does not need to be public anymore in Frames and can live inside Checkout iOS SDK. This also needs a update to API spec.

Source/UI/NewUI/CommonUI/View/Component/InputView.swift
```diff
@@ -156,6 +166,10 @@ extension InputView {
 
     private func setupIcon() {
         textFieldContainer.addArrangedSubview(iconView)
+
+        NSLayoutConstraint.activate([
```
deepesh-vasthimal-cko, line 170: Question : What is the reason for this ?

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -0,0 +1,59 @@
...
+//  Copyright © 2022 Checkout. All rights reserved.
+//
+
+import UIKit
+
+protocol CardNumberViewProtocol: AnyObject {
+  var schemeIcon: Constants.Bundle.SchemeIcon { get set }
+}
+
+final class CardNumberView: UIView, CardNumberViewProtocol {
```
deepesh-vasthimal-cko, line 15: I am assuming that `CardNumberView` will be used by merchant developers this should be public ? if so then can you also update the API Spec.

harry-brown-cko: Have left as internal for now

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -0,0 +1,59 @@
...
+  private let viewModel: CardNumberViewModelProtocol
+
+  var schemeIcon = Constants.Bundle.SchemeIcon.blank {
+    didSet {
+      if schemeIcon != oldValue {
```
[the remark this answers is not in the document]
harry-brown-cko: `oldValue` is automatically set to the previous value in a `didSet` block. Basically this is so we only update the icon when it changes

deepesh-vasthimal-cko (approved):
Just few comments on CardNumberView being public.

Source/UI/CardListCell.swift
```diff
@@ -82,7 +82,7 @@ public class CardListCell: UITableViewCell {
         schemeImageView.image = schemeIconLocation(scheme: scheme)?.rawValue.image(forClass: CardListCell.self)
     }
 
-    private func schemeIconLocation(scheme: Card.Scheme) -> Constants.Bundle.SchemeIcons? {
+    private func schemeIconLocation(scheme: Card.Scheme) -> Constants.Bundle.SchemeIcon? {
```
ehab-al-cko, line 85: to this can be extension to Constants.Bundle.SchemeIcon. It may be reused for header view and card view.

Tests/UI/CvvInputViewTests.swift
```diff
@@ -4,7 +4,7 @@ import XCTest
 class CvvInputViewTests: XCTestCase {
 
     var cvvInputView: CvvInputView!
-    var stubCardValidator: StubCardValidator! = StubCardValidator()
+    var stubCardValidator: MockCardValidator! = MockCardValidator()
```
ehab-al-cko, line 7:
```suggestion
    var stubCardValidator = MockCardValidator()
```

Source/UI/NewUI/CommonUI/View/Component/InputView.swift
```diff
@@ -7,6 +7,7 @@ class InputView: UIView {
     weak var delegate: TextFieldViewDelegate?
     private(set) var style: CellTextFieldStyle?
     private(set) lazy var textFieldContainerBottomAnchor = textFieldContainer.bottomAnchor.constraint(equalTo: bottomAnchor)
+    private(set) lazy var iconViewWidthAnchor = iconView.widthAnchor.constraint(equalToConstant: 32)
```
alex-ioja-yang-cko, line 10: Might be handy to set that 32 inside the Constants. Would even question if we're sure we want it to always be 32, as there seems to be no other logic to ever change it.

Source/UI/NewUI/CommonUI/View/Component/InputView.swift
```diff
@@ -86,9 +89,16 @@ class InputView: UIView {
         textFieldContainer.layer.cornerRadius = style.textfield.cornerRadius
         textFieldContainer.layer.borderWidth = style.textfield.borderWidth
         textFieldContainer.backgroundColor = style.textfield.backgroundColor
+    }
 
+    private func updateTextFieldContainer(image: UIImage?, withAnimation: Bool) {
```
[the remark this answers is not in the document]
alex-ioja-yang-cko: I'll make this more fun and add that UIViewController().present method uses `animated` 🙃. 

I would throw in more points that:
- withAnimation sounds like we will provide a completion handler for the animation action (similar to `UIView.animate(withDuration`
- `isAnimated` would sound like a stored property that describes the existing state rather than an instruction for how to perform something
- `shouldAnimate` would be syntactically more on point but a less concise option



Either way, whatever you decide (I'm not convinced there's an obvious right / wrong here), please make sure to keep consistency with the publicly exposed declaration at line 67
```func update(style: CellTextFieldStyle? = nil, textFieldValue: String? = nil, image: UIImage? = nil, withAnimation: Bool = false){```

File: Source/UI/NewUI/CommonUI/View/Core/ImageContainerView.swift [hunk 9]
```diff
@@ -1,8 +1,9 @@
 import UIKit
 
 class ImageContainerView: UIView {
+    private static let animationLength: TimeInterval = 0.25
```
alex-ioja-yang-cko, line 4: This I would love to be in Constants and make sure all our animation durations are consistent

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -0,0 +1,59 @@
...
+//  Frames
+//
+//  Created by Harry Brown on 05/07/2022.
+//  Copyright © 2022 Checkout. All rights reserved.
+//
+
+import UIKit
+
+protocol CardNumberViewProtocol: AnyObject {
+  var schemeIcon: Constants.Bundle.SchemeIcon { get set }
```
alex-ioja-yang-cko, line 12: When do we envision to reuse CardNumberViewProtocol such that we benefit now from creating the interface?

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -0,0 +1,59 @@
...
+}
+
+final class CardNumberView: UIView, CardNumberViewProtocol {
+  private let viewModel: CardNumberViewModelProtocol
+
+  var schemeIcon = Constants.Bundle.SchemeIcon.blank {
```
alex-ioja-yang-cko, line 18: Could we move this property inside the viewModel and have it changed that way? This would allow us to test that:
> Given scheme Icon has been set, a callback gets received with an appropriate image

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -0,0 +1,59 @@
...
+  private func setupView() {
+    addSubview(cardNumberInputView)
+    cardNumberInputView.setupConstraintEqualTo(view: self)
+  }
+
+  private func updateIcon() {
+    cardNumberInputView.update(image: image(for: schemeIcon), withAnimation: true)
+  }
+
+  private func image(for schemeIcon: Constants.Bundle.SchemeIcon) -> UIImage? {
```
alex-ioja-yang-cko, line 56: Looks like we may add this method to SchemeIcon itself?

```swift
enum SchemeIcon {

    // .....

    var image: UIImage {
        rawValue.image(forClass: Self.self)
    }

}
```

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
```diff
@@ -0,0 +1,115 @@
...
+  weak var cardNumberView: CardNumberViewProtocol?
+
+  private let cardUtils = CardUtils()
+  private let cardValidator: CardValidating
+
+  init(cardValidator: CardValidating) {
+    self.cardValidator = cardValidator
+  }
+
+  func buildView() -> CardNumberView {
```
alex-ioja-yang-cko, line 29: Not sure I have seen this flow previously:
- create viewModel
- ask view model to create view
- viewmodel registers itself on view, whilst also returning the newly created view.

Generally the expectation from most flows is:
```swift
let view = MyView()
let viewModel = MyViewModel()
viewModel.delegate = view // Should be a delegate for MyViewModelDelegate, enabling viewModel to ask UI to sync when needed
view.viewModel = viewModel
```

I am not too bothered about it for now, but I do worry a viewmodel normally should not have any UI knowledge. And in this case, the viewmodel file is importing UIKit, which is a red flag and can further weaken that separation over time.

alex-ioja-yang-cko (commented):
I am not sure the MVVM version used is ideal. Whilst not too bothered about it, importing UIKit inside a viewmodel object declaration is less than ideal as it allows to add poorer implementations without any flags being triggered

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
harry-brown-cko: We use this for tests

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
harry-brown-cko: Constants is currently a configuration struct with no business logic. As the scheme icon is only used here it makes sense to keep this method here

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
harry-brown-cko: The inclusion of UIKit is only to implement the TextFieldViewDelegate methods. I'll sort out the setting of delegates

alex-ioja-yang-cko: Specifically for handling delegate methods, I wonder if we could pull that into a separate file (CardNumberViewModel+Extension.swift), where we only create an extension that conforms to the protocol. And then we can maybe have a cleaner separation 🤔

Tests/UI/CvvInputViewTests.swift
harry-brown-cko: This logic and that in `tearDown()` are used to automatically reinit between tests. This is a pattern across Frames and 3DS

Source/Core/Constants+Style.swift
```diff
@@ -5,10 +5,14 @@
 //  Copyright © 2022 Checkout. All rights reserved.
 //
 
+import Foundation
+
 extension Constants {
 
     enum Style {
 
+      static let animationLength: TimeInterval = 0.25
```
alex-ioja-yang-cko, line 14: **Interesting fact 🤓**

 
TimeInterval is actually a **Double**
```
public typealias TimeInterval = Double
```
***
But whilst Double can be inferred by compiler without any import (probably CoreFoundation), the above mentioned typealias is actually declared within Foundation -> NSDate (declared as a global alongside the NSDate object declaration, hence type is TimeInterval and not NSDate.TimeInterval).
***
And specifically here, if we declare
```swift
static let animationLength = 0.25
```
an import to Foundation would not be added, as type gets automatically inferred to Double. But at use location it would automatically be handled as the typealias if that's what's needed.

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
alex-ioja-yang-cko: I think you have already created that property in the enum, so we may then be able to remove this method and reference that schemeIcon.image you added?

Source/UI/NewUI/CommonUI/View/Component/InputView.swift
harry-brown-cko: When I add in customisation this will be moved to constants

Source/Core/Constants+Style.swift
harry-brown-cko: I'd prefer to use TimeInterval here so that if this alias was ever changed then we would automatically move with it

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
harry-brown-cko: This has been addressed now. CardNumberView handles anything UIKit specific and the business logic is still in the viewModel

ehab-al-cko (approved):

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
```diff
@@ -0,0 +1,115 @@
...
+    case .tooLong:
+      return nil
+    case .invalidScheme:
+      cardNumberView?.schemeIcon = schemeIconLocation(scheme: .unknown)
+
+      return .unknown
+    case .cardNumber(let cardNumberError):
+      switch cardNumberError {
+      case .invalidCharacters:
+        return nil
```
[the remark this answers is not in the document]
harry-brown-cko: preferring to manually handle each case so that if any extra errors are added then the compiler will throw an error and we will have to address it

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
```diff
@@ -0,0 +1,74 @@
...
+  func update(cardNumber: String, scheme: Card.Scheme)
+}
+
+protocol CardNumberViewModelProtocol {
+  func textFieldUpdate(from: String) -> String?
+}
+
+class CardNumberViewModel {
+  weak var delegate: CardNumberViewModelDelegate?
+  weak var cardNumberView: CardNumberViewProtocol?
```
alex-ioja-yang-cko, line 21: I think this UI view interface could be merged with `CardNumberViewModelDelegate` and removed.

```swift
protocol CardNumberViewModelDelegate: AnyObject {
   func viewModelUpdate(cardNumber: String, scheme: Card.Scheme)
   func viewModelUpdate(schemeIcon: Constants.Bundle.SchemeIcon)
 }

// Then 
cardNumberView?.schemeIcon = .init(scheme: scheme)
// Becomes
delegate.viewModelUpdated(schemeIcon: .init(scheme: scheme)
```

Currently your view owns viewmodel and viewmodel owns the view twice, once as a delegate and again as a view interface. Implementation is safe now (not memory retain cycle), but that duplication of ownership could be merged into 1

Tests/Core/Extensions/ConstantsSchemeIconTests.swift
```diff
@@ -0,0 +1,49 @@
...
+    let testCases: [Constants.Bundle.SchemeIcon] = [
+      .blank,
+      .americanExpress,
+      .discover,
+      .jcb,
+      .maestro,
+      .mastercard,
+      .mada,
+      .visa
+    ]
```
alex-ioja-yang-cko, line 43: If you make the source enum conform to `CaseIterable`, you will save all this syntax.

```swift
enum Constants.Bundle.SchemeIcon: CaseIterable
....

let testCases = Constants.Bundle.SchemeIcon.allCases
```

Tests/Mocks/MockCardNumberViewModel.swift
```diff
@@ -0,0 +1,19 @@
...
+//  FramesTests
+//
+//  Created by Harry Brown on 13/07/2022.
+//  Copyright © 2022 Checkout. All rights reserved.
+//
+
+@testable import Frames
+
+final class MockCardNumberViewModel: CardNumberViewModelProtocol {
+  private(set) var textFieldUpdateCalledWith: String?
```
alex-ioja-yang-cko, line 12: Not important right now, but if we use arrays for these `calledWith` records, we can then make sure we also count the number of calls received, as well as check progress instead of just last value received.

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
```diff
@@ -0,0 +1,74 @@
...
+
+    if let scheme = shouldAllowChange(cardNumber: cardNumber) {
+      // using CardUtils until we add this functionality into Checkout SDK
+      return cardUtils.format(cardNumber: cardNumber, scheme: scheme)
+    }
+
+    return nil
+  }
+
+  private func shouldAllowChange(cardNumber: String) -> Card.Scheme? {
```
alex-ioja-yang-cko, line 43:
```suggestion
  private func checkScheme(for cardNumber: String) -> Card.Scheme? {
```

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
alex-ioja-yang-cko: As I look more, I also wonder whether you want to merge the functions as well, I think they are always called together anyway

alex-ioja-yang-cko (commented):
Looks good, just wonder if we can go without interfacing the UIView 🤔

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
harry-brown-cko: will address in a follow up pr as part of error checking. please approve

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
harry-brown-cko: will address in follow up pr

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
alex-ioja-yang-cko: Sure. If you have the approvals you can always merge without my explicit approval, I'm happy to follow up later or to have your opinion regardless 👍 . Will not take it personal 😛

PR #202, opened 2022-07-12
Apply SwiftLint

alex-ioja-yang-cko:
Updated SwiftLint file to cover Example Frame folder.

Run `swiftlint autocorrect` at project level and commit all changes.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,7 +1,9 @@
 included: # paths to include during linting. `--path` is ignored if present.
   - Source
-  - Tests
   - iOS Example
+  - iOS Example Frame
+  - iOS Example Frame Carthage
+  - iOS Example Frame SPM
 excluded: # paths to ignore during linting. Takes precedence over `included`.
   - Carthage
   - Pods
--- a/Source/Core/Constants+LocalizationKeys.swift
+++ b/Source/Core/Constants+LocalizationKeys.swift
@@ -28,7 +28,7 @@ extension Constants {
             }
         }
         struct BillingForm {
-            
+
             struct Cell {
                 static let optionalInput = "BillingFormOptional".localized
             }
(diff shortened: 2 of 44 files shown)


.swiftlint.yml
```diff
@@ -1,7 +1,9 @@
 included: # paths to include during linting. `--path` is ignored if present.
```
ehab-al-cko, line 1: What are swiftlint rules?

alex-ioja-yang-cko: Good question 😂

Running it hasn't changed anything other than spacings, so for now I am tempted to class it as `working`, but you can start a separate discussion for changing the rules if you want 👍.

From [SwiftLint rules list](https://realm.github.io/SwiftLint/rule-directory.html) this configuration uses all the **Default Rules** and none of the **Opt-In Rules**. It then customises the default `identifier_name` rule to allow `id` as a valid property name (would normally be excluded on the grounds of being less than 3 characters long - default behaviour of that rule).

deepesh-vasthimal-cko (approved):
LGTM

deepesh-vasthimal-cko:
@ehab-al-cko @alex-ioja-yang-cko We can also now enable / update CI / CD to check for linting.

.swiftlint.yml
```diff
@@ -1,7 +1,9 @@
...
   - Source
```
alex-ioja-yang-cko, line 2: @deepesh-vasthimal-cko 

> @ehab-al-cko @alex-ioja-yang-cko We can also now enable / update CI / CD to check for linting.

I wouldn't rush to enforce linting on pipelines yet, its a bit of a waste of resources imo. We could instead review how we develop and see if its appropriate to get SwiftLint autocorrect inside the build scripts?

@harry-brown-cko have you done any discovery on doing the linting locally?

ehab-al-cko (approved):

PR #203, opened 2022-07-13
Add CardValidator to payment view model

ehab-al-cko:
[Jira ticket](https://checkout.atlassian.net/jira/software/c/projects/PIMOB/boards/988?modal=detail&selectedIssue=PIMOB-1191)
## Proposed changes

ExpiryDateView(environment: viewModel.environment) is public and needs to pass in the environment for merchants to add this to their UI. 

Instead the ExpiryDateView should take in a CardValidator in the init method. 

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
+++ b/Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
@@ -4,28 +4,30 @@ import Checkout
 public typealias Card = Checkout.Card
 
 public struct PaymentFormFactory {
-    public static func getPaymentFormViewController(billingFormData: BillingForm?,
-                                                    paymentFormStyle: PaymentFormStyle?,
-                                                    billingFormStyle: BillingFormStyle?,
-                                                    supportedSchemes: [Card.Scheme] ) -> UIViewController {
+  public static func getPaymentFormViewController(environment: Environment,
+                                                  billingFormData: BillingForm?,
+                                                  paymentFormStyle: PaymentFormStyle?,
+                                                  billingFormStyle: BillingFormStyle?,
+                                                  supportedSchemes: [Card.Scheme]) -> UIViewController {
+    let cardValidator = CardValidator(environment: environment.checkoutEnvironment)
+    let viewModel = DefaultPaymentViewModel(cardValidator: cardValidator,
+                                            billingFormData: billingFormData,
+                                            paymentFormStyle: paymentFormStyle,
+                                            billingFormStyle: billingFormStyle,
+                                            supportedSchemes: supportedSchemes)
 
-      let viewModel = DefaultPaymentViewModel(environment: Environment.live ,
-                                              billingFormData: billingFormData,
-                                                paymentFormStyle: paymentFormStyle,
-                                              billingFormStyle: billingFormStyle,
-                                              supportedSchemes: supportedSchemes)
-        let viewController =  PaymentViewController(viewModel: viewModel)
+    let viewController = PaymentViewController(viewModel: viewModel)
 
-        if #available(iOS 13.0, *) {
-            viewController.isModalInPresentation  = true
-        }
-        return viewController
+    if #available(iOS 13.0, *) {
+      viewController.isModalInPresentation = true
     }
+    return viewController
+  }
 
-    static func getButton(style: CellButtonStyle, delegate: SelectionButtonViewDelegate?) -> UIView {
-        let view = SelectionButtonView()
-        view.delegate = delegate
-        view.update(style: style)
-        return view
-    }
+  static func getButton(style: CellButtonStyle, delegate: SelectionButtonViewDelegate?) -> UIView {
+    let view = SelectionButtonView()
+    view.delegate = delegate
+    view.update(style: style)
+    return view
+  }
 }
--- a/Tests/UI/New UI/PaymentFormFactoryTests.swift
+++ b/Tests/UI/New UI/PaymentFormFactoryTests.swift
@@ -1,0 +1,49 @@
+//
+//  PaymentFormFactoryTests.swift
+//  
+//
+//  Copyright © 2022 Checkout. All rights reserved.
+//
+
+import XCTest
+import Checkout
+@testable import Frames
+
+class PaymentFormFactoryTests: XCTestCase {
+
+  func testGetPaymentFormViewController() throws {
+
+    let billingFormStyle = BillingFormFactory.defaultBillingFormStyle
+    let paymentFormStyle = BillingFormFactory.defaultPaymentFormStyle
+    let address = Address(addressLine1: "Test line1",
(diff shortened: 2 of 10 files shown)


deepesh-vasthimal-cko (approved):

alex-ioja-yang-cko (changes requested):
> ExpiryDateView(environment: viewModel.environment) is public and needs to pass in the environment for merchants to add this to their UI.
> Instead the ExpiryDateView should take in a CardValidator in the init method.


As the description states, looks like the card validator is replacing the environment. That property is now unused.

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
```diff
@@ -1,27 +1,29 @@
 import UIKit
+import Checkout
 
 public struct PaymentFormFactory {
+  public static func getPaymentFormViewController(environment: Environment = .live,
```
alex-ioja-yang-cko, line 5: Could we unit test this method to check the factory creates a ViewController with all the expected properties?

Tests/UI/New UI/PaymentFormFactoryTests.swift
```diff
@@ -0,0 +1,21 @@
...
+@testable import Frames
+
+class PaymentFormFactoryTests: XCTestCase {
+
+  func testGetPaymentFormViewController() {
+    let viewController = PaymentFormFactory.getPaymentFormViewController(
+      billingFormData: nil,
+      paymentFormStyle: nil,
+      billingFormStyle: nil)
+    XCTAssertNotNil(viewController)
```
alex-ioja-yang-cko, line 18: Could we also dig into the properties of the controller to check that the visible properties are as expected?

Maybe also add tests for those properties not being nil and seeing the factory generates the VC with all the properties you would expect for the inputs?

alex-ioja-yang-cko: Had a look and the ViewModel inside VC is internal for reading, so you could check things like:
- viewController.viewModel. cardValidator
- viewController.viewModel. billingFormData
- viewController.viewModel. paymentForStyle
- viewController.viewModel. billingFormStyle

iOS Example Frame/iOS Example Frame/MainViewController.swift
```diff
@@ -66,10 +69,10 @@ final class MainViewController: UIViewController, CardViewControllerDelegate, Th
                           country: Self.countryGB)
         let name = "User Custom 1"
         let billingForm = BillingForm(name: name, address: address, phone: phone)
-        let paymentFormViewController = PaymentFormFactory.getPaymentFormViewController(
-            billingFormData: billingForm,
-            paymentFormStyle: Style.Custom1.paymentForm,
-            billingFormStyle: Style.Custom1.billingForm)
+        let paymentFormViewController = PaymentFormFactory.getPaymentFormViewController( environment: .live,
```
deepesh-vasthimal-cko, line 72: Change it to sandbox.

iOS Example Frame/iOS Example Frame/MainViewController.swift
```diff
@@ -85,10 +88,10 @@ final class MainViewController: UIViewController, CardViewControllerDelegate, Th
         let name = "User Custom 2"
         let billingForm = BillingForm(name: name, address: address, phone: nil)
 
-        let paymentFormViewController = PaymentFormFactory.getPaymentFormViewController(
-            billingFormData: billingForm,
-            paymentFormStyle: Style.Custom2.paymentForm,
-            billingFormStyle: Style.Custom2.billingForm)
+        let paymentFormViewController = PaymentFormFactory.getPaymentFormViewController( environment: .live,
```
deepesh-vasthimal-cko, line 91: Change it to sandbox.

Tests/UI/New UI/PaymentFormFactoryTests.swift
```diff
@@ -0,0 +1,47 @@
...
+    let name = "User 1"
+
+    let billingForm = BillingForm(name: name, address: address, phone: phone)
+
+    let viewController = PaymentFormFactory.getPaymentFormViewController(
+      billingFormData: billingForm,
+      paymentFormStyle: paymentFormStyle,
+      billingFormStyle: billingFormStyle)
+
+    XCTAssertNotNil(viewController)
```
alex-ioja-yang-cko, line 36: I see that `PaymentFormFactory.getPaymentFormViewController` returns `UIViewController`, not `UIViewController?`.

I think its worth testing things that are possible, not Swift type safety approach. Swift language team probably covers that on their side.

Tests/UI/New UI/PaymentFormFactoryTests.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    let paymentViewController = try XCTUnwrap(viewController as? PaymentViewController)
+
+    XCTAssertNotNil(paymentViewController.viewModel)
+    XCTAssertNotNil(paymentViewController.viewModel.billingFormData)
+    XCTAssertNotNil(paymentViewController.viewModel.billingFormStyle)
+    XCTAssertNotNil(paymentViewController.viewModel.paymentFormStyle)
```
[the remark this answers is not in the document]
ehab-al-cko: It will take a lot of time to make all objects are equatable. There are other tests to make sure all inside are tested.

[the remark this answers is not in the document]
alex-ioja-yang-cko: > It will take a lot of time to make all objects are equatable.
Sometimes instead of making object equatable we can just compare the properties in separate assertions. It depends how many properties there are


> There are other tests to make sure all inside are tested.
There's no unit test to confirm the data provided to the factory is the data ending up inside the returned VC or its VM. There's also no unit test to confirm that any custom changes introduced in the factory are unit tested to prevent unintended changes. There may be tests for VC and/or the VM, but we're looking at the factory, its behaviour should also be validated and constrained by appropriate unit tests imo.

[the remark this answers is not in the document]
ehab-al-cko: This will take alot of time to check every value in all features in frames sdk. it will cost alot of time.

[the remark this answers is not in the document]
ehab-al-cko: @deepesh-vasthimal-cko what do you think?

alex-ioja-yang-cko (approved):

Tests/UI/New UI/PaymentFormFactoryTests.swift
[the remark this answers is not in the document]
alex-ioja-yang-cko: Tried this myself, yes its not possible. Pretty disappointed as unit testing should be the goal of most design patterns and in this case this is a huge black hole we cannot enforce.

PR #208, opened 2022-07-15
PIMOB-1204: card number error handling

harry-brown-cko:
## Proposed changes

- Error messaging for card number
- Reworking of `CardNumberView` and `CardNumberViewModel`
- Update and add new tests

## Types of changes

## Checklist

## Further comments

N/A


Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -8,21 +8,31 @@
 
 import UIKit
 
-protocol CardNumberViewProtocol: AnyObject {
-  var schemeIcon: Constants.Bundle.SchemeIcon { get set }
+enum CardNumberError: Error {
+  case invalid
```
alex-ioja-yang-cko, line 12: I wouldn't create this enum for now to be honest, unless if you know there would be multiple types of error coming in the future. But at this stage, the number will either be valid, or invalid?

So I would remove the property `cardNumberError` from the VC and instead, from within `textFieldShouldEndEditing` you can just straight away call a function
```swift
func showError(_ show: Bool) {
     style.error?.isHidden = !show
}
```
(instead of the current `updateError` from line 73)

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -52,13 +69,31 @@ final class CardNumberView: UIView, CardNumberViewProtocol {
...
   func textFieldShouldReturn() -> Bool { true }
-  func textFieldShouldEndEditing(textField: UITextField, replacementString: String) -> Bool { true }
+
+  func textFieldShouldEndEditing(textField: UITextField, replacementString: String) -> Bool {
+    switch viewModel.validate(cardNumber: replacementString) {
+    case .success(let schemeIcon):
+      self.schemeIcon = schemeIcon
+      self.cardNumberError = nil
+    case .failure(let cardNumberError):
+      self.cardNumberError = cardNumberError
```
alex-ioja-yang-cko, line 92: Together with the other suggestion, I wouldn't hold the error, just call the method I suggested, on both paths.

❗ if I start inputting a VISA number and start deleting, you will not remove the scheme icon until I enter a different scheme. So I can have an input of 42, with a Mastercard logo, which may appear confusing.

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -70,8 +105,11 @@ extension CardNumberView: TextFieldViewDelegate {
 
     let text = oldValue.replacingCharacters(in: textRange, with: string)
 
-    if let newTextFieldValue = viewModel.textFieldUpdate(from: text) {
+    if let (newTextFieldValue, schemeIcon) = viewModel.eagerValidate(cardNumber: text) {
```
alex-ioja-yang-cko, line 108: Should we add some check for text count before we send to eager validate?

Thinking to maybe have
```swift
let eagerValidateLenghtRange = 3...6
if eagerValidateLenghtRange.contains(text.count),
  (newTextFieldValue, schemeIcon) = viewModel.eagerValidate(cardNumber: text) {
```

Not sure if it makes sense, let me know what you think. My main reasoning was that once _validated eagerly_, it might not make sense to keep validating. So if I enter 4 characters and you have the eager validation, whilst I keep adding characters we could save the UI updates 🤔

alex-ioja-yang-cko (commented):
Just wonder if we could use slightly less code & less complexity to achieve this 🤔 . Otherwise 👍

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
harry-brown-cko: I agree that we should be minimising UI updates, but there are a few edge cases here that block us from only checking in the first few digits or only checking until we have a scheme:

## Scenario A

1. Enter full Visa card number
1. Select all and paste in Amex card number

This would result in the wrong scheme icon shown, wrong formatting, and the wrong security code validation length

## Scenario B

1. Enter Mada card number

As Mada card numbers are a subset of Visa and Mastercard card numbers, the card will first register as a Visa/Mastercard and then when we have enough characters it will change to Mada. eg. 5297412542005689

alex-ioja-yang-cko: Makes sense, thanks for explaining it 👍

Source/Resources/en.lproj/Localizable.strings
```diff
@@ -23,7 +23,7 @@
 "SecurityCodeErrorMessage" = "Please enter a valid security code";
 
 "CardNumber" = "Card Number";
```
ehab-al-cko, line 25: Please change it to "Card number" like figma design

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -41,6 +51,13 @@ final class CardNumberView: UIView, CardNumberViewProtocol {
   }
 
   func update(style: CellTextFieldStyle) {
+    var style = style
+
+    if let errorStyle = style.error, errorStyle.text.isEmpty {
+      style.error?.text = Constants.LocalizationKeys.PaymentForm.CardNumber.error
```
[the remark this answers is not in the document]
deepesh-vasthimal-cko: @ehab-al-cko Merchant should be able to set this right ? for example we do something similar as below 
`public var error: ElementErrorViewStyle? = ErrorInputLabelStyleCustom1(text: "please fill expiry date")`

deepesh-vasthimal-cko (approved):
Just one comment from @alex-ioja-yang-cko on textFieldShouldEndEditing, otherwise looks good to me.

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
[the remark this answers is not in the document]
ehab-al-cko: @deepesh-vasthimal-cko also, Should integration with custom style 1 and 2 be in this PR or another one?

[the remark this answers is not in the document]
deepesh-vasthimal-cko: I assume so, as thats they way to test with sample apps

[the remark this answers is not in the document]
ehab-al-cko: @harry-brown-cko As you prefer, style 1 and 2 can be added for card number in PR or other PR.

ehab-al-cko (approved):

alex-ioja-yang-cko (approved):

PR #210, opened 2022-07-15
PIMOB-1191: Remove unused `isPlaceHolderHidden`

ghost:
[jira ticket ](https://checkout.atlassian.net/jira/software/c/projects/PIMOB/boards/988?modal=detail&selectedIssue=PIMOB-1191)
## Proposed changes

`isPlaceHolderHidden` when set to true does not work still can see the MM/YY in text field as place holder.

## Checklist

## Further comments

### DIFF
--- a/Tests/UI/New UI/BillingForm/View/TextFieldViewTests.swift
+++ b/Tests/UI/New UI/BillingForm/View/TextFieldViewTests.swift
@@ -26,12 +26,9 @@ class TextFieldViewTests: XCTestCase {
     XCTAssertEqual(view.textField.font, style.font)
     XCTAssertEqual(view.textField.textColor, style.textColor)
     XCTAssertEqual(view.textField.tintColor, style.tintColor)
-
-    let placeHolder = style.placeHolder.isEmpty ? nil : style.placeHolder
-    XCTAssertEqual(view.textField.placeholder, placeHolder)
-
-    let keypad: UIKeyboardType = style.isSupportingNumericKeyboard ? .numberPad : .default
-    XCTAssertEqual(view.textField.keyboardType, keypad)
+    XCTAssertEqual(view.textField.placeholder, style.placeHolder)
+    XCTAssertFalse(style.isSupportingNumericKeyboard)
+    XCTAssertEqual(view.textField.keyboardType, .default)
   }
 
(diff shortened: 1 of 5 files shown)


Tests/UI/New UI/BillingForm/View/TextFieldViewTests.swift
```diff
@@ -21,6 +21,19 @@ class TextFieldViewTests: XCTestCase {
...
+    XCTAssertEqual(view.textField.text, style.text)
+    XCTAssertEqual(view.textField.font, style.font)
+    XCTAssertEqual(view.textField.textColor, style.textColor)
+    XCTAssertEqual(view.textField.tintColor, style.tintColor)
+
+    let placeHolder = style.placeHolder.isEmpty ? nil : style.placeHolder
+    XCTAssertEqual(view.textField.placeholder, placeHolder)
+
+    let keypad: UIKeyboardType = style.isSupportingNumericKeyboard ? .numberPad : .default
+    XCTAssertEqual(view.textField.keyboardType, keypad)
```
alex-ioja-yang-cko, line 34: There should be no `? : ` of `if/else` inside unit tests.

Unit tests are testing that **Given this setup I have prepared, this is the outcome**. There should be no need for *Given this setup I have prepared, if this is that, otherwise this is the outcomes*.

Usually those paths would result in separate unit test paths. This is a single unit test that runs once, so we should know what behaviour is enforcing.

alex-ioja-yang-cko: Another problem with these tests where the view is declared at object level is tests are very flaky and are heavily dependant on your running logic. Xcode tests run alphabetically, so even renaming a unit test can start showing failures, which are not related to the actual logic being tested.

Whilst I agree this is how a lot of tests in frames have been setup, I would suggest trying to move away from it and use clear tests that have no dependency on environment outside of what can't be avoided.

@harry-brown-cko , @deepesh-vasthimal-cko this may be something to chat about at iOS meetings? Could extend to get Android guys feedback too?

ehab-al-cko: Updated

PR #220, opened 2022-07-25
PIMBO-1337: Update some localisations

ehab-al-cko:
## Proposed changes

Standard validation error text for County is incorrect
Should be:Missing county

Payment header and subtitle 

## Types of changes

## Checklist

## Further comments
Fix:

### DIFF
--- a/Source/Resources/en.lproj/Localizable.strings
+++ b/Source/Resources/en.lproj/Localizable.strings
@@ -20,7 +20,8 @@
 "SecurityCode" = "Security code";
 "SecurityCodeHint" = "3 - 4 digit code on your card";
 "SecurityCodeErrorMessage" = "Please enter a valid security code";
-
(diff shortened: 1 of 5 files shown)


Source/Resources/en.lproj/Localizable.strings
```diff
@@ -38,7 +38,7 @@
 "countryRegion" = "Country";
 "addressLine1" = "Address line 1";
 "addressLine2" = "Address line 2";
-"state" = "County";
+"County" = "County";
```
alex-ioja-yang-cko, line 41: I think consistency in hard to navigate files like this should be desired. All other keys are small cased so I'd just go with that. In this context `County` is not a name or title, its just another key in a file full of key-value pairs.

alex-ioja-yang-cko (approved):

deepesh-vasthimal-cko (approved):

PR #222, opened 2022-07-25
PIMOB:1337: Dismiss keyboard on touch on white spaces view for all screens

ehab-al-cko:
## Proposed changes

I should be able to click on the whitespace to exit a fieldThis works on the payment form but not the billing form

## Types of changes

## Checklist

## Further comments


Source/UI/NewUI/BillingForm/View/BillingFormCellTextField.swift
```diff
@@ -38,6 +45,15 @@ final class BillingFormCellTextField: UITableViewCell {
     required init?(coder _: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+
+  @objc private func hideKeyboard(gestureRecognizer: UITapGestureRecognizer) {
+    let view = gestureRecognizer.view
+    let loc = gestureRecognizer.location(in: view)
+    let subview = view?.hitTest(loc, with: nil)
+    guard !(subview is SecureDisplayView) else { return }
+    UIApplication.shared.sendAction(#selector(UIApplication.resignFirstResponder), to: nil, from: nil, for: nil)
```
alex-ioja-yang-cko, line 54: Given we have this method `hideKeyboard` written twice, what if we do something like:

```swift
extension UITapGestureRecognizer {
    
    static var keyboardDismissTap: UITapGestureRecognizer {
        let gesture = UITapGestureRecognizer(target: self, action: #selector(hideKeyboard(gestureRecognizer:)))
        gesture.cancelsTouchesInView = false
        return gesture
    }
    
    @objc private static func hideKeyboard(gestureRecognizer: UITapGestureRecognizer) {
        let view = gestureRecognizer.view
        let loc = gestureRecognizer.location(in: view)
        let subview = view?.hitTest(loc, with: nil)
        guard !(subview is SecureDisplayView) else { return }
        UIApplication.shared.sendAction(#selector(UIApplication.resignFirstResponder), to: nil, from: nil, for: nil)
    }
    
}

// Use site
addGestureRecognizer(UITapGestureRecognizer.keyboardDismissTap)
```

Source/UI/NewUI/BillingForm/View/BillingFormCellTextField.swift
```diff
@@ -21,8 +21,13 @@ final class BillingFormCellTextField: UITableViewCell {
         return view
     }()
 
+    private lazy var tapGesture: UITapGestureRecognizer = {
+      UIView.keyboardDismissTapGesture
+    }()
```
alex-ioja-yang-cko, line 26: This is used in the initialiser. I've raised this previously and still no outcome, but there is no benefit from making property lazy here, or by complicating the assignment to use a completion handler vs direct value. It does however add lines of code that result in unnecessary noise and distractions...

Source/Extensions/UIViewExtension.swift
```diff
@@ -0,0 +1,27 @@
...
+    gesture.cancelsTouchesInView = false
+    return gesture
+  }
+
+  @objc private static func hideKeyboard(gestureRecognizer: UITapGestureRecognizer) {
+    let view = gestureRecognizer.view
+    let loc = gestureRecognizer.location(in: view)
+    let subview = view?.hitTest(loc, with: nil)
+    guard !(subview is SecureDisplayView) else { return }
+    UIApplication.shared.sendAction(#selector(UIApplication.resignFirstResponder), to: nil, from: nil, for: nil)
```
alex-ioja-yang-cko, line 24: 1. Do we need `UIApplication.shared.sendAction(#selector(UIApplication.resignFirstResponder), to: nil, from: nil, for: nil)`, I thought `UIApplication.shared.resignFirstResponder()` might be sufficient?

2. Do we need to check for hit test? I've previously asked whether that view catches the gesture already. Sorry but haven't seen any reply and I still think it would 🤔

Source/UI/NewUI/CommonUI/View/Component/InputView.swift
```diff
@@ -45,9 +45,14 @@ class InputView: UIView {
         return view
     }()
 
+    private lazy var tapGesture: UITapGestureRecognizer = {
+      UIView.keyboardDismissTapGesture
+    }()
```
alex-ioja-yang-cko, line 50: Same long winded way to assign a property.

alex-ioja-yang-cko (approved):
Functionally looks fine, but would still like to clarify some things before merging.

Source/Extensions/UIViewExtension.swift
ehab-al-cko: UIApplication.shared.resignFirstResponder() is not working

Source/UI/NewUI/BillingForm/View/BillingFormCellTextField.swift
ehab-al-cko: `lazy` because the target is `self` in `UIView` extension

alex-ioja-yang-cko: How does that work, its a `static` method in the UIView extension, how does `self` exist? ❓

Source/Extensions/UIViewExtension.swift
```diff
@@ -0,0 +1,27 @@
...
+
+  static var keyboardDismissTapGesture: UITapGestureRecognizer {
+    let gesture = UITapGestureRecognizer(target: UIView.self, action: #selector(hideKeyboard(gestureRecognizer:)))
+    gesture.cancelsTouchesInView = false
+    return gesture
+  }
+
+  @objc private static func hideKeyboard(gestureRecognizer: UITapGestureRecognizer) {
+    let view = gestureRecognizer.view
+    let loc = gestureRecognizer.location(in: view)
```
deepesh-vasthimal-cko, line 21: can you rename `loc` to its full form ?

deepesh-vasthimal-cko (approved):

Source/Extensions/UIViewExtension.swift
alex-ioja-yang-cko: `resignFirstResponder()` was my mistake sorry. I've reviewed the behaviour and it is only working via the `sendAction` dispatch.

I've also tried locally to drop the hitTest and I can see this is a good approach in this situation 👍 . Consider this thread resolved from my side

iOS Example Frame Carthage/iOS Example Frame Carthage.xcodeproj/project.pbxproj
```diff
@@ -292,6 +292,7 @@
 		957323EB286EF50F0068C115 /* applePayTokenInvalid.json in Resources */ = {isa = PBXBuildFile; fileRef = 95732393286EF50F0068C115 /* applePayTokenInvalid.json */; };
 		957323EC286EF50F0068C115 /* cardTokenInvalidNumber.json in Resources */ = {isa = PBXBuildFile; fileRef = 95732394286EF50F0068C115 /* cardTokenInvalidNumber.json */; };
 		957323ED286EF50F0068C115 /* applePayToken.json in Resources */ = {isa = PBXBuildFile; fileRef = 95732395286EF50F0068C115 /* applePayToken.json */; };
+		958826FC28902ED400385E7D /* UIViewExtension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 958826FB28902ED400385E7D /* UIViewExtension.swift */; };
```
alex-ioja-yang-cko, line 295: you can still move the content of this file into the existing file `UIViewExtensions.swift`. We don't need 2 files with extensions

harry-brown-cko (approved):

PR #224, opened 2022-07-27
PIMOB-1278: Add Payment Form Presented analytics event

alex-ioja-yang-cko:
## Proposed changes

Send payment form presented from the new screen.

## Types of changes

## Checklist

* [] I have added necessary documentation (if appropriate)

## Further comments

The events for old UI presentation and new UI presentation will not exist in the same code. And the new UI will not track Theming currently due to the different level of complexity in the new UI theming, that will need to be handled separately at another date. [Source](https://github.com/checkout/frames-ios/pull/224#discussion_r930800685)

Also, I have eagerly added the API key to the factory in my original misguided attempt to create service for sending events. 
I then realised the PaymentViewModel will be the core object of the journey, and it will own a logger. This logger will then be injected into the service that will get created later. So no need for the API key from this PR, can remove if needed, but also tempted to leave as it will definitely be needed once we create the client and I've also _spent time_ documenting the API in code. Happy to remove from this PR if unhappy.


Source/Core/CheckoutAPIService.swift
```diff
@@ -21,15 +21,11 @@ public final class CheckoutAPIService: CheckoutAPIProtocol {
...
-    public convenience init(publicKey: String, environment: Environment) {
+    public init(publicKey: String, environment: Environment) {
         let checkoutAPIService = Checkout.CheckoutAPIService(publicKey: publicKey, environment: environment.checkoutEnvironment)
         let cardValidator = CardValidator(environment: environment.checkoutEnvironment)
         let logger = FramesEventLogger(environment: environment, getCorrelationID: { checkoutAPIService.correlationID })
-
-        self.init(checkoutAPIService: checkoutAPIService, cardValidator: cardValidator, logger: logger)
-    }
-
-    init(checkoutAPIService: Checkout.CheckoutAPIProtocol, cardValidator: CardValidating, logger: FramesEventLogger) {
```
alex-ioja-yang-cko, line 32: This initialiser was only used once, in this object by the convenience initialiser and all it was doing was assigning properties to self 😕. Have simplified.

Source/Core/Logging/FramesLogEvent.swift
```diff
@@ -86,8 +86,12 @@ enum FramesLogEvent: Equatable, PropertyProviding {
...
              .threeDSWebviewPresented:
             return [:]
-        case let .paymentFormPresented(theme, locale):
-            return [.theme: theme.rawProperties, .locale: locale.identifier].mapValues(AnyCodable.init(_:))
+        case let .paymentFormPresented(logTheme):
+            var dictionary: [Property: Any] = [.locale: Locale.current.identifier]
+            if logTheme {
+                dictionary[.theme] = Theme()
+            }
+            return dictionary.mapValues(AnyCodable.init(_:))
```
alex-ioja-yang-cko, line 94: Given the old UI and new will not be present at the same time, might be handy to drop `logTheme` altogether.

New UI theming is too complex to be logged in the same fashion as old one...

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
```diff
@@ -4,17 +4,18 @@ import Checkout
...
-                                                  paymentFormStyle: PaymentFormStyle?,
-                                                  billingFormStyle: BillingFormStyle?,
-                                                  supportedSchemes: [Card.Scheme]) -> UIViewController {
-    let cardValidator = CardValidator(environment: environment.checkoutEnvironment)
+
+  public static func buildViewController(configuration: PaymentFormConfiguration,
+                                         style: PaymentStyle) -> UIViewController {
+    let cardValidator = CardValidator(environment: configuration.environment.checkoutEnvironment)
+    let sessionCorrelationID = UUID().uuidString
+    let logger = FramesEventLogger(environment: configuration.environment, getCorrelationID: { sessionCorrelationID })
```
alex-ioja-yang-cko, line 12: I think this `sessionCorrelationID` might need to be moved into a static private property, just to make sure it persists accurately and doesn't turn into a `nil` 🤔. Will run some local QA to check this.

iOS Example Frame/iOS Example Frame/MainViewController.swift
```diff
@@ -50,12 +50,13 @@ final class MainViewController: UIViewController, CardViewControllerDelegate, Th
 
         let cardFormData = Self.defaultCardFormData()
 
-      let paymentFormViewController = PaymentFormFactory.getPaymentFormViewController(
-        environment: environment,
-        billingFormData: cardFormData.billingForm,
-        paymentFormStyle: cardFormData.paymentFormStyle,
-        billingFormStyle: cardFormData.billingFormStyle,
-        supportedSchemes: [.visa, .mastercard, .maestro])
+        let paymentConfiguration = PaymentFormConfiguration(apiKey: "pk_test_6e40a700-d563-43cd-89d0-f9bb17d35e73",
```
alex-ioja-yang-cko, line 53: For reference, this key is already hardcoded in the object on line 25 (no code changes to have it appear). As documented in PR comment, this is not currently used and was the result of my misguided initial approach to the logging.

I do however think the task where this is consumed will show up very soon as we need the network client before this is completed.

Source/Core/Logging/FramesLogEvent.swift
patrick-hoban-cko: We'll handle new UI theming logging separately. 

We probably need to revisit how theming as set, as now we have so many configurations per component, we'll get messy / oversaturated logs.

alex-ioja-yang-cko: But are we certain the old UI will not be shown alongside the new UI at all? Because then I'd be able to start removing some things that were created for logging the old UI.

patrick-hoban-cko: For 4.0.0 we're going to delete old UI methods, so if you want to use the old UI you'll need to use 3.5.3.

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
deepesh-vasthimal-cko: Would be also good to check existing frames 3.5.3 behaviour.

Source/UI/NewUI/PaymentForm/Model/PaymentFormConfiguration.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  PaymentFormConfiguration.swift
+//  
+//
+//  Created by Alex Ioja-Yang on 26/07/2022.
+//
+
+import Foundation
+
+public struct PaymentFormConfiguration {
```
deepesh-vasthimal-cko, line 10: 👍

deepesh-vasthimal-cko: If any changes made to public configuration should be agreed with devs and also updated on the API spec.

Source/UI/NewUI/PaymentForm/Model/PaymentStyle.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  PaymentStyle.swift
+//  
+//
+//  Created by Alex Ioja-Yang on 26/07/2022.
+//
+
+import Foundation
+
+public struct PaymentStyle {
```
deepesh-vasthimal-cko, line 10: Is there a reason for this to be public, Do merchants use this for default UI set up? 
I can see that cardFormData is assigned in MainViewcontroller as  `Self.defaultCardFormData()` so it gets the style from the there. Why declare this public ?

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
alex-ioja-yang-cko: Yes, this is the behaviour that was happening before, I have checked how it was done and found it to be excessively convoluted for no apparent reason.

The correlation ID was not persisted anywhere, it was just kept in memory via a complex chain of `logger -> service -> log manager -> static UUID string`. You can review source of it all inside `LogManager.swift`.

Only behaviour change is that I'm not enforcing lowercasing as it was previously done. Would be quite curious to know why we would willingly reduce our uniqueness chance of the correlation id?

Source/UI/NewUI/PaymentForm/Model/PaymentFormConfiguration.swift
deepesh-vasthimal-cko: refer : https://checkout.atlassian.net/wiki/spaces/FS/pages/5266440242/Draft+Frames+Component+API+Spec

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
alex-ioja-yang-cko: Additionally, the PaymentForm (via its VC & VM) are the core owners of the payment journey. Anything presented after them will be able to share the state they have (ie: logger correlation id).

If they are dismissed (ie: user cancels intention to make payment) then in this current approach we do generate a new correlation ID for their second start of the journey.

Big ❗ here, the event for payment billing form being shown is marked as unique, so will only be reported once per correlation id. In practice here, that means if we persist that id we lose insights:
- open payment form - 🟢 
- change mind and cancel 🔴 
- open payment form 🔴 
- open billing form 🟢 
- change mind and cancel all 🔴 
- open payment form 🔴 
- open billing form 🔴 
- make payment 🟢 

There is potential for large time gaps between these events as the user may be doing online shopping or generally interacting further with the application. To us it may appear that from start to end it took user 20 minutes, when in reality they dropped twice at different levels of the journey.

Source/UI/NewUI/PaymentForm/Model/PaymentFormConfiguration.swift
alex-ioja-yang-cko: Please advise me on what to do and what requirements are there to be met ❓

Source/UI/NewUI/PaymentForm/Model/PaymentStyle.swift
alex-ioja-yang-cko: We previously had multiple objects passed into the factory to build the root vc for the journey. I took those objects and split them into 2 which should better guide consumer on setting up.

As such, these groups are:
- configuration (made of default billing data and other logical components)
- UI ☝️ (made of styling for the UI components)

I did not have time to refactor `Self.defaultCardFormData`, I can do that as well if this approach is accepted (no point to go back & forward on everything)

Source/UI/NewUI/PaymentForm/Model/PaymentFormConfiguration.swift
alex-ioja-yang-cko: Looking at that doc its so out of sync with the state of frames that I might not be the best person to add this change there. There's just no reference point to associate this with 😵

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
deepesh-vasthimal-cko: Good point we can track all the red events as well with a unique correlation id, But we then we need to track cancel event if we want to make more sense with our logs ( otherwise we can just have presented logs fired up more times  for first release @patrick-hoban-cko is that ok with you ? ).

Source/UI/NewUI/PaymentForm/Model/PaymentFormConfiguration.swift
deepesh-vasthimal-cko: @ehab-al-cko can help.

Source/UI/NewUI/PaymentForm/Model/PaymentStyle.swift
deepesh-vasthimal-cko: This will also need to be added to API spec, @ehab-al-cko can you please add this to the API spec as well.

deepesh-vasthimal-cko (approved):

ehab-al-cko (changes requested):

Cartfile.resolved
```diff
@@ -1,2 +1,2 @@
 github "checkout/checkout-event-logger-ios-framework" "1.2.0"
-github "marmelroy/PhoneNumberKit" "3.3.7"
+github "marmelroy/PhoneNumberKit" "3.4.4"
```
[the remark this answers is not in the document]
alex-ioja-yang-cko: huh, interesting 🤔

ehab-al-cko (approved):

PR #228, opened 2022-07-28
PIMOB-987: Add event for start of payment journey

alex-ioja-yang-cko:
## Proposed changes

[Jira task](https://checkout.atlassian.net/browse/PIMOB-987)

🟢 Add event to track start of payment journey
🔴 Did not add unit tests as they would require opening up and injecting a logger inside the builder. That would very much invalidate the purpose of the builder to be responsible for creating everything
🟡 We may be able to test if we wanted the builder to have state and limited the access for consumer. See code example
```swift
// not testable as we cannot inject state inside generateContent
Builder.generateContent()

// testable as we can use internal access control to override properties that will be used
Builder().generateContent()
```

## Further comments

Since the changes have been so insignificant, have allowed some SwiftLint fixes to blend in. Must have been introduced in the PR between starting the SwiftLint branch and its merging into main.

### DIFF
--- a/Source/Core/Logging/FramesLogEvent.swift
+++ b/Source/Core/Logging/FramesLogEvent.swift
@@ -35,7 +35,7 @@ enum FramesLogEvent: Equatable, PropertyProviding {
         case success
     }
 
-    case checkoutAPIClientInitialised(environment: Environment)
+    case paymentJourneyStart(environment: Environment)
     case paymentFormPresented
     case billingFormPresented
     case threeDSWebviewPresented
@@ -49,8 +49,8 @@ enum FramesLogEvent: Equatable, PropertyProviding {
 
     private var typeIdentifierSuffix: String {
         switch self {
-        case .checkoutAPIClientInitialised:
-            return "checkout_api_client_initialised"
+        case .paymentJourneyStart:
+            return "payment_journey_start"
         case .paymentFormPresented:
             return "payment_form_presented"
         case .billingFormPresented:
@@ -68,7 +68,7 @@ enum FramesLogEvent: Equatable, PropertyProviding {
 
     var monitoringLevel: MonitoringLevel {
         switch self {
-        case .checkoutAPIClientInitialised,
+        case .paymentJourneyStart,
              .paymentFormPresented,
              .billingFormPresented,
              .threeDSWebviewPresented:
@@ -88,7 +88,7 @@ enum FramesLogEvent: Equatable, PropertyProviding {
             return [:]
         case .paymentFormPresented:
             return [Property.locale: Locale.current.identifier].mapValues(AnyCodable.init(_:))
-        case let .checkoutAPIClientInitialised(environment):
+        case let .paymentJourneyStart(environment):
             let environmentString = environment.rawValue == "live" ? "production" : environment.rawValue
             return [.environment: environmentString].mapValues(AnyCodable.init(_:))
         case let .threeDSChallengeLoaded(success):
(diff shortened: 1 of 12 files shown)


Source/Core/Logging/FramesLogEvent.swift
```diff
@@ -49,8 +49,8 @@ enum FramesLogEvent: Equatable, PropertyProviding {
 
     private var typeIdentifierSuffix: String {
         switch self {
-        case .checkoutAPIClientInitialised:
-            return "checkout_api_client_initialised"
+        case .paymentJourneyStart:
+            return "payment_journey_start"
```
alex-ioja-yang-cko, line 53: @deepesh-vasthimal-cko & @patrick-hoban-cko follow up on the Slack discussion, let me know if I was convincing on the need to separate this event name from syntactically getting tied to the VC init and whether you have alternative names for event.

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
```diff
@@ -13,6 +13,7 @@ public struct PaymentFormFactory {
     // Ensure a consistent identifier is used for the monitoring of a journey
     Self.sessionCorrelationID = UUID().uuidString
     let logger = FramesEventLogger(environment: configuration.environment, getCorrelationID: { Self.sessionCorrelationID })
+    logger.log(.paymentJourneyStart(environment: configuration.environment))
```
alex-ioja-yang-cko, line 16: @ehab-al-cko , @deepesh-vasthimal-cko see PR description for why this is not testable in current call site syntax. Let me know if you want to move from
```swift
PaymentFormFactory. buildViewController

// TO

PaymentFormFactory(). buildViewController
```

that apparent minor change makes all the difference :)

Tests/Mocks/MockCardValidator.swift
```diff
@@ -23,11 +23,17 @@ final class MockCardValidator: CardValidating {
...
 
     var validateCVVToReturn: ValidationResult<ValidationError.CVV> = .success
+    var expectedMaxLenghtCVV = 4
+    var maxLenghtCVVCalledWith = [Card.Scheme]()
     private(set) var validateCVVCalledWith: (cvv: String, cardScheme: Card.Scheme)?
     func validate(cvv: String, cardScheme: Card.Scheme) -> ValidationResult<ValidationError.CVV> {
         validateCVVCalledWith = (cvv, cardScheme)
         return validateCVVToReturn
     }
+    func maxLenghtCVV(for scheme: Card.Scheme) -> Int {
```
patrick-hoban-cko, line 33: Sorry to be picky @alex-ioja-yang-cko, but there are a number of places in these fields where Length is spelt incorrectly as _Lenght_

Could we update that please?

Source/Core/Logging/FramesLogEvent.swift
patrick-hoban-cko: I'll let @deepesh-vasthimal-cko give his steer.

Tests/Mocks/MockCardValidator.swift
alex-ioja-yang-cko: waaaait, I'm working on different task on wrong branch, this should not be here 🙈. But will fix that in correct place once I steer it back 👍

alex-ioja-yang-cko: ✅ please look again, this time its just what's supposed to be here

Source/Core/Logging/FramesLogEvent.swift
patrick-hoban-cko: @alex-ioja-yang-cko let's go with _payment_form_initialised_

That means we have more choice later on if there's another way to begin the payment journey in the future (for example, what if we have some other sort of integration option of the SDKs, I'd want to be distinct about which event I'm actually looking at).

patrick-hoban-cko (approved):

ehab-al-cko (approved):

PR #234, opened 2022-08-02
PIMOB-1359: Visa card should allow 19-digit length

alex-ioja-yang-cko:
## Proposed changes

[Jira task](https://checkout.atlassian.net/browse/PIMOB-1359)

## Types of changes

## Checklist

## Further comments

Small change to enable VISA cards to input up to 19 characters


Checkout/Checkout/Source/Model/Card+Scheme.swift
```diff
@@ -138,8 +138,9 @@ extension Card {
 
     var maxCardLength: Int {
       switch self {
+      case .visa:
+          return 19
```
ehab-al-cko, line 142: i am wondering about this logic, it should allow 19 or 16 or 19 only?

alex-ioja-yang-cko: Most Visa cards are 16 digits. Apparently they can also be 19 characters long. So both 16 and 19 are valid, and the regex seems to already support 16 and 19. The only issue was with allowing input to reach 19.

patrick-hoban-cko: Yep - looks okay, as that variable sets the _maxCardLength_, which in VISA's case, is indeed 19.

deepesh-vasthimal-cko (approved):

ehab-al-cko:
@alex-ioja-yang-cko  
Kind reminder, please change `alex/` to `bugfix/` or `feature/` next time. we are following `git flow`

ehab-al-cko (approved):

ehab-al-cko:
@alex-ioja-yang-cko  On merge please change commit to 
`Merge pull request #234 from checkout/bugfix/PIMOB-1359-i-os-checkout-i-os-card-validator-should-allow-visa-cards-of-19-digit-length`

PR #236, opened 2022-08-02
PIMOB-1365: Integrate supported card scheme list with all UI components

ehab-al-cko:
[Jira ticket](https://checkout.atlassian.net/browse/PIMOB-1365?atlOrigin=eyJpIjoiZDM2MjYyNGM5OGY1NDBiMzgxMGIwOWI4Mzk2NDM2ODMiLCJwIjoiaiJ9)

## Proposed changes

- Header view icons now is depending on supported card scheme
- Card number validation now is depending on supported card scheme

## Types of changes

## Checklist

## Further comments


iOS Example Frame/iOS Example Frame/MainViewController.swift
```diff
@@ -97,7 +97,15 @@ final class MainViewController: UIViewController, CardViewControllerDelegate, Th
 
         let paymentConfiguration = PaymentFormConfiguration(apiKey: "pk_test_6e40a700-d563-43cd-89d0-f9bb17d35e73",
                                                             environment: environment,
-                                                            supportedSchemes: [.visa, .mastercard, .maestro],
+                                                            supportedSchemes: [.visa, .visa, .visa,
```
ehab-al-cko, line 100: multi .visa is added like to this to show uniqueness of supported scheme

patrick-hoban-cko:
Should we also set the text for the string above the accepted card scheme icons, using the enum of card schemes that the merchant accepts? 

Then it's consistent with the designs. We should make this configurable as to whether the merchant wants to show this too.

image

ehab-al-cko:
@patrick-hoban-cko Merchant can provide the localized string for them. what do you think? 
> Should we also set the text for the string above the accepted card scheme icons, using the enum of card schemes that the merchant accepts?
> 
> Then it's consistent with the designs. We should make this configurable as to whether the merchant wants to show this too.
> 
> image

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
```diff
@@ -34,18 +36,15 @@ extension CardNumberViewModel: CardNumberViewModelProtocol {
...
-      return Constants.Bundle.SchemeIcon(scheme: scheme)
+        // An unknown scheme can also be generated when an eager validation was matched
+        //    but the number is not complete.
+        // In this case we only update delegate if we have a final result to avoiding
+        //    overriding the eager validation with a less informative update
+      case .success((let isComplete, let scheme)):
+        guard isComplete, supportedSchemes.contains(scheme) else { return nil }
+        delegate?.update(cardNumber: cardNumber, scheme: scheme)
+        return Constants.Bundle.SchemeIcon(scheme: scheme)
+      case .failure: return nil
```
harry-brown-cko, line 47:
```suggestion
      case let .success((isComplete, scheme)) where isComplete && supportedSchemes.contains(scheme):
        delegate?.update(cardNumber: cardNumber, scheme: scheme)
        return Constants.Bundle.SchemeIcon(scheme: scheme)
      default: return nil
```

Source/UI/NewUI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
```diff
@@ -27,7 +27,7 @@ class DefaultPaymentViewModel: PaymentViewModel {
        paymentFormStyle: PaymentFormStyle?,
        billingFormStyle: BillingFormStyle?,
        supportedSchemes: [Card.Scheme]) {
-    self.supportedSchemes = supportedSchemes
+    self.supportedSchemes = NSOrderedSet(array: supportedSchemes).map({ $0 as! Card.Scheme })
```
harry-brown-cko, line 30: compiles fine without this change

```suggestion
    self.supportedSchemes = supportedSchemes
```

harry-brown-cko: if you want to remove duplicates you can also remove the force cast
```suggestion
    self.supportedSchemes = Array(_immutableCocoaArray: NSOrderedSet(array: supportedSchemes))
```

ehab-al-cko: @harry-brown-cko i didn't find `_immutableCocoaArray` in swift. please explain more. thanks

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
harry-brown-cko: My main point is that we can integrate the check for `isComplete` and `supportedSchemes.contains(scheme)` into the switch, cleaning up the branching here. Could do something like this instead:
```suggestion
      case let .success((isComplete, scheme)) where isComplete && supportedSchemes.contains(scheme):
        delegate?.update(cardNumber: cardNumber, scheme: scheme)
        return Constants.Bundle.SchemeIcon(scheme: scheme)
      case .success, .failure: return nil
```

Source/UI/NewUI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
harry-brown-cko: yeah was just testing stuff and xcode suggested that init. how about this instead? Saves the mapping step
```suggestion
    self.supportedSchemes = NSOrderedSet(array: supportedSchemes).array as! [Card.Scheme]
```

alex-ioja-yang-cko: Sorry to hop in, but whatever you go with could you please remove the forced cast? Removing duplicates is a caring fix if consumer mistakenly adds duplicate inputs, but if the price to that is force casting, we might want to review value of the feature.

I would add [OptionSet](https://www.hackingwithswift.com/example-code/language/how-to-create-a-custom-optionset) for consideration on this. There can be no duplicates if input is an option set

Source/UI/NewUI/PaymentForm/StyleProtocol/PaymentHeaderCellStyle.swift
```diff
@@ -9,8 +9,8 @@
 import UIKit
 
 public protocol PaymentHeaderCellStyle {
+    var iconsIsHidden: Bool { get }
```
deepesh-vasthimal-cko, line 12: suggestion : rename to `isIconsHidden`

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
```diff
@@ -40,13 +40,10 @@ struct PaymentFormStyleCustom1: PaymentFormStyle {
 // MARK: - Header View
 
 struct PaymentHeaderCellStyleCustom1: PaymentHeaderCellStyle {
+  var iconsIsHidden: Bool = true
   var backgroundColor = Constants.redBackgroundColor
   var headerLabel: ElementStyle? = HeaderLabelFormStyleCustom1(text: "Payment details")
-  public var subtitleLabel: ElementStyle?
-  public var schemeIcons: [UIImage?] = [
```
deepesh-vasthimal-cko, line 46: 👍

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
```diff
@@ -133,26 +130,26 @@ struct AddBillingDetailsButtonStyleCustom1: ElementButtonStyle {
...
-public struct ExpiryDateFormStyleCustom1: CellTextFieldStyle {
-  public var isMandatory = true
-  public var backgroundColor: UIColor = .clear
-  public var title: ElementStyle? = TitleLabelStyleCustom1(text: "Expiry date")
-  public var hint: ElementStyle? = HintInputLabelStyleCustom1(text: "Format is MM/YY")
-  public var mandatory: ElementStyle?
-  public var textfield: ElementTextFieldStyle = TextFieldCustom1(text: "test", placeHolder: "MM/YY")
-  public var error: ElementErrorViewStyle? = ErrorInputLabelStyleCustom1(text: "please fill expiry date")
+struct ExpiryDateFormStyleCustom1: CellTextFieldStyle {
+  var isMandatory = true
```
deepesh-vasthimal-cko, line 134: Unrelated to this PR : But for Card Number & Expiry this value should always be mandatory by default ( not allow merchants to hide this ). My understanding is that CVV can be empty / optional for few card schemes.

iOS Example Frame/iOS Example Frame/MainViewController.swift
deepesh-vasthimal-cko: Does this include visa just once then ?

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
patrick-hoban-cko: Yes that's correct. Card number and expiry date are mandatory, but we should allow Security Code to be an optional field, as merchant's don't actually have to include this.

alex-ioja-yang-cko: fyi, there is no simple solution for this, as these objects appear more to be designed for internal consumption, not for external. I would add that you have additional similar issues coming from `isHidden` & `isEnabled` which don't really make sense to be inputed by consumer?

Source/UI/NewUI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
ehab-al-cko: I suggest 
```suggestion
    self.supportedSchemes = NSOrderedSet(array: supportedSchemes).array as? [Card.Scheme] ?? [.unknown]
```

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
ehab-al-cko: Please suggest code

iOS Example Frame/iOS Example Frame/MainViewController.swift
ehab-al-cko: yes, unique schemes

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
deepesh-vasthimal-cko: Good Question @alex-ioja-yang-cko : I think we should review & discuss the purpose of isMandatory, isHidden & isEnabled and then create a separate story to fix if needed. My understanding is as below. 

- `isMandatory` is allowing the merchant to show a asterisk sign on field, Does it have any more logic associated with it ? 
- `isHidden` is allowing a merchant to show and hide a field. In this case we don't need this for expiry date and card number fields as merchant needs these fields for create tokens. 
- `isEnabled` allows merchants to edit a particular text field. I am assuming we dont want our merchants to control when the text fields are enabled or disabled for editing as we are using our validation logic.

Source/UI/NewUI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
harry-brown-cko: Looks better. Last question, why are we using `[.unknown]` as the fallback instead of just `[]`?

ehab-al-cko: As default is `.uknown`, should i change it?

Source/UI/NewUI/PaymentForm/StyleProtocol/PaymentHeaderCellStyle.swift
```diff
@@ -9,8 +9,8 @@
 import UIKit
 
 public protocol PaymentHeaderCellStyle {
+    var isIconsViewHidden: Bool { get }
```
alex-ioja-yang-cko, line 12: The naming in my opinion is leaking implementation details. Also is being negative whereas I believe we should always strive for positive details. My requirements for this name would be:
- don't mention icons view or UI components
- try to be positive and not negative (show vs hide, enable vs disable)

My suggestion would still be something in the lines of `showAcceptedCardsList`, but happy for other ideas as long as they tick those 2 points

Source/UI/NewUI/PaymentForm/View/PaymentHeaderView.swift
```diff
@@ -71,14 +72,10 @@ public final class PaymentHeaderView: UIView {
...
-      iconsStackView.isHidden = true
-      return
-    }
-    iconsStackView.isHidden = false
-    for image in icons where image != nil {
+    iconsStackView.isHidden = style?.isIconsViewHidden ?? true
+    for scheme in supportedSchemes {
       let imageView = UIImageView().disabledAutoresizingIntoConstraints()
-      imageView.image = image?.withRenderingMode(.alwaysOriginal)
+      imageView.image = Constants.Bundle.SchemeIcon(scheme: scheme).image?.withRenderingMode(.alwaysOriginal)
```
alex-ioja-yang-cko, line 78: You really should remove all the images that exist in the stack **before** adding new ones. I know you mentioned this list is not expected to be updated (_yet_), but if your lifecycle calls the same method multiple times (ie: even simple things like scrolling in & out of view), you end up keep adding images.

Source/Extensions/UIStackViewExtension.swift
```diff
@@ -0,0 +1,21 @@
...
+
+extension UIStackView {
+
+    func removeAllArrangedSubviews() {
+        let removedSubviews = arrangedSubviews.reduce([]) { (allSubviews, subview) -> [UIView] in
+            self.removeArrangedSubview(subview)
+            return allSubviews + [subview]
+        }
+        NSLayoutConstraint.deactivate(removedSubviews.flatMap({ $0.constraints }))
+        removedSubviews.forEach({ $0.removeFromSuperview() })
```
alex-ioja-yang-cko, line 19: Do we need to deactivate constraints? I would think they should automatically be removed?

I also found from UIKit description that:
```swift
    /* Removes a subview from the list of arranged subviews without removing it as
     a subview of the receiver.
        To remove the view as a subview, send it -removeFromSuperview as usual;
     the relevant UIStackView will remove it from its arrangedSubviews list
     automatically.
     */
    open func removeArrangedSubview(_ view: UIView)
```

With these in mind, I feel this could be optimised to:
```suggestion
    func removeSubviews() {
        subviews.forEach { $0.removeFromSuperview() }
    }
```

And then once its generic and functionality is not strictly related to UIStackView, we can move to UIViewExtensions file

Source/UI/NewUI/PaymentForm/View/PaymentHeaderView.swift
```diff
@@ -71,14 +72,12 @@ public final class PaymentHeaderView: UIView {
   }
 
   private func updateIconsStack() {
-    guard let icons = style?.schemeIcons else {
-      iconsStackView.isHidden = true
-      return
-    }
-    iconsStackView.isHidden = false
-    for image in icons where image != nil {
+    iconsStackView.isHidden = style?.showAcceptedCardsList ?? true
```
alex-ioja-yang-cko, line 75:
```suggestion
    iconsStackView.isHidden = style?.showAcceptedCardsList != true
```

alex-ioja-yang-cko (commented):
Thank you for agreeing on previous suggestions, but I am not convinced the way these changes have been made is correct. Please let me know what you think.

Source/UI/NewUI/PaymentForm/View/PaymentHeaderView.swift
ehab-al-cko: I don't think this right solution

🟢 expected result = true  

- if style?.showAcceptedCardsList = nil 
`nil != true`  
then result  = true 

🔴 expected result = true

- if style?.showAcceptedCardsList = true 
`true != true`  
then result  = false 

🔴 expected result = false

- if style?.showAcceptedCardsList = false 
`true != false`  
then result  = true

alex-ioja-yang-cko: I think that is exactly the right solution :)

if style?.showAcceptedCardsList = nil
nil != true
then result = true `=> Hide icon stack view when showAcceptedCardsList is nil`

if style?.showAcceptedCardsList = true
true != true
then result = false `=> Show icon stack view when showAcceptedCardsList is true`

if style?.showAcceptedCardsList = false
false != true
then result = true `=> Hide icon stack view when showAcceptedCardsList is false`

alex-ioja-yang-cko: Look at it differently, 
> if style says **SHOW**, UI responds **HIDE**
Basically the assignment to `.isHidden` should be the opposite of `show`:
`view.isHidden = !showSomething`.

And as you pointed out, my suggestion also covers correctly the nil case.

ehab-al-cko: I got confused more :D  I think`?? true` is  more simple.

alex-ioja-yang-cko: That is one of the issues of our acceptance of excessive optionality. I think we can definitely improve on that across the new code.
***
`?? true` may seem clearer for now, but your code is currently using the opposite logic. 

If it wants to show card list its hiding the list. 
If it doesn't want to show card list it will show the list.
I am not sure whether you can negate an optional assignment with that syntax without adding some form of comparison. And when you do, my suggestion might make more sense :)

deepesh-vasthimal-cko: @ehab-al-cko are you planning to make any changes on this thread ? Otherwise all looks good.

ehab-al-cko: I would like to keep it as default value for now :)

alex-ioja-yang-cko: I am not sure this is about preference, its about behaviour at this stage. Your current code does the wrong thing, I'm not sure if you've ran it?

harry-brown-cko (approved):

Source/UI/NewUI/PaymentForm/View/PaymentHeaderView.swift
ehab-al-cko: Naming `isHidden = show` cause confusion. I have choose better naming.

alex-ioja-yang-cko: I don't agree with the new change, I have previously suggested [being positive](https://github.com/checkout/frames-ios/pull/236#discussion_r938716528).

We want the language to be clear to consumer, and at our implementation level we absolutely should think clearly what we're doing to map correctly. This is what my comment was for, fixing implementation to map correctly.

ehab-al-cko: Thanks for suggestion. I don't agree this is negative approach. :) i would like to keep it for avoid confusion.

PR #238, opened 2022-08-03
PIMOB-1332: Add Matrix theming

alex-ioja-yang-cko:
## Proposed changes

[Jira task](https://checkout.atlassian.net/browse/PIMOB-1332)

✅ Update theme
✅ Add theme assets **to demo application**
✅ Remove calls to load SDK font from the application

[Matrix design source](https://www.figma.com/file/ABFqgxna0VH5TcfMgwnQHs/02-•-Vision?node-id=1811%3A8765)


patrick-hoban-cko:
image

Nice work

iOS Example Frame/iOS Example Frame/MainViewController.swift
```diff
@@ -39,11 +39,6 @@ final class MainViewController: UIViewController, CardViewControllerDelegate, Th
         cardViewController?.addressViewController.setCountrySelected(country: Self.countryGB)
     }
 
-    override func viewDidLoad() {
-        super.viewDidLoad()
-        UIFont.loadAllCheckoutFonts
```
ehab-al-cko, line 44: project is using same fonts for Frames, please undelete it

iOS Example Frame/iOS Example Frame/MainViewController.swift
```diff
@@ -76,7 +70,7 @@ final class MainViewController: UIViewController, CardViewControllerDelegate, Th
         let paymentConfiguration = PaymentFormConfiguration(apiKey: "pk_test_6e40a700-d563-43cd-89d0-f9bb17d35e73",
                                                             environment: environment,
                                                             supportedSchemes: [.visa, .mastercard, .maestro],
-                                                            billingFormData: billingForm)
```
ehab-al-cko, line 79: I think, this needs to be filled with `billingForm`

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
```diff
@@ -11,14 +11,14 @@ import Frames
 
 // MARK: - Color Constants -
 
-/// Predefined custom 1 theme colors
 private enum Constants {
-  static let fontColorLabel = UIColor(red: 40 / 255.0, green: 46 / 255.0, blue: 54 / 255.0, alpha: 1.0)
-  static let grayBackgroundColor = UIColor(red: 240 / 255.0, green: 245 / 255.0, blue: 249 / 255.0, alpha: 1.0)
-  static let whiteBackgroundColor = UIColor(red: 255 / 255.0, green: 255 / 255.0, blue: 255 / 255.0, alpha: 1.0)
-  static let errorLabelBackgroundColor = UIColor(red: 226 / 255.0, green: 27 / 255.0, blue: 46 / 255.0, alpha: 1.0)
```
ehab-al-cko, line 19: please apply same changes to `Style2`

ehab-al-cko (changes requested):

ehab-al-cko:
After testing on simulator, 

Payment form 
- when scroll to up, the navigation bar background is transparent
- summary gray line should be green

- error view missing the image 
Screenshot 2022-08-03 at 8 47 12 pm

**Billing form** 
- All input fields are showing numeric key pad
- Country field is not showing current country like United Kingdom
-  All errors are hidden 
- Optional fields are not optionals anymore

iOS Example Frame/iOS Example Frame/MainViewController.swift
alex-ioja-yang-cko: ✅ I think I misunderstood what you meant [here](https://checkout.slack.com/archives/GS5KC6HU3/p1659452713661439?thread_ts=1659452584.733489&cid=GS5KC6HU3) then. It appeared to me font is registered by SDK so no more need for this.

iOS Example Frame/iOS Example Frame/MainViewController.swift
alex-ioja-yang-cko: The design shows no pre filled billing form data. The test data is still there just not passed at this stage of the PR to match design layout.

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
alex-ioja-yang-cko: I am unsure what changes you want to apply? If using a different type of color naming logic (usage vs describing), I don't think the themes need to be in sync. After all this is demo implementation and consumer can decide how they prefer to name and structure their objects.

I would be more concerned about the verbosity of it all than UI creation patterns.

iOS Example Frame Carthage/iOS Example Frame Carthage.xcodeproj/project.pbxproj
```diff
@@ -23,6 +23,7 @@
 /* Begin PBXBuildFile section */
```
alex-ioja-yang-cko, line 23: @ehab-al-cko 

> Payment form
> 1. when scroll to up, the navigation bar background is transparent
> 2. summary gray line should be green
> 3. error view missing the image

1. ✅ background fixed
2. 🔴 summary gray line is not in the design and I personally do not believe it should be green. Task was to implement design, would be happy to go through design decisions with the designer :)
3. 🔴 you are assuming every error should show an image. Similar to previous comment, given the design does not show errors I think we should be prepared that consumer might want to not use image


> Billing form
> 1. All input fields are showing numeric key pad
> 2. Country field is not showing current country like United Kingdom
> 3. All errors are hidden
> 4. Optional fields are not optionals anymore

1. ✅ Thank you, missed that in reusing components. Fixed now
2. ❓ Is it supposed to do that? Only reason its been doing it so far is the existence of pre-filled data, which I've purposely left out now. Not sure how you plan to achieve this, may be worth catching up over it 🤔 
3. 🟡 The design doesn't show any errors, do we consider consumer has to provide errors for all the input fields?
4. ❓ I am not sure what optional fields are non optional anymore means. I've just tested it and everything that is optional is optional and everything that is mandatory is mandatory. Can you please detail?

alex-ioja-yang-cko: I would further dig into options being hidden on Billing form. When are they supposed to be shown as my understanding if the flow would be:
- all included - Done button enabled
- any missing - Done button disabled

I know I must be misunderstanding, but not sure where and how?

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
ehab-al-cko: `CustomStyle1` is refactored. `CustomStyle2` is following the same pattern for `CustomStyle1` because this the sample guide for the merchant.

iOS Example Frame Carthage/iOS Example Frame Carthage.xcodeproj/project.pbxproj
ehab-al-cko: If it's optional field then it can be empty or filled then `done` will be enabled anyway.

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
alex-ioja-yang-cko: There are many ways to create and declare UI, I don't think the way I did for CustomStyle1 is any better than CustomStyle2 and do not think it needs to be identical. 

Different UIs have different problems and the way to reach their needs may be different.

@patrick-hoban-cko , @deepesh-vasthimal-cko ?

iOS Example Frame Carthage/iOS Example Frame Carthage.xcodeproj/project.pbxproj
alex-ioja-yang-cko: The problem is the buttons enabled / disabled states were not passed accurately then. Please see [latest commit](https://github.com/checkout/frames-ios/pull/238/commits/11039c12e279bc5fd94884c29cce70f221c09977) which fixes that in the current logic.

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
patrick-hoban-cko: Happy to have different 'flavours' of how the code does customisation, if there are different ways to getting to the answer

Things just need to be pretty simple - let's discuss tomorrow.

ehab-al-cko (approved):

deepesh-vasthimal-cko (approved):

PR #241, opened 2022-08-07
PIMBOB-1265: Create Card token using sample app with new UI

ehab-al-cko:
[Jira ticket](https://checkout.atlassian.net/browse/PIMOB-1265)
## Proposed changes

## Types of changes

## Checklist

## Further comments


Source/UI/NewUI/CommonUI/View/Core/ButtonView.swift
```diff
@@ -39,6 +41,10 @@ class ButtonView: UIView {
     func update(with style: ElementButtonStyle) {
         self.style = style
         backgroundColor = style.backgroundColor
+        clipsToBounds = true
+        layer.borderColor = style.normalBorderColor.cgColor
+        layer.cornerRadius = style.cornerRadius
+        layer.borderWidth = style.borderWidth
```
alex-ioja-yang-cko, line 47: Does this now mean that UIs that use `SelectionButtonView` will now have double border, one around SelectionButtonView and another around the actual button?

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
```diff
@@ -9,30 +9,28 @@ public struct PaymentFormFactory {
   internal static var sessionCorrelationID = ""
 
   public static func buildViewController(configuration: PaymentFormConfiguration,
-                                         style: PaymentStyle) -> UIViewController {
+                                         style: PaymentStyle,
+                                         cardTokenResultClousure: @escaping (Result<TokenDetails, TokenisationError.TokenRequest>) -> Void) -> UIViewController {
```
alex-ioja-yang-cko, line 13: Any specific reason why you chose a completion handler over a delegate for this callback?

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
```diff
@@ -9,7 +9,7 @@
 import Checkout
 
 protocol CardNumberViewModelDelegate: AnyObject {
-  func update(cardNumber: String, scheme: Card.Scheme)
+  func update(cardNumber: String?, scheme: Card.Scheme?)
```
alex-ioja-yang-cko, line 12: I thought we were using `.unknown` when a scheme could not be determined?

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
```diff
@@ -35,13 +35,15 @@ extension CardNumberViewModel: CardNumberViewModelProtocol {
...
     case .failure:
+        delegate?.update(cardNumber: nil, scheme: nil)
         return nil
     // An unknown scheme can also be generated when an eager validation was matched
     //    but the number is not complete.
     // In this case we only update delegate if we have a final result to avoiding
     //    overriding the eager validation with a less informative update
     case .success((let isComplete, let scheme)):
       guard isComplete else {
+          delegate?.update(cardNumber: nil, scheme: nil)
```
alex-ioja-yang-cko, line 46: There are a lot of nils passed around. Why can't we just use the `cardNumber` as its been determined and different scheme, like `unknown`? Seems excessive to pass nil so frequently.

Source/Resources/en.lproj/Localizable.strings
```diff
@@ -24,6 +24,7 @@
 "PaymentHeaderSubtitle" = "Visa, Mastercard and American Express accepted.";
 "CardNumber" = "Card number";
 "CardNumberErrorMessage" = "Please enter a valid card number";
+"Pay £99.99" = "Pay £99.99";
```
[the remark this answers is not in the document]
ehab-al-cko: We have agreed with Alexander the designer :) to have only `Pay` without amount. I am gonna update it

Checkout/Checkout/Source/Model/Card.swift
```diff
@@ -8,14 +8,14 @@
...
-  public let phone: Phone?
+  public var number: String?
+  public var expiryDate: ExpiryDate?
+  public var name: String?
+  public var cvv: String?
+  public var billingAddress: Address?
+  public var phone: Phone?
 
-  public init(number: String, expiryDate: ExpiryDate, name: String?, cvv: String?, billingAddress: Address?, phone: Phone?) {
+  public init(number: String? = nil, expiryDate: ExpiryDate? = nil, name: String? = nil, cvv: String? = nil, billingAddress: Address? = nil, phone: Phone? = nil) {
```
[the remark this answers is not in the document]
ehab-al-cko: IMO, this suggestions is little bit organising, and it is not solve the current challenge. 

Current changes:
- Change `let` to `var` to be able to set it alone like card.cvv = 123
- Accept nil values for `number` and `expiryDate` to be able to set to `nil`
- add default values to be able to initialised `Card()` without setter for every value

Checkout/Checkout/Source/Model/TokenDetails.swift
```diff
@@ -34,5 +34,29 @@ public struct TokenDetails: Equatable {
...
+    self.bin = bin
+    self.cardType = cardType
+    self.cardCategory = cardCategory
+    self.issuer = issuer
+    self.issuerCountry = issuerCountry
+    self.productId = productId
+    self.productType = productType
+    self.billingAddress = billingAddress
+    self.phone = phone
+    self.name = name
```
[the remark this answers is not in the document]
ehab-al-cko: It's in another test framework. Cannot be accessed from another test framework.

Source/UI/NewUI/CommonUI/View/Core/ButtonView.swift
ehab-al-cko: Nope. only one around the view.

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
ehab-al-cko: If you see any pros for this solution, please suggest code.

Source/Core/Constants/Constants+LocalizationKeys.swift
```diff
@@ -35,6 +35,10 @@ extension Constants {
               static let hint = "SecurityCodeHint".localized
               static let error = "SecurityCodeErrorMessage".localized
           }
+
+          struct PayButton {
+            static let title = "Pay".localized
```
alex-ioja-yang-cko, line 40:
```suggestion
            static func title(value: String) { 
                String(format: "Pay".localized, value)
            }
```

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
```diff
@@ -175,9 +175,9 @@ public class CardValidator: CardValidating {
   }
 
   public func validate(_ card: Card) -> ValidationResult<ValidationError.Card> {
+    guard let number = card.number else { return .failure(.cardNumber(.invalidCharacters))}
```
alex-ioja-yang-cko, line 178: If we return an error, I think it may be appropriate to declare a correct error for the failure reported, as invalidCharacters is not accurate.

I think we can add another case, `missingNumber` ?

Source/UI/NewUI/BillingForm/Default Implementation/BillingForm/Header/Cancel/DefaultCancelButtonFormStyle.swift
```diff
@@ -1,6 +1,7 @@
 import UIKit
 
 public struct DefaultCancelButtonFormStyle: ElementButtonStyle {
+    public var textAlignment: NSTextAlignment = .natural
```
alex-ioja-yang-cko, line 4: I've tried to do a search in the PR for `textAlignment` and its been used 82 times. I can't see from the task or the code changes why that is needed, especially to the extent its been done. Do you think something that repetitive may be worth its own task, or is there any way we can understand why it was needed in the first place?

Source/UI/NewUI/CommonUI/View/Core/TextFieldView.swift
```diff
@@ -36,6 +36,7 @@ class TextFieldView: UIView {
         textField.placeholder = style.placeHolder
         textField.textColor = style.textColor
         textField.tintColor = style.tintColor
+        textField.textAlignment = style.textAlignment
```
alex-ioja-yang-cko, line 39: Documentation shows UILabel has a different default from UITextField in relation to the alignment, so it may be better to only apply when provided in all cases. 

Having said this, could we pull out the textAlignment from PR and have a separate discussion on why / how it benefits?

iOS Example Frame SPM/iOS Example Frame SPM.xcodeproj/project.pbxproj
```diff
@@ -320,7 +320,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if which swiftlint > /dev/null; then\n  swiftlint --fix --config ../.swiftlint.yml\nelse\n  echo \"error: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi\n";
+			shellScript = "# Adds support for Apple Silicon brew directory\nexport PATH=\"$PATH:/opt/homebrew/bin\"\n\nif which swiftlint > /dev/null; then\n  swiftlint --fix --config ../.swiftlint.yml\nelse\n  echo \"error: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi\n";
```
alex-ioja-yang-cko, line 323: Was this intended? Shouldn't we try to have a separate focussed chat if we change unrelated things like SwiftLint build script?

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
```diff
@@ -9,7 +9,7 @@
 import Checkout
 
 protocol CardNumberViewModelDelegate: AnyObject {
-  func update(cardNumber: String, scheme: Card.Scheme)
+  func update(cardNumber: String?, scheme: Card.Scheme)
```
alex-ioja-yang-cko, line 12: Could we pass an empty String for cardNumber if we don't have one? Why does it need to be nil?

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
```diff
@@ -35,13 +35,15 @@ extension CardNumberViewModel: CardNumberViewModelProtocol {
...
     case .failure:
+        delegate?.update(cardNumber: nil, scheme: .unknown)
         return nil
     // An unknown scheme can also be generated when an eager validation was matched
     //    but the number is not complete.
     // In this case we only update delegate if we have a final result to avoiding
     //    overriding the eager validation with a less informative update
     case .success((let isComplete, let scheme)):
       guard isComplete else {
+        delegate?.update(cardNumber: nil, scheme: .unknown)
```
alex-ioja-yang-cko, line 46: If the number is not complete, does that mean we shouldn't update our record of it? I am unsure why all these return paths are returning `nil` when if looks like a partial number would also work? Would prefer to avoid handling optionality if we don't have to

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
alex-ioja-yang-cko: I think code reviews are a good place for having discussions around implementation and details. I'm trying to understand why you chose to go for a completion handler in the factory versus injecting a delegate.

Personally I find delegates safer for retain cycles as it gives us the control over weakening the callback reference and helps provide a safer solution for consumer of SDK.

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -73,18 +73,27 @@ extension CardNumberView: TextFieldViewDelegate {
...
   }
 
   func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+
+    // some characters are deleted
+    // notify payment view controller to disable the pay button
+    if string.count == 0 && range.length > 0 {
+      _ = viewModel.validate(cardNumber: string)
+      return true
+    }
```
alex-ioja-yang-cko, line 95: Your current approach relies on the viewModel receiving incomplete inputs (like an empty string here) and prevented from making informed decisions over changes. 

This is at odds with the other code you wrote that relies on the viewmodel receiving input for validation so the contrasting approach for the code within this method is really unhealthy for comprehension. Could you please remove this code and find a way for the viewModel to control the button state whilst integrating with existing logic and keeping it simpler?

Source/UI/NewUI/PaymentForm/View/ExpiryDateView.swift
```diff
@@ -146,6 +146,9 @@ public final class ExpiryDateView: UIView {
       Constants.LocalizationKeys.PaymentForm.ExpiryDate.Error.past :
       Constants.LocalizationKeys.PaymentForm.ExpiryDate.Error.invalid
     style?.error?.isHidden = isHidden
+    if !isHidden {
+      delegate?.update(expiryDate: nil)
+    }
```
alex-ioja-yang-cko, line 151: I tried finding but I'm unable, could you please show me where the `ExpiryDateView` is calling delegate with non nil  values? This code shows some logic problems in my understanding so far 🤔

Source/UI/NewUI/PaymentForm/View/SecurityCodeView.swift
```diff
@@ -51,6 +51,9 @@ public final class SecurityCodeView: UIView {
   }
 
   private func updateErrorViewStyle(isHidden: Bool, textfieldText: String?) {
+    if !isHidden {
+      delegate?.update(securityCode: nil)
+    }
```
alex-ioja-yang-cko, line 56: Similar to previous concerns, this delegate is only ever called with `nil`??

alex-ioja-yang-cko (commented):
I think we have a few iterations of reviews to go to be honest.

Source/Resources/en.lproj/Localizable.strings
```diff
@@ -24,6 +24,7 @@
 "PaymentHeaderSubtitle" = "Visa, Mastercard and American Express accepted.";
 "CardNumber" = "Card number";
 "CardNumberErrorMessage" = "Please enter a valid card number";
+"Pay" = "Pay";
```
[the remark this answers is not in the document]
ehab-al-cko: There's no amount at all for default screen.

Source/Core/Constants/Constants+LocalizationKeys.swift
ehab-al-cko: same

Source/UI/NewUI/BillingForm/Default Implementation/BillingForm/Header/Cancel/DefaultCancelButtonFormStyle.swift
ehab-al-cko: _why it was needed in the first place?_ it's text alignment :)

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
ehab-al-cko: if we need to update payment screen with invalid value for cardNumber to update pay button

Checkout/Checkout/Source/Model/TokenDetails.swift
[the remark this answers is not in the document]
deepesh-vasthimal-cko: When we call `checkoutAPIService.createToken` we pass back the `TokenDetails` to merchant so they can use TokenDetails as it contains all the related information about that token ( read only ). Also this is the response from the tokens [end point](https://api-reference.checkout.com/?utm_source=Google&utm_medium=paid_search&utm_campaign=brand_UK_en_desktop_ex&utm_adcontent=brand&gclid=CjwKCAjw6MKXBhA5EiwANWLODBckOEkmtZKmUU6qdzVqjKe7Li0DyS3LMujdxDBs5w_GjZ6x1m7kPRoCot4QAvD_BwE#operation/requestAToken). So I agree with @alex-ioja-yang-cko that our merchants will never create their own response and this does not need to be a public init.

Checkout/Checkout/Source/Validation/Validators/CardValidator.swift
deepesh-vasthimal-cko: Instead of adding new error. Please check why is the card.number even a optional value at first place. Can the number inside the Card struct just be a string.

Source/Resources/en.lproj/Localizable.strings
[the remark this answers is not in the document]
deepesh-vasthimal-cko: Note : This amount will be passed by merchants if they want the amount to be visible after the Pay Text as per the design. 

@alex-ioja-yang-cko the %@ will be replaced by any string passed by merchant. 

However the other problem I see here is there is no way for us to validate this value. Merchant can create a token, call the /payments end point ( for 500£ ) and on the screen show just 5£. In this case it will be merchants responsibility they pass right values to payments end point and what they show on screen. @patrick-hoban-cko are we fine with that behaviour ?

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
deepesh-vasthimal-cko: My high level understanding was, since we are using callbacks everywhere in the code we are preferring callbacks and not using delegate. More of the design choice between callback or delegate. @ehab-al-cko Please correct me if I am wrong on this.

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
deepesh-vasthimal-cko: Also if you are returning true I can see in the code we are doing a bit of validation below to update error labels etc. Just do some test so you are not breaking anything. Its same for ExpiryDateView, SecurityCodeView & CardNumberView

Tests/Mocks/PaymentViewControllerMockDelegate.swift
```diff
@@ -0,0 +1,46 @@
...
+import Checkout
+@testable import Frames
+
+class PaymentViewControllerMockDelegate: PaymentViewControllerDelegate {
+
+  var addBillingButtonIsPressedWithSender: [UINavigationController?] = []
+  var editBillingButtonIsPressedWithSender: [UINavigationController?] = []
+  var expiryDateIsUpdatedWithValue: [ExpiryDate?] = []
+  var securityCodeIsUpdatedWithValue: [String?] = []
+  var cardNumberIsUpdatedWithValue: [String?] = []
```
deepesh-vasthimal-cko, line 19: Question : Why is this a array containing card numbers. I can see `delegate?.cardNumberIsUpdated(value: cardNumber)` is called which passes the cardNumber and its appending in array.

Source/Resources/en.lproj/Localizable.strings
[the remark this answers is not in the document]
alex-ioja-yang-cko: Should we then remove the number from the default? The user of SDK can pass the title as they wish (with value, currency, however they want it) and our default will remove this false expectation.

We probably should add the number from the demo app (CustomStyle), not from SDK then

Source/UI/NewUI/BillingForm/Default Implementation/BillingForm/Header/Cancel/DefaultCancelButtonFormStyle.swift
alex-ioja-yang-cko: Well yes, I get that, but moving all of `UITextInput` properties into our own version of UI declaration doesn't really help. I think we went overboard with properties already (see height, width, isHidden, isEnabled, etc.

This is why before we add anything new I think we should really evaluate whether its needed and what for, can we do without, can we strip something useless and replace... We can't just keep migrating UIControl, UITextInput and UIView through our configurations...

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
alex-ioja-yang-cko: Ok, what is *invalid value*? 

In my understanding invalid value is an input that doesn't quite make it as a card number. 
Nil is not an input that doesn't make it as a card number, nil is the absence of such an input.

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
alex-ioja-yang-cko: I don't think that assertion should apply here as this is the only place the user gets to catch the output of the journey. All the others are internal, between the VC and VM.

Consumer will say

> I want to start a tokenisation request with this input and on output I want this to happen

Anything else is hidden from consumer (dev team). And completion handlers whilst they read easier at declaration site are more prone for implementation errors. And whilst we are not responsible for our consumers technical expertise, making our solution work for more people at no cost (delegate pattern is common in iOS) might be a nice small detail of our product.

Source/Resources/en.lproj/Localizable.strings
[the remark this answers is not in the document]
patrick-hoban-cko: Yes, we should add the number from the demo app, not from the SDK.
 
That would imitate how a merchant actually uses this - they pass the value into the string in order to show the correct number in the Pay button. We don't define that for them of course, as we don't know the value of the basket.

Checkout/Checkout/Source/Model/Card.swift
harry-brown-cko, line 18: A card must have a `number` and an `expiryDate`. Please undo this change as it makes the API for headless SDKs confusing

Checkout/Checkout/Source/Model/TokenDetails.swift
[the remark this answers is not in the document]
harry-brown-cko: Agree with @alex-ioja-yang-cko and @deepesh-vasthimal-cko here

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
harry-brown-cko: Apple and Swift are now moving away from delegates towards callbacks and eventually `async`/`await`s. Delegate methods are tried and true but if we want to follow examples from newer Apple APIs, such as those in SwiftUI, then this should be a callback.

alex-ioja-yang-cko: > Apple and Swift are now moving away from delegates towards callbacks

I think that is a pretty bold statement. SwiftUI is a different programming paradigm, and async/await tries to serialise async logic for reader.

***
> Delegate methods are tried and true but if we want to follow examples from newer Apple APIs, such as those in SwiftUI, then this should be a callback.

Given SwiftUI is declarative UI, the whole paradigm is different so that does not mean delegates are discouraged or becoming obsolete. 
Delegates, completion handlers, bindings are all ways for achieving the same task which is to communicate from child to parent. SwiftUI being declarative works naturally with bindings. That is not relevant therefore to our assumptions on delegates vs completion handlers, unless if I missed or misunderstood something?

Source/UI/NewUI/BillingForm/Default Implementation/BillingForm/Header/Cancel/DefaultCancelButtonFormStyle.swift
ehab-al-cko: We can discuss required changes for these Style properties with team

iOS Example Frame SPM/iOS Example Frame SPM.xcodeproj/project.pbxproj
```diff
@@ -523,7 +523,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_STYLE = Automatic;
-				DEVELOPMENT_TEAM = 635KC97JGT;
+				DEVELOPMENT_TEAM = E32XBQK4Q5;
```
deepesh-vasthimal-cko, line 526: Was this change intentional ?

deepesh-vasthimal-cko (changes requested):
Just remove the public init for TokenDetails, otherwise looks good.

Source/Extensions/UIStackViewExtension.swift
```diff
@@ -0,0 +1,15 @@
+//
+//  File.swift
```
harry-brown-cko, line 2:
```suggestion
//  UIStackViewExtension.swift
```

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
```diff
@@ -9,30 +9,28 @@ public struct PaymentFormFactory {
...
+    let viewModel = DefaultPaymentViewModel(checkoutAPIService: checkoutAPIService,
+                                            cardValidator: cardValidator,
                                             logger: logger,
                                             billingFormData: configuration.billingFormData,
                                             paymentFormStyle: style.paymentFormStyle,
                                             billingFormStyle: style.billingFormStyle,
                                             supportedSchemes: configuration.supportedSchemes)
 
     let viewController = PaymentViewController(viewModel: viewModel)
+    viewController.cardTokenRequested = cardTokenResultClousure
```
deepesh-vasthimal-cko, line 29: Not part of your PR but Clousure is misspelled-> Closure

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
deepesh-vasthimal-cko, line 13: `cardTokenResultClousure` is the naming convention right ? calling a closure ending with Closure keyword. Can you please guide me with some examples I can refer.

ehab-al-cko: Following apple naming convention, it's changed to `completionHandler`

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -73,18 +77,27 @@ extension CardNumberView: TextFieldViewDelegate {
...
 
   func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+
+    // some characters are deleted
+    // notify payment view controller to disable the pay button
+    if string.count == 0 && range.length > 0 {
+      delegate?.textFieldDidStartDeleteText()
+      return true
+    }
+
```
alex-ioja-yang-cko, line 100: shouldn't this just result in a standard update to the input which then triggers a validation that will obviously disable the button? Why do we need to say "deleted"? 

Fyi, this will also break if you happen to have some card scheme that has consecutive valid number of inputs, or for VISA as its been implemented to handle 19 characters. 17 is not valid, but deleting another to 16 may have valid number. 

This logic, assuming a char delete will result in invalid input, is inherently flawed.

_Also I will only leave 1 comment here, but same applies to every input field in this PR._

deepesh-vasthimal-cko (approved):

Source/UI/NewUI/PaymentForm/ViewController/PaymentViewController.swift
```diff
@@ -4,15 +4,17 @@ import Checkout
...
+  func securityCodeIsUpdated(result: Result<String, SecurityCodeError>)
+  func payButtonIsPressed()
 }
 
 final class PaymentViewController: UIViewController {
 
   // MARK: - Variables
 
   weak var delegate: PaymentViewControllerDelegate?
+  var cardTokenRequested: ((Result<TokenDetails, TokenisationError.TokenRequest>) -> Void)?
```
alex-ioja-yang-cko, line 17: Should outcomes be coming from a ViewModel, not from a ViewController?

iOS Example Frame/iOS Example Frame/MainViewController.swift
```diff
@@ -72,10 +74,21 @@ final class MainViewController: UIViewController, CardViewControllerDelegate, Th
...
         let paymentStyle = PaymentStyle(paymentFormStyle: Style.Custom1.paymentForm,
                                         billingFormStyle: Style.Custom1.billingForm)
-        let paymentFormViewController = PaymentFormFactory.buildViewController(configuration: paymentConfiguration, style: paymentStyle)
+        let paymentFormViewController = PaymentFormFactory.buildViewController(configuration: paymentConfiguration, style: paymentStyle) { [weak self] result in
+          self?.handleTokenResponse(with: result)
+        }
         navigationController?.pushViewController(paymentFormViewController, animated: true)
     }
 
+  func handleTokenResponse(with result: Result<TokenDetails, TokenisationError.TokenRequest>) {
```
alex-ioja-yang-cko, line 83:
```suggestion
  private func handleTokenResponse(with result: Result<TokenDetails, TokenisationError.TokenRequest>) {
```

alex-ioja-yang-cko (dismissed):
The behaviour of the input validation is flawed

Source/UI/NewUI/BillingForm/Default Implementation/BillingForm/Header/Cancel/DefaultCancelButtonFormStyle.swift
alex-ioja-yang-cko: Absolutely. As part of the PR though I would still like to understand what problem you've tried to solve that required you to add this extra property, that imo we should scope separately?

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
ehab-al-cko: what about validating firstly then if not valid then fire the delegate ?

ehab-al-cko: @alex-ioja-yang-cko what about this suggestion ?

```suggestion
    // some characters are deleted
    // notify payment view controller to disable the pay button
    if string.count == 0 && range.length > 0 {
       if let text = textField.text, viewModel.validate(cardNumber: text) == nil {
          delegate?.textFieldDidStartDeleteText()
       }
       return true
   }
```

Source/UI/NewUI/CommonUI/View/Core/ButtonView.swift
```diff
@@ -50,19 +55,16 @@ class ButtonView: UIView {
...
 
     private func updateButtonStyle(with style: ElementButtonStyle) {
+        backgroundColor = isEnabled ? style.backgroundColor : style.disabledTintColor
         button.tintColor = .clear
-        button.clipsToBounds = true
-        button.layer.borderColor = style.normalBorderColor.cgColor
-        button.layer.borderWidth = style.borderWidth
-        button.setTitleColor(style.disabledTextColor, for: .disabled)
-        button.setTitleColor(style.textColor, for: .normal)
+        button.heightAnchor.constraint(equalToConstant: style.height).isActive = true
```
alex-ioja-yang-cko, line 60: Can we do without forcing the height of the button?

Alternatively, we could create a constraint for the height to be "At least 44 pixels height" (in line with HIG) but not actually limit it from growing taller if it so needs to

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberView.swift
```diff
@@ -9,7 +9,7 @@
 import UIKit
 
 final class CardNumberView: UIView {
-  private let viewModel: CardNumberViewModelProtocol
+  private var viewModel: CardNumberViewModelProtocol
```
alex-ioja-yang-cko, line 12:
```suggestion
  private let viewModel: CardNumberViewModelProtocol
```

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
```diff
@@ -32,6 +40,8 @@ class CardNumberViewModel {
 }
 
 extension CardNumberViewModel: CardNumberViewModelProtocol {
+
+  // TODO: separate the validation logic from getting the icon
```
alex-ioja-yang-cko, line 44: Do we need this TODO? Why do we need to separate that logic, if the validation ties this logic already?

I validate a card so now I know if it is of any known scheme and whether its a complete valid entry for it. Its all 1 validation, not sure how you'd like to separate?

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
```diff
@@ -41,14 +51,16 @@ extension CardNumberViewModel: CardNumberViewModelProtocol {
...
+        delegate?.update(result: .success((cardNumber, scheme)))
         return Constants.Bundle.SchemeIcon(scheme: scheme)
       case .success,
            .failure:
+        delegate?.update(result: .failure(.isNotComplete))
         return nil
     }
   }
 
+  // TODO: separate the validation logic from getting the icon
```
alex-ioja-yang-cko, line 63:
```suggestion
```

Source/UI/NewUI/PaymentForm/ViewController/PaymentViewController.swift
alex-ioja-yang-cko, line 17: This was moved to ViewModel so can probably be deleted from here.

Source/UI/NewUI/PaymentForm/ViewController/PaymentViewController.swift
```diff
@@ -56,10 +58,9 @@ final class PaymentViewController: UIViewController {
   }()
 
   private lazy var cardNumberView: CardNumberView = {
-    let cardNumberViewModel = CardNumberViewModel(cardValidator: viewModel.cardValidator, supportedSchemes: viewModel.supportedSchemes)
-    cardNumberViewModel.delegate = self
+let cardNumberViewModel = CardNumberViewModel(cardValidator: viewModel.cardValidator, supportedSchemes: viewModel.supportedSchemes)
+    cardNumberViewModel.delegate = viewModel as? CardNumberViewModelDelegate
```
alex-ioja-yang-cko, line 62:
```suggestion
    let cardNumberViewModel = CardNumberViewModel(cardValidator: viewModel.cardValidator, supportedSchemes: viewModel.supportedSchemes)
    cardNumberViewModel.delegate = viewModel as? CardNumberViewModelDelegate
```

Source/UI/NewUI/PaymentForm/ViewController/PaymentViewController.swift
```diff
@@ -70,7 +71,23 @@ final class PaymentViewController: UIViewController {
...
+    view.isEnabled = false
+    return view
+  }()
+
+  private lazy var activityIndicator: UIActivityIndicatorView = {
+    let view =  UIActivityIndicatorView(frame: CGRect(x: 0, y: 0, width: 20, height: 20))
+    let barButton = UIBarButtonItem(customView: view)
+    navigationItem.setRightBarButton(barButton, animated: true)
+    return view
+  }()
```
alex-ioja-yang-cko, line 86: This is useful for some dev feedback, but realistically a small loading indicator in navigation bar is not serving any other purpose than dev checks. And we may end up forgetting it, so might be useful to delete and make sure @patrick-hoban-cko has some task for some type of loading state on the screen.

Source/UI/NewUI/PaymentForm/ViewController/PaymentViewController.swift
```diff
@@ -70,7 +71,23 @@ final class PaymentViewController: UIViewController {
...
+
+  private lazy var headerBackgroundView: UIView = {
+    UIView().disabledAutoresizingIntoConstraints()
+  }()
```
alex-ioja-yang-cko, line 90:
```suggestion
  private let headerBackgroundView: UIView = UIView().disabledAutoresizingIntoConstraints()
```

I have previously mentioned how lazy loading properties that get loaded on init anyway is not "a pattern" and this is genuine functionality that we are misusing for no good reason other than "most properties are assigned by closure in this file".

Source/UI/NewUI/PaymentForm/ViewController/PaymentViewController.swift
```diff
@@ -319,25 +378,21 @@ extension PaymentViewController: BillingFormSummaryViewDelegate {
 }
 
 extension PaymentViewController: ExpiryDateViewDelegate {
-  func update(expiryDate: ExpiryDate) {
-    delegate?.expiryDateIsUpdated(value: expiryDate)
+  func update(result: Result<ExpiryDate, ExpiryDateError>) {
+    delegate?.expiryDateIsUpdated(result: result)
```
alex-ioja-yang-cko, line 382: Sorry @ehab-al-cko , I'll need to pull the plug again as I just came across this serious omission which we absolutely need to rectify.

In current design, our ViewController owns a ViewModel, but also owns a delegate, which is the ViewModel.

I will not block this PR for it, but please open a task and link it here, we'll need to cover that ASAP. To me right now it just looks like the **delegate should not exist**, given that delegate is the child view model already. It just doesn't make sense how we got here?

Source/UI/NewUI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
```diff
@@ -1,32 +1,83 @@
...
+  func getCard() -> Card? {
+    guard let number = number, let expiryDate = expiryDate else { return nil }
+    return Card(number: number,
+                expiryDate: expiryDate,
+                name: name,
+                cvv: cvv,
+                billingAddress: billingAddress,
+                phone: phone)
+  }
+}
```
alex-ioja-yang-cko, line 30: Not sure if we want to have objects private on a file? I have previously mentioned creating a helper internal object that keeps your state, like `CardBuilder` or similar.

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
```diff
@@ -25,6 +25,29 @@ private enum Constants {
...
+struct PayButtonFormStyleCustom1: ElementButtonStyle {
+  var image: UIImage?
+  var textAlignment: NSTextAlignment = .center
+  var text: String = "Pay 100$"
+  var font: UIFont = UIFont.systemFont(ofSize: 15)
+  var disabledTextColor: UIColor = Constants.secondaryFontColor
+  var disabledTintColor: UIColor = Constants.secondaryFontColor
+  var activeTintColor: UIColor = Constants.mainFontColor
+  var backgroundColor: UIColor = Constants.mainFontColor
+  var textColor: UIColor = .white
```
alex-ioja-yang-cko, line 37: Some background colour here is failing as the text is invisible when button is disabled.

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
```diff
@@ -168,6 +197,7 @@ private struct StyleOrganiser {
     }
 
     struct TextFieldStyle: ElementTextFieldStyle {
+        var textAlignment: NSTextAlignment = .natural
```
alex-ioja-yang-cko, line 200:
```swift
 open var textAlignment: NSTextAlignment // default is NSLeftTextAlignment
```

```suggestion
        var textAlignment: NSTextAlignment = .left
```

iOS Example Frame/iOS Example Frame/CustomStyle2.swift
```diff
@@ -24,9 +24,33 @@ private enum Constants {
...
+struct PayButtonFormStyleCustom2: ElementButtonStyle {
+  var image: UIImage?
+  var text: String = "Pay 100$"
+  var textAlignment: NSTextAlignment = .center
+  var font: UIFont = UIFont.systemFont(ofSize: 15)
+  var disabledTextColor: UIColor = Constants.grayBackgroundColor
+  var disabledTintColor: UIColor =  Constants.grayBackgroundColor
+  var activeTintColor: UIColor =  Constants.greenBackgroundColor
+  var backgroundColor: UIColor = Constants.greenBackgroundColor
+  var textColor: UIColor = .white
```
alex-ioja-yang-cko, line 36: This button is invisible, is that intended?

Source/UI/NewUI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
```diff
@@ -1,32 +1,83 @@
...
   }
+  var isLoading: Bool = false {
+    didSet {
+      updateLoading?()
+    }
+  }
+
+  private var cardDetails: CardDetails = CardDetails() {
+    didSet {
+      shouldEnablePayButton?(cardDetails.expiryDate != nil && cardDetails.number != nil)
```
alex-ioja-yang-cko, line 69: Whilst @patrick-hoban-cko did say CVV may not be required, I think that was down to the developer to decide if they want it required or not? Shouldn't we use the `isMandatory` on the SecurityNumberStyle to decide if we include it or not?

Also, is this completion handler called every time we update a property within the object? I used to think this setter is called only when `setting the object`, which would probably be done at initialisation time?

Source/UI/NewUI/CommonUI/View/Core/ButtonView.swift
ehab-al-cko: button height  is controlled by the merchant `style.height`. after current release, merchant will be able to change any sizes

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
ehab-al-cko: `func validate(cardNumber rawText: String) -> Constants.Bundle.SchemeIcon?` 
function is asking for validation and return image. 
`validate` is expected return boolean then with the boolean get image or locatization or or anything.

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
ehab-al-cko: same. i will add ticket for refactoring. they are tightly coupled and should have apply SRP/

Source/UI/NewUI/PaymentForm/ViewController/PaymentViewController.swift
ehab-al-cko: We are waiting for Alexander the designer about this. for now, it is temp solution.

Source/UI/NewUI/PaymentForm/ViewController/PaymentViewController.swift
ehab-al-cko: Following the same pattern for readability and consistency.

Source/UI/NewUI/PaymentForm/ViewController/PaymentViewController.swift
ehab-al-cko: It was before, if you want to refactoring then please add jira ticket.

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
alex-ioja-yang-cko: @ehab-al-cko , I strongly disagree with your understanding of what validation does. It obviously is not a question of:
"Is this valid", but more of "What type is it". Please try to understand the wider picture clearer. 

I will push back against such a refactor here as it just does not make sense. API methods should not only be expected to return booleans, they are perfectly fine returning other types that can then be used to infer your needs.

As a result to move forward, please remove the TODO and open a Jira task that we can scope and discuss on.

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
ehab-al-cko: We are support RTL.
_NSTextAlignment.natural
Text uses the default alignment for the current localization of the app._
_The default alignment for left-to-right scripts is NSTextAlignment.left, and the default alignment for right-to-left scripts is NSTextAlignment.right._
https://developer.apple.com/documentation/uikit/nstextalignment/natural

Source/UI/NewUI/PaymentForm/View/CardNumber/CardNumberViewModel.swift
alex-ioja-yang-cko: I can understand why this looks tightly coupled, the validation method also reformats the input 👍 . I think I disagree on how this will be improved, but would like a task for it to be created and this TODO removed. TODOs won't get anywhere and just trigger warnings once merged...

Source/UI/NewUI/PaymentForm/ViewController/PaymentViewController.swift
alex-ioja-yang-cko: Sure, its a good temp solution for you to test behaviour, but not a good solution to merge into a main develop branch. That will absolutely not stay like this and you only risk adding code that will be forgotten on refactor, please remove.

Source/UI/NewUI/PaymentForm/ViewController/PaymentViewController.swift
alex-ioja-yang-cko: I think you have created this pattern and then built on it, from my side I've joined late and only just realised. If you are unhappy to review this behaviour and create a task for fixing it, I can do so from my side, sure.

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
alex-ioja-yang-cko: 👍  What I'm unsure of then is why the default from Apple is `left`. The line I added for documentation is from UIKit.TextField.textAlignment 🤔 🤷‍♂️

Source/UI/NewUI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
ehab-al-cko: please suggest code

alex-ioja-yang-cko: Do you know a way of suggesting GitHub changes where we move objects in new files? Otherwise all I could suggest is:

```suggestion

```

But I'm not sure its very helpful

alex-ioja-yang-cko: FYI, I've [found original suggestion](https://github.com/checkout/frames-ios/pull/241#discussion_r939879279)

iOS Example Frame/iOS Example Frame/CustomStyle1.swift
ehab-al-cko: updated to `var disabledTextColor: UIColor = .white`

iOS Example Frame/iOS Example Frame/CustomStyle2.swift
ehab-al-cko: Same

Source/UI/NewUI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
ehab-al-cko: @alex-ioja-yang-cko, i think, this can be after isMandotry [ticket](https://checkout.atlassian.net/browse/PIMOB-1381?atlOrigin=eyJpIjoiZTJhMDJkY2RiNDlkNDU4OGE1MTY1NjlkN2NlM2QyMjAiLCJwIjoiaiJ9) 
Adding a [comment](https://checkout.atlassian.net/browse/PIMOB-1381?focusedCommentId=182223) to the ticket. it will be a small update for this logic.

Source/UI/NewUI/PaymentForm/ViewModel/CardCreationModel.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  File.swift
```
alex-ioja-yang-cko, line 2:
```suggestion
//  CardCreationModel.swift
```

Source/UI/NewUI/PaymentForm/ViewModel/CardCreationModel.swift
```diff
@@ -0,0 +1,36 @@
...
+  var phone: Phone?
+
+  init(number: String? = nil, expiryDate: ExpiryDate? = nil, name: String? = nil, cvv: String? = nil, billingAddress: Address? = nil, phone: Phone? = nil) {
+    self.number = number
+    self.expiryDate = expiryDate
+    self.name = name
+    self.cvv = cvv
+    self.billingAddress = billingAddress
+    self.phone = phone
+  }
```
alex-ioja-yang-cko, line 25: I think this initialiser is not needed as we plan to keep it internal anyway.

```suggestion
```

alex-ioja-yang-cko (approved):
There seems to be a conflict with base branch and some failing on CI. Minor comments left, but removing obsolete initialisers may be a necessary improvement

PR #242, opened 2022-08-08
PIMOB-1366: Match Maestro card scheme with different options of CVV lengths

alex-ioja-yang-cko:
## Proposed changes

[Jira task](https://checkout.atlassian.net/browse/PIMOB-1366)

## Types of changes

* [X] New feature (non-breaking change which adds functionality

## Checklist

The consequences of the fix were more complex than it might appear on first sight:
- Card.Scheme is an enum that didn't have associated values to any type. Creating one for maestro led to losing some inherited behaviours that had to be updated
- created `CardScheme` as a frames client interface for `Card.Scheme`. This enables end user to continue adding cases via nicer non associated value syntax for all cases. The parameter is also irrelevant to frames its only for internal handling so wouldn't make sense to expose it if we can avoid.
ie:
```Swift
// CardScheme
[.visa, .maestro, .mastercard]

// Card.Scheme
[visa, .maestro(), .mastercard]
```


Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
```diff
@@ -89,12 +89,21 @@ final class CardNumberValidator: CardNumberValidating {
...
     }
+    return addSchemePropertyIfNeeded(scheme: matchedScheme, cardNumber: cardNumber)
   }
 
   private func validateDigitsOnly(in string: String) -> Bool {
     return CharacterSet.decimalDigits.isSuperset(of: CharacterSet(charactersIn: string))
   }
+    
+  private func addSchemePropertyIfNeeded(scheme: Card.Scheme?, cardNumber: String) -> Card.Scheme? {
+    if scheme == .maestro(length: 0) {
```
harry-brown-cko, line 103: this is now dependent on the Maestro scheme in `Card.Scheme.allCases` being exactly `.maestro(length: 0)`. We can make this a bit simpler
```suggestion
    if scheme == .maestro {
```

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
```diff
@@ -89,12 +89,21 @@ final class CardNumberValidator: CardNumberValidating {
...
+        return .maestro(length: cardNumber.filter { Int("\($0)") != nil }.count)
```
harry-brown-cko, line 104: we've already checked for whitespace and non-digit characters. We can remove this extraneous check
```suggestion
        return .maestro(length: cardNumber.count)
```

Source/Core/Models/CardScheme.swift
```diff
@@ -0,0 +1,41 @@
+//
+//  CardScheme.swift
+//  
+//
+//  Created by Alex Ioja-Yang on 05/08/2022.
+//
+
+public enum CardScheme {
```
harry-brown-cko, line 8: no need. please undo

alex-ioja-yang-cko: This is needed, have documented in PR description. Will copy here as well for ease of discovery:

> created CardScheme as a frames client interface for Card.Scheme. This enables end user to continue adding cases via nicer non associated value syntax for all cases. The parameter is also irrelevant to frames its only for internal handling so wouldn't make sense to expose it if we can avoid.

```Swift
// CardScheme
[.visa, .maestro, .mastercard]

// Card.Scheme
[visa, .maestro(), .mastercard]
```

Short version to why its needed:
**Consumer should not be concerned with variations of maestro, so the associated value should not be exposed**

iOS Example Frame/iOS Example Frame/MainViewController.swift
```diff
@@ -119,7 +119,7 @@ final class MainViewController: UIViewController, CardViewControllerDelegate, Th
         cardViewController = createCardViewController(address: address,
                                                       phone: phone,
                                                       checkoutAPIService: checkoutAPIService)
-        cardViewController?.availableSchemes = [.visa, .mastercard, .maestro]
+        cardViewController?.availableSchemes = [.visa, .mastercard, .maestro()]
```
deepesh-vasthimal-cko, line 122: Why is the being passed as maestro() ?

deepesh-vasthimal-cko (approved):

iOS Example Frame/iOS Example Frame/MainViewController.swift
alex-ioja-yang-cko: This is the old screen that will get removed. New one uses the `CardScheme` interface I've had to add, whereas here it just uses Card.Scheme and is therefore dependant on passing associated value. Its using a default one (0), but syntax is still impacted

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
alex-ioja-yang-cko: That is not possible because the associated type cannot just be ignored and equalise without (see screenshot for error).
Screenshot 2022-08-10 at 12 06 43
 
 ***
 
 Even if it did work, we would only want to `addSchemePropertyIfNeeded` and if a length existed we wouldn't want to override it.

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
alex-ioja-yang-cko: They are not actually being validated to the same standard, just removing white spaces. I have moved this validation to the callers of this method, so as suggested I did remove this filter here (you were correct), but had to extend that filter.

Working on issuing, I can see how a consumer may want to display input numbers in many formats:
```
4242 4242 4242 4242
4242 - 4242 - 4242 - 4242
4242 : 4242 : 4242 : 4242
4242 . 4242 . 4242 . 4242
```

Only the first one was being stripped correctly, others would result in false results.

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
```diff
@@ -89,12 +90,21 @@ final class CardNumberValidator: CardNumberValidating {
...
 
   private func validateDigitsOnly(in string: String) -> Bool {
     return CharacterSet.decimalDigits.isSuperset(of: CharacterSet(charactersIn: string))
   }
+    
+  private func addSchemePropertyIfNeeded(scheme: Card.Scheme?, cardNumber: String) -> Card.Scheme? {
+    if scheme == .maestro(length: 0) {
+        return .maestro(length: cardNumber.count)
+    }
+    return scheme
```
harry-brown-cko, line 107: suggestion to link to other comment
```suggestion
    switch scheme {
    case .maestro:
        return .maestro(length: cardNumber.count)
    default:
        return scheme
    }
```

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
harry-brown-cko: sorry this is only possible in a switch. if you're wondering what that might look like https://github.com/checkout/frames-ios/pull/242/files#r942569571

harry-brown-cko: sorry got the syntax a bit wrong here
```suggestion
    if case .maestro = scheme {
```

Source/Core/Models/CardScheme.swift
harry-brown-cko: Ok makes sense. Is this now preferred to use across the SDK instead of `Card.Scheme`?

Checkout/Checkout/Source/Model/Card+Scheme.swift
```diff
@@ -29,13 +31,13 @@ extension Card {
...
-    case maestro
+    case maestro(length: Int = 0)
     case americanExpress
     case discover
     case dinersClub
     case jcb
 
-    public init?(rawValue: String) {
+    public init?(rawValue: String, length: Int? = nil) {
       switch rawValue.lowercased() {
```
ehab-al-cko, line 41: `Scheme` rawValue is no longer String. so why need for init with string rawValue?

Checkout/Checkout/Source/Model/Card+Scheme.swift
```diff
@@ -19,7 +19,9 @@ extension Card {
   ///   - Maestro
   ///   - Mastercard
   ///   - Visa
-  public enum Scheme: String, CaseIterable, CustomStringConvertible {
+  public enum Scheme: CaseIterable, Equatable, Hashable {
```
ehab-al-cko, line 22: this file is full of 100s of switches cases, it will need to be refactored to separate calculations from model

Checkout/Checkout/Source/Model/Card+Scheme.swift
```diff
@@ -120,8 +122,11 @@ extension Card {
       switch self {
       case .unknown:
         return Constants.validCVVLengthsUnknownScheme
-      case .maestro:
-          return [0, 3]
+      case .maestro(let length):
+        if length == 16 {
+          return [3]
+        }
+        return [0, 3]
```
ehab-al-cko, line 129:
```suggestion
        return  length == 16  ? [3] :   [0, 3]
```

Checkout/Checkout/Source/Model/Card+Scheme.swift
```diff
@@ -133,7 +138,43 @@ extension Card {
...
+        case .americanExpress:
+            return "americanExpress"
+        case .discover:
+            return "discover"
+        case .dinersClub:
+            return "dinersClub"
+        case .jcb:
+            return "jcb"
+        }
+    }
```
ehab-al-cko, line 163: Do we need string raw value?

Checkout/Checkout/Source/Model/Card+Scheme.swift
```diff
@@ -133,7 +138,43 @@ extension Card {
...
+      
+    /// map of card scheme to indexes of spaces in formatted card number string
+    /// eg. a visa card has gaps at 4, 8 and 12. 4242424242424242 becomes 4242 4242 4242 4242
+    public var cardGaps: [Int] {
+        switch self {
+        case .unknown:
+            return []
+        case .mada, .visa, .mastercard, .discover, .maestro, .jcb:
+            return [4, 8, 12]
```
ehab-al-cko, line 172: Should we add card gap for more than 16?

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
```diff
@@ -44,7 +44,8 @@ final class CardNumberValidator: CardNumberValidating {
   }
     
   func validateCompleteness(cardNumber: String) -> Result<ValidationScheme, ValidationError.CardNumber> {
-    let cardNumber = cardNumber.filter { !$0.isWhitespace }
+      "".first?.isWhitespace
```
ehab-al-cko, line 47: please add more details for this line

Source/Core/Models/CardScheme.swift
ehab-al-cko, line 8:
```suggestion
@frozen public enum CardScheme {
```

Checkout/Checkout/Source/Model/Card+Scheme.swift
ehab-al-cko, line 22:
```suggestion
@frozen public enum Scheme: CaseIterable, Equatable, Hashable {
```

Checkout/Checkout/Source/Model/Card+Scheme.swift
```diff
@@ -120,8 +122,11 @@ extension Card {
       switch self {
```
ehab-al-cko, line 122: Small suggestion if we can inject `cardNumberLength` then it would be small change then no need for extra code like `public var rawValue: String`, `allCases` and so on
```suggestion
func cvvLengths(with cardNumberLength: Int = 0) { 
      switch self {
```

Source/Core/Models/CardScheme.swift
alex-ioja-yang-cko: It depends on use case, they solve different problems. I think internally Card.Scheme is still the preferred way as it exposes the complete scheme object functionality, whereas CardScheme is a prettier version for the public interface.

Checkout/Checkout/Source/Model/Card+Scheme.swift
alex-ioja-yang-cko: The only reason its no longer having a rawValue of String is the limitation on enums with associated values. Their conformance cannot be inherited.

We do however still need the behaviour for encoding, decoding and lots of parsing happening, which is why we do need to enable behaviour ourselves.

Checkout/Checkout/Source/Model/Card+Scheme.swift
alex-ioja-yang-cko: I am not sure which paradigm you're referring to, but I see nothing wrong with Models containing logic that enables their functionality 🤔. In this case its true its not pretty, but functionally I see nothing wrong with this object.

Would be a different story if we were talking about a Data Object or other designated specific purpose kind of object, but this is just a definition of a card scheme with its standard behaviours.

Checkout/Checkout/Source/Model/Card+Scheme.swift
alex-ioja-yang-cko: Yes, all existing behaviour relies on that, for decoding from network, encoding and parsing.

Checkout/Checkout/Source/Model/Card+Scheme.swift
alex-ioja-yang-cko: @patrick-hoban-cko has had a look at that and said we'd do it separately as he wasn't sure. He was actually planning to ask @harry-brown-cko , so maybe we can sort it out here.

Source/Core/Models/CardScheme.swift
alex-ioja-yang-cko: I am not a fan of using frozen as this may as well change and it would become a breaking change for consumers. Would work if we were defining comparisons (< = >) or other things that would never change, but card schemes do come and go (example: we're looking at CB)

Checkout/Checkout/Source/Model/Card+Scheme.swift
alex-ioja-yang-cko: > I am not a fan of using frozen as this may as well change and it would become a breaking change for consumers. Would work if we were defining comparisons (< = >) or other things that would never change, but card schemes do come and go (example: we're looking at CB)

Checkout/Checkout/Source/Model/Card+Scheme.swift
alex-ioja-yang-cko: I have had this discussion before going down this path in our standup (I think you were present), and presented options, with @harry-brown-cko suggesting associated enum may be the way to go.

I have also researched your suggestion prior to that and it gets messy quickly because the CardValidator is inside the Checkout framework and as such any changes to its interface will have additional impact and increase complexity in multiple places.

Having tried both ways, I do agree with @harry-brown-cko 's opinion when we decided to go with this, that this is actually the easier to digest solution and is the cleanest solution to implement. Extending an interface for validator would have way more consequences than are apparent on first sight.

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
alex-ioja-yang-cko: @harry-brown-cko I've had to revert this change, the behaviour was not quite the same. The original validation is looking for invalid characters and will issue a rejection with a reason _invalidCharacters_ if such an occurange is encountered.

This function on the other hand attaches the counter to number of Int's found inside the provided String to correctly sync maestro length. Whilst current flows do have them behave similarly, the expectation of this function is different and merging them is not really appropriate.

Checkout/Checkout/Source/Model/Card+Scheme.swift
harry-brown-cko: [check out this blog for some examples](https://baymard.com/checkout-usability/credit-card-patterns)

I think we can start by using `[4, 8, 12, 16]` for these cards and fix specific formatting issues if they are raised by merchants. @ehab-al-cko @alex-ioja-yang-cko thoughts?

Checkout/Checkout/Source/Tokenisation/TokenDetailsFactory.swift
```diff
@@ -18,7 +18,7 @@ final class TokenDetailsFactory: TokenDetailsProviding {
       token: tokenResponse.token,
       expiresOn: tokenResponse.expiresOn,
       expiryDate: create(expiryMonth: tokenResponse.expiryMonth, expiryYear: tokenResponse.expiryYear),
-      scheme: tokenResponse.scheme.flatMap(Card.Scheme.init(rawValue:)),
+      scheme: tokenResponse.scheme.flatMap({ Card.Scheme(rawValue: $0) }),
```
harry-brown-cko, line 21: what is the purpose of this change? Function reference is cleaner IMO

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
```diff
@@ -2,6 +2,9 @@ import UIKit
 import Checkout
 
 public typealias Card = Checkout.Card
+public typealias Address = Checkout.Address
+public typealias Phone = Checkout.Phone
+public typealias Country = Checkout.Country
```
harry-brown-cko, line 7: Can we move these out to a centralised place? There are a good amount of types to re-export and we can also clean up implementation within the Frames module

Source/UI/NewUI/PaymentForm/Model/PaymentFormConfiguration.swift
```diff
@@ -24,11 +25,11 @@ public struct PaymentFormConfiguration {
      */
     public init(apiKey: String,
                 environment: Environment,
-                supportedSchemes: [Card.Scheme],
+                supportedSchemes: [CardScheme],
                 billingFormData: BillingForm?) {
         self.serviceAPIKey = apiKey
         self.environment = environment
-        self.supportedSchemes = supportedSchemes
+        self.supportedSchemes = supportedSchemes.compactMap { $0.mapToCheckoutCardScheme() }
```
harry-brown-cko, line 32: May have been left in by mistake. No need to remove optionals here
```suggestion
        self.supportedSchemes = supportedSchemes.map { $0.mapToCheckoutCardScheme() }
```

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
alex-ioja-yang-cko: Applied as suggested

Checkout/Checkout/Source/Model/Card+Scheme.swift
alex-ioja-yang-cko: Updated

Checkout/Checkout/Source/Model/Card+Scheme.swift
alex-ioja-yang-cko: When I spoke to @patrick-hoban-cko on it his opinion was to ask you @harry-brown-cko . Since you and @ehab-al-cko obviously also feel the space could be there, I have submitted it now as part of this PR 👍

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
alex-ioja-yang-cko: That was a big omission from my side when I filtered the commit 🤦‍♂️. When I'm not sure about syntax I usually just write it in the nearest open Xcode but I'd normally see it and not commit.

Thanks for noticing it.

Checkout/Checkout/Source/Tokenisation/TokenDetailsFactory.swift
alex-ioja-yang-cko: The syntax wouldn't compile. I think it worked before when rawValue was coming from compiler, whereas now we're mimicking that behaviour to reduce number of small syntax changes (and because its a familiar pattern), but ultimately Card.Scheme no longer has an actual String rawValue as far as compiler is concerned (associated value side effect)

Screenshot 2022-08-11 at 14 41 10

harry-brown-cko: Looked into this and its because of the default for the length. The init is `(String, Int) -> Card.Scheme?`

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
alex-ioja-yang-cko: Moved

Checkout/Checkout/Source/Tokenisation/TokenDetailsFactory.swift
alex-ioja-yang-cko: IDE is not very supportive, I'm not sure the syntax could be made to compile? In my book I am happy to use either approach, but only one allows for sufficient brevity and links well in this case

Source/UI/NewUI/PaymentForm/Model/PaymentFormConfiguration.swift
alex-ioja-yang-cko: The first one is a habit, I think wherever you do a map, making it compactMap is good practice as it future proofs your expectation of not having nils. Unless if you have a counter argument, I'm still preferring it as good practice (don't go down the optimisation argument, I'd just say the difference makes it an over-optimisation - nice to have if needed; large sets of expensive operation at some key point, not UI triggered for less then 100 possible elements) :).

Second one, I'm not a huge fan of computed properties for enabling this, as I find them to uglify (if that's a word) the syntax inside the object. They're useful sometimes and absolutely should be used, but wouldn't do just for something like this.

Checkout/Checkout/Source/Tokenisation/TokenDetailsFactory.swift
harry-brown-cko: What you have already is the cleanest way of doing this

Source/UI/NewUI/PaymentForm/Model/PaymentFormConfiguration.swift
harry-brown-cko: Disagree with the first point. We should be minimising the optionals in our codebase for the sake of simplicity. As a reader, compactMap infers that supportedSchemes is type [CardScheme?], which is quite confusing. Same way you wouldn't have optionals all over your models if the values were always present.

alex-ioja-yang-cko: > We should be minimising the optionals in our codebase for the sake of simplicity.

💯 Agree, but would point out its not always possible and there are genuine times you just need to handle optionals in different shapes and forms.

***

> As a reader, compactMap infers that supportedSchemes is type [CardScheme?], which is quite confusing

Not necessarily, as you can always check the type and see that is not true. It simply keeps it working if for some reason in the future you do need to handle optionals in that source array.

*** 

Found this nice comparison of [map vs flatMap vs compactMap](https://www.hackingwithswift.com/articles/205/whats-the-difference-between-map-flatmap-and-compactmap)

> map() will take a value out of its container, transform it using the code you specify, then put it back in its container. compactMap() does the same thing, but if your transformation returns an optional it will be unwrapped and have any nil values discarded.

To me this syntax just better protects the code as we know we absolutely want to have a result without optionals. If you end up changing the data source, that's a separate problem which you'll need looking at in isolation, but it will keep this assignment guaranteed to not handle optionals.

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
```diff
@@ -44,7 +44,7 @@ final class CardNumberValidator: CardNumberValidating {
   }
     
   func validateCompleteness(cardNumber: String) -> Result<ValidationScheme, ValidationError.CardNumber> {
-    let cardNumber = cardNumber.filter { Int("\($0)") != nil }
+    let cardNumber = cardNumber.filter { !$0.isWhitespace }
 
     guard validateDigitsOnly(in: cardNumber) else {
```
ehab-al-cko, line 49: I notice duplicate `.filter { !$0.isWhitespace }` 
I suggest to make String extension to remove white space and replaced in all Frames
```suggestion
 guard validateDigitsOnly(in: cardNumber.removeWhiteSpaces)
```

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
```diff
@@ -61,7 +61,7 @@ final class CardNumberValidator: CardNumberValidating {
   }
 
   func eagerValidate(cardNumber: String) -> Result<Card.Scheme, ValidationError.EagerCardNumber> {
-    let cardNumber = cardNumber.filter { Int("\($0)") != nil }
+    let cardNumber = cardNumber.filter { !$0.isWhitespace }
 
     guard validateDigitsOnly(in: cardNumber) else {
```
ehab-al-cko, line 66: Same

Checkout/Checkout/Source/Validation/Validators/CardNumberValidator.swift
```diff
@@ -59,7 +60,8 @@ final class CardNumberValidator: CardNumberValidating {
   }
 
   func eagerValidate(cardNumber: String) -> Result<Card.Scheme, ValidationError.EagerCardNumber> {
-    guard validateDigitsOnly(in: cardNumber.removeWhitespaces()) else {
+    let cardNumber = cardNumber.filter { !$0.isWhitespace }
+    guard validateDigitsOnly(in: cardNumber) else {
```
ehab-al-cko, line 64:
```suggestion
    let cardNumber = cardNumber.removeWhitespaces()
    guard validateDigitsOnly(in: cardNumber) else {
```

alex-ioja-yang-cko: Done

ehab-al-cko (approved):

PR #243, opened 2022-08-08
PIMOB-1369: Move fonts from SDK to sample app

alex-ioja-yang-cko:
[Jira task](https://checkout.atlassian.net/browse/PIMOB-1369)

## Further comments

- ✅ move Graphik font to example app from SDK
- ✅ use system font for all defaults (as they are present inside SDK and don't have access to font)
- ✅ remove calls to loading fonts that are not present or need loading for SDK


Source/UI/NewUI/BillingForm/Default Implementation/BillingForm/Header/Title/DefaultHeaderLabelFormStyle.swift
```diff
@@ -4,6 +4,6 @@ public struct DefaultHeaderLabelFormStyle: ElementStyle {
     public var backgroundColor: UIColor = .clear
     public var isHidden: Bool = false
     public var text: String = Constants.LocalizationKeys.BillingForm.Header.title
-    public var font: UIFont = UIFont(graphikStyle: .medium, size: Constants.Style.BillingForm.HeaderTitle.fontSize.rawValue)
+    public var font: UIFont = UIFont.systemFont(ofSize: Constants.Style.BillingForm.HeaderTitle.fontSize.rawValue)
```
deepesh-vasthimal-cko, line 7: 👍

iOS Example Frame Carthage/iOS Example Frame Carthage.xcodeproj/project.pbxproj
```diff
@@ -21,9 +21,22 @@
 /* End PBXAggregateTarget section */
 
 /* Begin PBXBuildFile section */
+		0B194AFC28A1132B001AEAFD /* UIFont+SFMono.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0B194AED28A1132B001AEAFD /* UIFont+SFMono.swift */; };
```
deepesh-vasthimal-cko, line 24: Do we need to add them to pod example as well ?

deepesh-vasthimal-cko (approved):

iOS Example Frame Carthage/iOS Example Frame Carthage.xcodeproj/project.pbxproj
alex-ioja-yang-cko: 🤔 good question. Not being in use for now its hard to check, so I think it should be left for the CustomStyle2 refresh task when we'll use it and we may find that yes, it is needed :). For now I'd leave it as it is, since nothing is failing and pipeline is definitely compiling all distribution formats and demo projects.

ehab-al-cko (approved):

PR #251, opened 2022-08-18
PIMOB-1379: Add Cardholder Name on Payment Form

alex-ioja-yang-cko:
[Jira task](https://checkout.atlassian.net/browse/PIMOB-1379)

1. Add new style for cardholder name input
2. Prevent duplication in cardholder name input between payment screen & billing. Will always remove billing one if duplicated
3. Remove some optionals
4. Add unit tests for new code and some missing ones

| Payment Form     | Billing Form |
| ---      | ---       |
| Simulator Screen Shot - iPhone 13 Pro Max - 2022-08-18 at 13 44 26 | Simulator Screen Shot - iPhone 13 Pro Max - 2022-08-18 at 13 44 41 |
| Simulator Screen Shot - iPhone 13 Pro Max - 2022-08-18 at 13 44 30 | Simulator Screen Shot - iPhone 13 Pro Max - 2022-08-18 at 13 44 35 |


Source/Resources/en.lproj/Localizable.strings
```diff
@@ -9,9 +9,12 @@
 "billingFormPhoneNumberHint" = "Phone Number";
 "billingAddressTitle" = "Billing address";
 "BillingFormCountryButtonTitle" = "Please select a country";
-"BillingFormOptional" = "Optional";
 "PaymentFormSummaryViewHintText" = "We need this information as an additional security measure to verify this card.";
 "AddBillingAddress" = "Add billing address";
+/* Accessory shown alongside input fields to denote that the input is not required for validation when completing request */
+"OptionalInputField" = "Optional";
+/* Title of input section on the Payment Form where user is expected to input the Cardholder name as displayed on card */
+"CardholderInputTitle" = "Cardholder Name";
```
patrick-hoban-cko, line 17: Change to 

_Cardholder name_

Please!

alex-ioja-yang-cko: Thanks for this, done now 👍

Source/UI/NewUI/PaymentForm/Default Implementation/Cardholder/DefaultCardholderFormStyle.swift
```diff
@@ -0,0 +1,22 @@
+//
+//  DefaultCardholderFormStyle.swift
+//  
+//
+//  Created by Alex Ioja-Yang on 16/08/2022.
+//
+
+import UIKit
+
+public struct DefaultCardholderFormStyle: CellTextFieldStyle {
```
deepesh-vasthimal-cko, line 10: Do we need to have Form in `DefaultCardholderFormStyle` ? Can this just be `DefaultCardholderStyle` ?

Source/UI/NewUI/PaymentForm/Default Implementation/Cardholder/DefaultCardholderFormStyle.swift
```diff
@@ -0,0 +1,22 @@
...
+    public var isMandatory: Bool = false
+    public var backgroundColor: UIColor = .clear
+    public var title: ElementStyle? = DefaultTitleLabelStyle(text: Constants.LocalizationKeys.PaymentForm.Cardholder.title)
+    public var hint: ElementStyle?
+    public var mandatory: ElementStyle? = DefaultTitleLabelStyle(
```
deepesh-vasthimal-cko, line 15: Question : why is mandatory variable of type `ElementStyle` and then its assigned to a `DefaultTitleLabelStyle`

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
```diff
@@ -15,13 +15,15 @@ public struct PaymentFormFactory {
...
                                                 environment: configuration.environment)
-    let viewModel = DefaultPaymentViewModel(checkoutAPIService: checkoutAPIService,
+    var viewModel = DefaultPaymentViewModel(checkoutAPIService: checkoutAPIService,
                                             cardValidator: cardValidator,
                                             logger: logger,
                                             billingFormData: configuration.billingFormData,
                                             paymentFormStyle: style.paymentFormStyle,
                                             billingFormStyle: style.billingFormStyle,
                                             supportedSchemes: configuration.supportedSchemes)
+    viewModel.preventDuplicateCardholderInput()
```
deepesh-vasthimal-cko, line 25: Question : Is `preventDuplicateCardholderInput` checking if Billing Form has another card holder text field and removes it ? if so do we need to really restrict this from SDK.

Source/UI/NewUI/PaymentForm/Default Implementation/Cardholder/DefaultCardholderFormStyle.swift
alex-ioja-yang-cko: I have kept the formatting with similar components "DefaultCardNumberFormStyle", "DefaultExpiryDateFormStyle" & "DefaultPayButtonFormStyle". I don't mind dropping the Form keyword but we'd need to do it across all the elements so we follow some pattern.

Source/UI/NewUI/PaymentForm/Default Implementation/Cardholder/DefaultCardholderFormStyle.swift
alex-ioja-yang-cko: All those labels are conforming to `ElementStyle` which nearly matches an UILabel's behaviour. `DefaultTitleLabelStyle` is an object conforming to `ElementStyle` that is being used for Title kind of Label representation.

... and one of the reasons I was against protocols for this, the complexity is not little 🙈

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
alex-ioja-yang-cko: > Question : Is preventDuplicateCardholderInput checking if Billing Form has another card holder text field and removes it

The changes inside `PaymentViewModel.swift` are implementing this logic. Simply put, if payment form has a cardholder name field, it will remove a fullName field from billing form cells. This is so we do not ask multiple times for the same property, either having a property we discard or allowing the user to change the same thing from 2 places. 

@patrick-hoban-cko has responded to this [on Jira task](https://checkout.atlassian.net/browse/PIMOB-1379?focusedCommentId=182585), with outcome:
_Agree that a merchant choosing the cardholder name to show on the payment form should supersede the billing address._


***


> if so do we need to really restrict this from SDK.

I'm not sure what you mean? The method is internal and consumer cannot call it. The call site is also part of the factory screen creation and should not expose anything to consumer that may allow them to call on demand, undo logic or in any way misuse it 🤔

Source/UI/NewUI/PaymentForm/Default Implementation/Cardholder/DefaultCardholderFormStyle.swift
deepesh-vasthimal-cko: Looking at the code I realised that `mandatory` is used for adding the optional text on a field or not and its style. Which now makes sense, But the naming is bit confusing as we have `isMandatory` and `mandatory` on `CellTextFieldStyle`. 
My understanding is 
`isMandatory` is for responsible to show or hide the mandatory field. 
`mandatory` is for the label style. 

For now we don't need to change anything on the PR, But some comments on these public properties would really help what they are responsible for. ( I dont think this is a POP issue but a improvement in naming )

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
deepesh-vasthimal-cko: Looking at the [API spec](https://api-reference.checkout.com/#operation/requestAToken) and also old UI. Should we not just add the card holder name to payment form always to keep it simple and not have additional logic to remove it from billing form screen. Please correct me if I have missed anything here.

deepesh-vasthimal-cko (approved):
Really Minor changes please have a look, But over all looks good.

patrick-hoban-cko:
@alex-ioja-yang-cko given that there are so many conflicts after @ehab-al-cko has deleted the old code, I suppose you will need to rebase and re-raise this?

Source/UI/NewUI/PaymentForm/Factory/PaymentFormFactory.swift
alex-ioja-yang-cko: I don't think you missed anything and you might actually have a better knowledge of Frames altogether so you might know why it has been implemented to display full name within billing to begin with? I am afraid I don't have an answer from my part, didn't see discoveries, plannings or tickets on that 🤷‍♂️. 

From my side I am only trying to avoid bad UX, but if you're thinking there's another path to it I am happy to discuss on how we can achieve.

ehab-al-cko (approved):

PR #259, opened 2022-08-25
PIMOB-1396: Fix Done states validation in the Billing Form 

ehab-al-cko:
[Jira ticket](https://checkout.atlassian.net/browse/PIMOB-1396?atlOrigin=eyJpIjoiMmE1NDZkZGE3YzA0NDYzNDhlZWE5ZjBhMjQwMWYzOTkiLCJwIjoiaiJ9)

## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
+++ b/Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
@@ -73,10 +73,10 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
         style.cells.forEach { type in
             let isMandatory = type.style?.isMandatory ?? true
             let value = type.getText(from: data)
-            let hasValue = value != nil
+            let isEmpty = value?.isEmpty ?? true
             country = data?.address?.country
-            textValueOfCellType[type.index] = hasValue ? value : (!isMandatory ? "" : nil)
-            errorFlagOfCellType[type.index] = hasValue ? false : (value != nil ? isMandatory : nil)
+            textValueOfCellType[type.index] = !isEmpty ? value : (!isMandatory ? "" : nil)
+            errorFlagOfCellType[type.index] = !isEmpty ? false : (value != nil ? isMandatory : nil)
         }
         if country == nil,
            let regionCode = Locale.current.regionCode,
@@ -172,9 +172,17 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
         if shouldRemoveText {
             textValueOfCellType[type.index] = nil
         }
+        var areAllFieldsAreFulfilled = true
 
-        let areAllFieldsAreFulfilled = textValueOfCellType.values.count == style.cells.count && !hasErrorValue
-        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldsAreFulfilled)
+        for cell in style.cells where cell.style?.isMandatory ?? false {
+            let isEmpty = textValueOfCellType[cell.index]?.isEmpty
+            guard isEmpty == false else {
+                areAllFieldsAreFulfilled = false
+                break
+            }
+        }
+
+        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldsAreFulfilled && !hasErrorValue)
     }
 
     private func validateTextOnEndEditing(textField: BillingFormTextField) {
--- a/Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
+++ b/Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
@@ -75,8 +75,8 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
             let value = type.getText(from: data)
             let isEmpty = value?.isEmpty ?? true
             country = data?.address?.country
-            textValueOfCellType[type.index] = !isEmpty ? value : (!isMandatory ? "" : nil)
-            errorFlagOfCellType[type.index] = !isEmpty ? false : (value != nil ? isMandatory : nil)
+            textValueOfCellType[type.index] = isEmpty ? (isMandatory ? nil : "") : value
+            errorFlagOfCellType[type.index] = isEmpty ? (value != nil ? isMandatory : nil) : false
         }
         if country == nil,
            let regionCode = Locale.current.regionCode,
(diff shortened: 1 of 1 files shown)


Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -73,10 +73,10 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
         style.cells.forEach { type in
             let isMandatory = type.style?.isMandatory ?? true
             let value = type.getText(from: data)
-            let hasValue = value != nil
+            let isEmpty = value?.isEmpty ?? true
             country = data?.address?.country
-            textValueOfCellType[type.index] = hasValue ? value : (!isMandatory ? "" : nil)
-            errorFlagOfCellType[type.index] = hasValue ? false : (value != nil ? isMandatory : nil)
+            textValueOfCellType[type.index] = !isEmpty ? value : (!isMandatory ? "" : nil)
+            errorFlagOfCellType[type.index] = !isEmpty ? false : (value != nil ? isMandatory : nil)
```
alex-ioja-yang-cko, line 79:
```suggestion
let hasValue = value?.isEmpty == false
country = data?.address?.country
textValueOfCellType[type.index] = hasValue ? value : (!isMandatory ? "" : nil)
errorFlagOfCellType[type.index] = hasValue ? false : (value != nil ? isMandatory : nil)
```

Reason for this suggestion is to avoid the `!isEmpty` for the comparison. Can also achieve by swapping the last 2 parts in the ternary operator and using `isEmpty`. Just thinking `hasValue` sounded better, but up to you how you'd prefer to avoid the syntax.

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -172,9 +172,17 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
...
-        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldsAreFulfilled)
+        for cell in style.cells where cell.style?.isMandatory ?? false {
+            let isEmpty = textValueOfCellType[cell.index]?.isEmpty
+            guard isEmpty == false else {
+                areAllFieldsAreFulfilled = false
+                break
+            }
+        }
+
+        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldsAreFulfilled && !hasErrorValue)
```
alex-ioja-yang-cko, line 185: I think we should be able to modify this syntax using `first(where: ` (not valid) and checking if such a value exists. I'll try to suggest syntax (lack of time now), but a unit tests for this would be very useful.

ehab-al-cko: IMHO, `first(where:` has differnt usage and not suitable for this case because it's searching and return first value of type `CellTextFieldStyle` where some condition. in this situation there's only check on mandatory value then another check if it's empty then change `areAllFieldsAreFulfilled` then break.

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
ehab-al-cko: I would like to keep it `isEmpty` like `value?.isEmpty`.

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
alex-ioja-yang-cko: Exactly why you don't need `areAllFieldsAreFulfilled`. Ignore that and rethink what you're looking for.

You're looking to see **if any field is not fulfilled**. 
You can tell that the first time you meet an item that is not fulfilled (this is the `first`). 
If such an early return is not met, you know you've successfully validated all fields are fulfilled!

In code that means something in the lines of:
```swift
let array = ["a", "b"]
let hasC = array.first(where: { $0 == "c" }) != nil
```

your logic appears slightly more complex, which is why I can only give you a generic example. If it still doesn't make sense I'll switch to your branch and find the actual code you need.

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
alex-ioja-yang-cko: I was not suggesting to remove `isEmpty` from the string check, I was suggesting to not use ternary operator with a negation.

```swift
// Current
!bool ? value1 : value2

// Suggested
bool ? value2 : value1
```

Ternary operator is great for being concise, but it is a complex line to understand. Booleans are easy to get wrong. I'd like us to keep it simple and don't use ternary operator while negating booleans.

If you review my code suggestion, I have not suggested we don't use `value?.isEmpty`, I was merely suggesting that you inverse the Boolean logic to avoid the negation on your assertion result

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
alex-ioja-yang-cko: Got the code, should be
```Swift
// Most on point
let areAllFieldFulfilled = style.cells.first(where: {
  guard let style = $0.style
            style.isMandatory else {
               return false
  }
  return textValueOfCellType[$0.index]?.isEmpty == true
}) == nil


// Breaking down the logic
let incompleteField = style.cells.first(where: {
  guard let style = $0.style
            style.isMandatory else {
               return false
  }
  return textValueOfCellType[$0.index]?.isEmpty == true
})
let areAllFieldsFulfilled = incompleteField == nil
```

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -173,8 +173,12 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
...
         }
 
-        let areAllFieldsAreFulfilled = textValueOfCellType.values.count == style.cells.count && !hasErrorValue
-        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldsAreFulfilled)
+        let areAllFieldFulfilled = style.cells.first(where: {
+          guard let style = $0.style, style.isMandatory else { return false }
+          return textValueOfCellType[$0.index]?.isEmpty == true
+        }) == nil
+
+        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldFulfilled && !hasErrorValue)
```
alex-ioja-yang-cko, line 181:
```suggestion
        guard !hasErrorValue else {
            editDelegate?.didFinishEditingBillingForm(successfully: false)
            return
        }
        
        let areAllFieldFulfilled = style.cells.first(where: {
          guard let style = $0.style, style.isMandatory else { return false }
          return textValueOfCellType[$0.index]?.isEmpty == true
        }) == nil

        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldFulfilled)
```

We can avoid iterating through the form altogether if an error value is found early, saving work time.

alex-ioja-yang-cko (approved):
I think we can do a minor optimisation improvement to prevent iterating through form if we know already there is an error (since this decides outcome anyway).

Otherwise 👍

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
ehab-al-cko: Added in a new [PR](https://github.com/checkout/frames-ios/pull/267)

PR #260, opened 2022-08-25
PIMOB:1413: Fix validate mandatory input provided for cvv

ehab-al-cko:
[Jira ticket](https://checkout.atlassian.net/browse/PIMOB-1413?atlOrigin=eyJpIjoiMWM2ZTU1OWU3OTUyNGM3OWEwMzQ2MWJjMGEzYmQ1MGUiLCJwIjoiaiJ9)

**Disabled button on error**

**Enabled button on success**

## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
+++ b/Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
@@ -148,7 +148,7 @@ extension DefaultPaymentViewModel: PaymentViewControllerDelegate {
   func securityCodeIsUpdated(result: Result<String, SecurityCodeError>) {
     switch result {
       case .failure:
-        cardDetails.cvv = ""
+        cardDetails.cvv = nil
       case .success(let cvv):
         cardDetails.cvv = cvv
     }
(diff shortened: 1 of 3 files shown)


Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
```diff
@@ -192,7 +192,9 @@ extension DefaultPaymentViewModel: PaymentViewControllerDelegate {
     // Check if security code is displayed and if so whether it is valid
     // This is business logic that wants Security code to be mandatory whenever its shown
     let isSecurityCodeRequired = paymentFormStyle?.securityCode != nil
-    if isSecurityCodeRequired && !cardValidator.isValid(cvv: cardDetails.cvv, for: cardScheme) { return }
+    if isSecurityCodeRequired {
+        guard let cvv = cardDetails.cvv, cardValidator.isValid(cvv: cvv, for: cardScheme) else { return }
+    }
```
alex-ioja-yang-cko, line 197: I am unsure how is this fixing it? The only way this may improve is if an empty CVV is valid option for the card scheme, which doesn't sound wrong?

ehab-al-cko: `securityCode` can be accept empty number string when card number is 19 digits. 
Issue is that `securityCode`  was before set to empty number string  when there's any error. 
Scenario: 
1. user enter in `card number` 19 digits `6799990100000000`
2. then user enter in `security code` value `0`
3. in our side, we validate `security code` then it's error 
4. then we return in the delegate `failure` then set `cardDetails.cvv` to be empty string ""
5. then we call `validateMandatoryInputProvided` 
6- then `isSecurityCodeRequired` is `true` `&&` `cardValidator.isValid(cvv: cardDetails.cvv, for: cardScheme)` is returning `true` then it's passing the `if condition`

alex-ioja-yang-cko: But all you've done with this PR is to make the default value `nil` instead of empty. As a rule of thumb we should avoid the ambiguity of nil.

Please try to add some CVV value and delete, then see if this code still validates correctly. Because your CVV will no longer be nil, it will be empty string, and exact same logic as previously will occur.

ehab-al-cko:
Duplicate

PR #264, opened 2022-08-31
PIMOB-1379: Improve cardholder input view behaviour

alex-ioja-yang-cko:
## Proposed changes

[Jira task](https://checkout.atlassian.net/browse/PIMOB-1379)
Improve handling of name input validation.

## Types of changes

## Checklist

## Further comments

Changes were needed due to a different location updating a field that by default would start empty. Used the same pattern as all other communication paths between viewModel and viewController.
Changed default names provided to cater for more extensive alphabet.


Source/UI/CommonUI/View/Component/InputView.swift
```diff
@@ -89,6 +89,10 @@ class InputView: UIView {
         updateTextFieldContainer(style: style)
         textFieldView.update(with: style.textfield)
         updateErrorView(style: style)
+
+        if let textFieldValue = textFieldValue {
+            textFieldView.textField.text = textFieldValue
+        }
```
ehab-al-cko, line 95: UI elements should not be overridden from another class. i suggest to set style.text
You may update it when update the style in  `CardHolderView` 
`func update(style: CellTextFieldStyle, text: String?)`
 ```suggestion
```

File: Source/UI/PaymentForm/View/CardholderName/CardholderView.swift [hunk 2]
```diff
@@ -30,9 +37,9 @@ final class CardholderView: UIView {
         fatalError("init(coder:) has not been implemented")
     }
 
-    func update(style: CellTextFieldStyle) {
+    func update(style: CellTextFieldStyle, text: String?) {
         self.style = style
-        cardholderInputView.update(style: style)
+        cardholderInputView.update(style: style, textFieldValue: text)
```
ehab-al-cko, line 42: I would suggest to override the text here 
```suggestion
       if let text = text {
            style.textfield.text = text
        }
        cardholderInputView.update(style: style)
```

Source/UI/CommonUI/View/Component/InputView.swift
alex-ioja-yang-cko: Absolutely agree and to be honest this is why I would have liked the UI properties to always be private. Just thought if they're already internal, will use them...

Will change as I do agree with the point you raised ➕

Source/UI/PaymentForm/View/CardholderName/CardholderView.swift
alex-ioja-yang-cko: Cool observation, I knew I must have missed something 🤦‍♂️

PR #268, opened 2022-09-01
PIMOB-1416: Add UI Test for full UI journey

alex-ioja-yang-cko:
## Proposed changes

[Jira task](https://checkout.atlassian.net/browse/PIMOB-1416)

## Types of changes

## Checklist

## Further comments

Ended up covering a few minor fixes that came out from the UI test.

### DIFF
--- a/Source/UI/BillingForm/View/BillingFormHeaderCell.swift
+++ b/Source/UI/BillingForm/View/BillingFormHeaderCell.swift
@@ -18,6 +18,7 @@ final class BillingFormHeaderCell: UIView {
     private lazy var doneButton: ButtonView = {
         let view = ButtonView().disabledAutoresizingIntoConstraints()
         view.delegate = self
+        view.isEnabled = false
         return view
     }()
 
@@ -43,10 +44,6 @@ final class BillingFormHeaderCell: UIView {
         doneButton.update(with: style.doneButton)
         cancelButton.update(with: style.cancelButton)
         headerLabel.update(with: style.headerLabel)
-    }
-
-    @objc private func doneAction() {
-        delegate?.doneButtonIsPressed()
     }
 
     @objc private func cancelAction() {
--- a/Source/UI/BillingForm/View/BillingFormHeaderCell.swift
+++ b/Source/UI/BillingForm/View/BillingFormHeaderCell.swift
@@ -45,10 +45,6 @@ final class BillingFormHeaderCell: UIView {
         headerLabel.update(with: style.headerLabel)
     }
 
-    @objc private func cancelAction() {
-        delegate?.cancelButtonIsPressed()
-    }
-
     private func shouldEnableDoneButton(flag: Bool) {
         doneButton.isEnabled = flag
     }
--- a/Source/UI/BillingForm/View/BillingFormHeaderCell.swift
+++ b/Source/UI/BillingForm/View/BillingFormHeaderCell.swift
@@ -16,9 +16,8 @@ final class BillingFormHeaderCell: UIView {
     }()
 
     private lazy var doneButton: ButtonView = {
-        let view = ButtonView().disabledAutoresizingIntoConstraints()
+        let view = ButtonView(startEnabled: false).disabledAutoresizingIntoConstraints()
         view.delegate = self
-        view.isEnabled = false
         return view
     }()
 
--- a/Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
+++ b/Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
@@ -163,7 +163,6 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
         let isEmptyText = textField.text?.isEmpty ?? true
         let isMandatoryField = type.style?.isMandatory ?? false
         let shouldRemoveText = isEmptyText && isMandatoryField
-        let hasErrorValue = errorFlagOfCellType.isEmpty || errorFlagOfCellType.values.allSatisfy({$0})
 
         if !isEmptyText {
             textValueOfCellType[type.index] = textField.text
@@ -173,17 +172,7 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
             textValueOfCellType[type.index] = nil
         }
 
-        guard !hasErrorValue else {
-            editDelegate?.didFinishEditingBillingForm(successfully: false)
-            return
-        }
-
-        let areAllFieldFulfilled = style.cells.first(where: {
-          guard let style = $0.style, style.isMandatory else { return false }
-          return textValueOfCellType[$0.index]?.isEmpty != false
-        }) == nil
-
-        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldFulfilled)
+        notifyContentChangeToDelegate()
     }
 
     private func validateTextOnEndEditing(textField: BillingFormTextField) {
@@ -196,6 +185,21 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
         textValueOfCellType[type.index] =  shouldSaveText ? textField.text : nil
 
         updatedRow = textField.tag
+    }
+
+    private func notifyContentChangeToDelegate() {
+        let hasErrorValue = errorFlagOfCellType.isEmpty || errorFlagOfCellType.values.allSatisfy({$0})
+        guard !hasErrorValue else {
+            editDelegate?.didFinishEditingBillingForm(successfully: false)
+            return
+        }
+
+        let areAllFieldFulfilled = style.cells.first(where: {
+          guard let style = $0.style, style.isMandatory else { return false }
+          return textValueOfCellType[$0.index]?.isEmpty != false
+        }) == nil
+
+        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldFulfilled)
     }
 }
 
@@ -246,6 +250,7 @@ extension DefaultBillingFormViewModel: BillingFormViewControllerDelegate {
         let index = BillingFormCell.country(nil).index
         textValueOfCellType[index] = country.name
         updatedRow = countryRow
+        notifyContentChangeToDelegate()
     }
 
     func getViewForHeader(sender: UIViewController) -> UIView? {
--- a/Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
+++ b/Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
@@ -180,7 +180,7 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
 
         let areAllFieldFulfilled = style.cells.first(where: {
           guard let style = $0.style, style.isMandatory else { return false }
-          return textValueOfCellType[$0.index]?.isEmpty == true
+          return textValueOfCellType[$0.index]?.isEmpty != false
         }) == nil
 
         editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldFulfilled)
--- a/Source/UI/PaymentForm/View/BillingFormSummaryView.swift
+++ b/Source/UI/PaymentForm/View/BillingFormSummaryView.swift
@@ -59,7 +59,6 @@ class BillingFormSummaryView: UIView {
 
         titleLabel.update(with: style.title)
         hintLabel.update(with: style.hint)
-        summaryLabel.update(with: style.hint)
         buttonView.update(with: style.button)
         imageContainerView.update(with: style.button.image, tintColor: style.button.imageTintColor)
         errorView.update(style: style.error)
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -68,6 +68,7 @@ workflows:
         inputs:
         - project_path: $FRAMES_SPM_PROJECT
         - scheme: $FRAMES_UITEST_SCHEME
+        - simulator_device: $UITEST_SIMULATOR_DEVICE
     - deploy-to-bitrise-io@1: {}
     - cache-push@2: {}
   _frames-slack:
@@ -296,6 +297,9 @@ workflows:
 
 app:
   envs:
+  - opts:
+      is_expand: false
+    UITEST_SIMULATOR_DEVICE: iPhone 13 Pro
   - opts:
       is_expand: false
     FRAMES_COCOAPODS_PROJECT: iOS Example Frame/iOS Example Frame.xcworkspace
--- a/iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift
+++ b/iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift
@@ -28,6 +28,22 @@ extension Factory {
 
         return viewController
     }
+
+    static func getCompleteUITestVC(completionHandler: @escaping (Result<TokenDetails, TokenisationError.TokenRequest>) -> Void) -> UIViewController {
+        let supportedSchemes: [CardScheme] = [.visa, .mastercard, .maestro, .americanExpress, .mada]
+
+        let configuration = PaymentFormConfiguration(apiKey: apiKey,
+                                                     environment: environment,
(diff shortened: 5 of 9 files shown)


Source/UI/PaymentForm/View/BillingFormSummaryView.swift
```diff
@@ -59,7 +59,6 @@ class BillingFormSummaryView: UIView {
 
         titleLabel.update(with: style.title)
         hintLabel.update(with: style.hint)
-        summaryLabel.update(with: style.hint)
```
alex-ioja-yang-cko, line 62: This method already gets called with correct style passed  on line 65 so this was doing an extra refresh with wrong style

Source/UI/BillingForm/View/BillingFormHeaderCell.swift
```diff
@@ -16,7 +16,7 @@ final class BillingFormHeaderCell: UIView {
     }()
 
     private lazy var doneButton: ButtonView = {
-        let view = ButtonView().disabledAutoresizingIntoConstraints()
+        let view = ButtonView(startEnabled: false).disabledAutoresizingIntoConstraints()
```
ehab-al-cko, line 19: please share some details why need to inject `startEnabled` f or common Button View?

Source/UI/BillingForm/View/BillingFormHeaderCell.swift
```diff
@@ -45,14 +45,6 @@ final class BillingFormHeaderCell: UIView {
...
     }
 
-    @objc private func doneAction() {
-        delegate?.doneButtonIsPressed()
-    }
-
-    @objc private func cancelAction() {
-        delegate?.cancelButtonIsPressed()
-    }
-
```
ehab-al-cko, line 55: please share details for deleting these delegates?

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -197,6 +186,21 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
...
+            return
+        }
+
+        let areAllFieldFulfilled = style.cells.first(where: {
+          guard let style = $0.style, style.isMandatory else { return false }
+          return textValueOfCellType[$0.index]?.isEmpty != false
+        }) == nil
+
+        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldFulfilled)
+    }
```
ehab-al-cko, line 203: this is duplicate from another PR. I think, you can share same logic to the other PR and i will test it.

bitrise.yml
```diff
@@ -296,6 +297,9 @@ workflows:
 
 app:
   envs:
+  - opts:
+      is_expand: false
+    UITEST_SIMULATOR_DEVICE: iPhone 13 Pro
```
ehab-al-cko, line 302: I think, main device is iPhone 13. 
```suggestion
    UITEST_SIMULATOR_DEVICE: iPhone 13
```

iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift
```diff
@@ -28,6 +28,22 @@ extension Factory {
 
         return viewController
     }
+
```
ehab-al-cko, line 31: please move `iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift` to UITest target

iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift
```diff
@@ -74,5 +90,66 @@ extension ThemeDemo {
                             billingFormStyle: billingFormStyle)
     }
 
+    static func buildCompleteUITest() -> PaymentStyle {
```
ehab-al-cko, line 93: Is there UI test for current default screen without duplicating the code?

iOS Example Frame/iOS Example Frame/Factory/Factory.swift
```diff
@@ -56,6 +56,9 @@ struct Factory {
   }
 
   static func getMatrixPaymentViewController(completionHandler: @escaping (Result<TokenDetails, TokenisationError.TokenRequest>) -> Void) -> UIViewController {
+    #if UITEST
+    return getCompleteUITestVC(completionHandler: completionHandler)
+    #endif
```
ehab-al-cko, line 61: this is code is for merchants. it should not include UITest logic

ehab-al-cko (changes requested):
Please add more details how to test it because it's not working with me

Source/UI/BillingForm/View/BillingFormHeaderCell.swift
alex-ioja-yang-cko: The button starts by default as enabled. Some buttons (like Pay and Billing Done) should start as disabled and only be enabled once validation passes.

This may not be visible to us in normal testing due to the speed of the action, but it becomes easily flagged by UI Test

Source/UI/BillingForm/View/BillingFormHeaderCell.swift
alex-ioja-yang-cko: This is dead code which was not called anywhere.

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
alex-ioja-yang-cko: Actually this is what testing is for, to flag issues in the code. And I think this PR did well to catch the boolean difference, as well as the fact the Country update itself was not triggering a validation recheck.

Also as mentioned in the other PR, the solution proposed there is to use inefficient and verbose syntax, not to actually fix a simple boolean comparison.

bitrise.yml
alex-ioja-yang-cko: I am not sure what "main device" refers to and what impact this proposed change has?

I chose the Pro version on the naive assumption that "was the simulator going to provide improve improved performance (as real device would claim), our simulator may theoretically run the tests faster". I don't think this is a valid assumption, but either way, no harm to use the pro version...

What is your motivation for not using the Pro version?

iOS Example Frame/iOS Example Frame/Factory/Factory.swift
alex-ioja-yang-cko: This is a demo app which comes with UITests. 
This is not code for merchants as it does not form the SDK consumer would use, its built on top of it.

iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift
alex-ioja-yang-cko: If we were to do UI tests without specific UI customisations designed for the UI tests we would end up with one of the following:
- demo designs that are not what UX wants
- UI tests that do not test for "edgy" behaviours

This can be improved by creating our demo designs to match UX templates and also contain our edge case scenarios, but then it may extend the number of buttons, making it harder to digest on first sight and would make the UI tests flakier as their UI itself may become subject to change on UX whim. 
Using the specific scheme with unique template overrides ensures the UI test is separated from the demo UI we present to clients.

iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift
alex-ioja-yang-cko: The call site for the API should be able to discover the declarations of that API.

In practice that means if we were to declare the methods contained within this file to UITest target, our application would need to have access to the UITest module. That would be a wrong dependency, as UITest needs to depend on source code it tests, not the other way around.

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -173,17 +172,7 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
...
-        let areAllFieldFulfilled = style.cells.first(where: {
-          guard let style = $0.style, style.isMandatory else { return false }
-          return textValueOfCellType[$0.index]?.isEmpty == true
-        }) == nil
-
-        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldFulfilled)
+        notifyContentChangeToDelegate()
     }
 
     private func validateTextOnEndEditing(textField: BillingFormTextField) {
```
alex-ioja-yang-cko, line 178: @ehab-al-cko please leave comments that require some response on a code comment, not on final review, so we can actually discuss.

You said
> Please add more details how to test it because it's not working with me

Could you please describe what exactly does that mean? The UI tests are passing on pipeline, so it appears to work on a 3rd party machine.

Source/UI/BillingForm/View/BillingFormHeaderCell.swift
ehab-al-cko: UI Test **must not** change normal common UI component. please undo it.

Source/UI/BillingForm/View/BillingFormHeaderCell.swift
ehab-al-cko: How come? if you did a little search you will find it in 
```
extension BillingFormViewController: BillingFormHeaderCellDelegate {

    func doneButtonIsPressed() {
        delegate?.doneButtonIsPressed(sender: self)
    }

    func cancelButtonIsPressed() {
        delegate?.cancelButtonIsPressed(sender: self)
    }
}```

File: Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift (hunk 4 above)
Comment (ehab-al-cko) [reply Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift:203]: Let's wait for other PR to be merged then this part will be solved.

File: bitrise.yml (hunk 5 above)
Comment (ehab-al-cko) [reply bitrise.yml:302]: I think, we need to agree on with device should run UI test

File: iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift (hunk 7 above)
Comment (ehab-al-cko) [reply iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift:93]: I think, one of ACs to test default screen as a normal user. 
[- _re create the same UI test with new UI flow_](https://checkout.atlassian.net/browse/PIMOB-1360)

File: iOS Example Frame/iOS Example Frame/Factory/Factory.swift (hunk 8 above)
Comment (ehab-al-cko) [reply iOS Example Frame/iOS Example Frame/Factory/Factory.swift:61]: It's an Example app and they will reuse code. We may discuss this with team in next standup

File: Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift (hunk 9 above)
Comment (ehab-al-cko) [reply Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift:178]: Like steps to  run UITest. It's not working with normal scheme.

File: Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift (hunk 4 above)
Comment (deepesh-vasthimal-cko) [reply Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift:203]: @alex-ioja-yang-cko As i understand this small change might conflict with other PR what @ehab-al-cko has already worked on, So would be worth discussing the changes before or waiting for other PRs to be merged. Will save time and effort.

File: Source/UI/BillingForm/View/BillingFormHeaderCell.swift (hunk 3 above)
Comment (alex-ioja-yang-cko) [reply Source/UI/BillingForm/View/BillingFormHeaderCell.swift:55]: Please see the code you referred to and the code being deleted

File: Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift (hunk 9 above)
Comment (alex-ioja-yang-cko) [reply Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift:178]: If its not working with normal scheme, could you please try to use UITest scheme instead? UITests are normally a separate target and as such commonly they require a separate scheme to run.

File: Source/UI/BillingForm/View/BillingFormHeaderCell.swift (hunk 2 above)
Comment (alex-ioja-yang-cko) [reply Source/UI/BillingForm/View/BillingFormHeaderCell.swift:19]: Not sure if you read my message properly. 
The current behaviour is buggy, enabling button when it should be disabled. 
The tests did what any tests do, catch bugs.
To get the test passing, code level changes were done.

Please don't make this harder, take a minute to understand replies and ask if unclear. Don't demand things without understanding the basics behind.

Comment (ehab-al-cko) [reply Source/UI/BillingForm/View/BillingFormHeaderCell.swift:19]: I think you misunderstand me. 
 if `ButtonView` has a new bug then it should be unit tested not a UI test.  `startEnabled ` is confusing name. you may change it to `isEnabled` like native apple button. Also, remove `isEnabled` from `ElementButtonStyle` if you want to inject button disable state.
```
public protocol ElementButtonStyle: ElementStyle {
    var isEnabled: Bool { get set }
```
Please undo changes here and a new PR to fix this issue with unit test would be great because this change will affect behaviour of buttons and needs to be tested alone.

Source/UI/BillingForm/View/BillingFormHeaderCell.swift
ehab-al-cko: ??

alex-ioja-yang-cko: The code being deleted is `doneAction` and `cancelAction`. You're looking at different things

Source/UI/BillingForm/View/BillingFormHeaderCell.swift
alex-ioja-yang-cko: Ok, so different problems to address here.

 

## Testing

Unit tests and UITests are both **tests** (same as snapshots, integration etc).
The general role for testing is to find bugs, enforce behaviours and prevent unintended changes.
Different kind of tests find bugs in different ways, for example:
- **unit tests** are great at handling business logic issues, inconsistent states, parsings, data flows, etc
- **snapshot tests** help identify UI level changes and ensure with some degree of certainty the layout, positioning, formatting of the UI, etc
- **UI tests** don't do either of these. They check the usability of the UI, navigation, presence ...

> if ButtonView has a new bug then it should be unit tested not a UI test.

As I have already stated 
_Some buttons (like Pay and Billing Done) should start as disabled and only be enabled once validation passes_ 
    and 
_This may not be visible to us in normal testing due to the speed of the action, but it becomes easily flagged by UI Test_

ButtonView does not have a bug as far as Unit Tests are concerned. Furthermore ButtonView is an UIView type and Unit Testing UI components is one of the reasons MVVM exists (**to not have to**).

 

## Inject button state

The way I see these components being designed to work, they get told their configuration (via a Style interface with multiple inheritances) what their state should be.

The bug is in the way ButtonView is being enabled very early in the lifecycle of the ViewController. This is due to its default state and the delay from UI being shown to the state being updated. 

I don't think affecting that update flow is part of the solution for this problem. The solution is to be able to control the start state, which is prior to the style becoming available.

So please, if you have a preference to a different approach feel free to detail it and as always I am happy to try if technically sound.

bitrise.yml
alex-ioja-yang-cko: Agreement starts with discussion. From my side I have proposed "iPhone 13 Pro" in the PR and as you have raised the subject I have asked you to have your say.

Please let me know when you'd like to touch this complex subject.

iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift
alex-ioja-yang-cko: Could you please justify why you think so? And what does "default screen as a normal user" actually mean to you, because I am unsure we have the same understanding for that.

ehab-al-cko (approved):

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
ehab-al-cko: i think, if this PR is merged then no need for other PR.

deepesh-vasthimal-cko (approved):

PR #270, opened 2022-09-01
PIMOB-1396: Fix billing done state issue

ehab-al-cko:
## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
+++ b/Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
@@ -177,13 +177,17 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
             editDelegate?.didFinishEditingBillingForm(successfully: false)
             return
         }
+        var areAllFieldsFulfilled = true
 
-        let areAllFieldFulfilled = style.cells.first(where: {
-          guard let style = $0.style, style.isMandatory else { return false }
-          return textValueOfCellType[$0.index]?.isEmpty == true
-        }) == nil
+        for cell in style.cells where cell.style?.isMandatory ?? false {
+            let isEmpty = textValueOfCellType[cell.index]?.isEmpty
+            guard isEmpty == false else {
+                areAllFieldsFulfilled = false
+                break
+            }
+        }
 
-        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldFulfilled)
+        editDelegate?.didFinishEditingBillingForm(successfully: areAllFieldsFulfilled)
     }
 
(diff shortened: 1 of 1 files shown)


Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -177,13 +177,17 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
             editDelegate?.didFinishEditingBillingForm(successfully: false)
             return
         }
+        var areAllFieldsFulfilled = true
 
-        let areAllFieldFulfilled = style.cells.first(where: {
-          guard let style = $0.style, style.isMandatory else { return false }
-          return textValueOfCellType[$0.index]?.isEmpty == true
```
alex-ioja-yang-cko, line 183: All the fix that was needed was to swap the boolean.

```swift
 return textValueOfCellType[$0.index]?.isEmpty == false
```

I don't think the first/for loop is the issue here. We are looking at a line of 
`textValueOfCellType[$0.index]?.isEmpty`, is there any way we could improve that for readability instead?

alex-ioja-yang-cko (changes requested):
Addressing the wrong problem

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
ehab-al-cko: There are not wrong or right problem. Both solutions are right. i will do your solution to unblock PR.

ehab-al-cko: I have tried your solution, it's not working. it disable done button on remove. please advice

alex-ioja-yang-cko: Have modified. @ehab-al-cko this is very simple Swift syntax, it would be nice if you actually took a minute to understand it before dismissing it. Again, the many levels of optionality present in this solution and the generic naming that doesn't carry meaning at use site make it hard to understand. Its not the for vs first looping...

I have now modified my suggestion to match the [tested solution on the PR you commented on](https://github.com/checkout/frames-ios/pull/268/files#r961532545)

ehab-al-cko: I have tried but it's not fully readable for me when != false and when == false. I would like to keep my solution as it's readable for me.

ehab-al-cko:
Fixed in another PR 
https://github.com/checkout/frames-ios/pull/268

PR #253, opened 2022-08-21
PIMOB-1319: Enable swift lint rules

ehab-al-cko:
[Jira ticket ](https://checkout.atlassian.net/browse/PIMOB-1319)
## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

What types of changes does your code introduce to Frames Ios?
_Put an `x` in the boxes that apply_

* [ ] Bugfix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

* [ ] Lint and unit tests pass locally with my changes
* [ ] I have added tests that prove my fix is effective or that my feature works
* [ ] I have added necessary documentation (if appropriate)

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #273, opened 2022-09-02
SwiftLint(3)

alex-ioja-yang-cko:
## Proposed changes

Follow up for [SwiftLint(2)](https://github.com/checkout/frames-ios/pull/272) removing some of the previously disabled rules for SwiftLint which were marked to be temporary.


deepesh-vasthimal-cko:
@alex-ioja-yang-cko was the work not already covered as part of PR https://github.com/checkout/frames-ios/pull/253 ?

iOS Example Frame/iOS Example Frame/Extension/UIViewExtension.swift
```diff
@@ -31,7 +31,10 @@ extension UIView {
 
   @IBInspectable var borderColorV: UIColor? {
     get {
-      return UIColor(cgColor: layer.borderColor!)
+      guard let layerBorderColor = layer.borderColor else {
```
deepesh-vasthimal-cko, line 34: 👍

iOS Example Frame/iOS Example Frame/Factory/Factory.swift
```diff
@@ -11,8 +11,11 @@ import Checkout
 import UIKit
 
 enum Factory {
+
+  // swiftlint:disable force_unwrapping
   static let successURL = URL(string: "https://httpstat.us/200")!
   static let failureURL = URL(string: "https://httpstat.us/403")!
```
deepesh-vasthimal-cko, line 17: should we just add // swiftlint:disable next force_unwrapping for line 17 and 15 as well instead of adding enable again on 18.

Source/CheckoutTheme.swift
```diff
@@ -1,5 +1,6 @@
 import UIKit
 
+// TODO: This file should have been removed
```
deepesh-vasthimal-cko, line 3: Is there something in backlog to remove this ? 

Was this not part of deleting old code. @ehab-al-cko

deepesh-vasthimal-cko (approved):

iOS Example Frame/iOS Example Frame/Factory/Factory.swift
alex-ioja-yang-cko: Sure, updated. Suppose it does make it less likely stuff could get added inside the block by mistake 👍

PR #276, opened 2022-09-07
PIMOB-NONE: Isolate consumption of PhoneNumberKit components to genuine use cases

alex-ioja-yang-cko:
## Proposed changes

Only use PhoneNumber input textfield types when needed and ensure they are not allocated memory otherwise

This saves about 10 MB of memory from being created but there seems to be another 4-8 MB we should investigate, which occur when textField becomes focussed (4MB) and then on dismiss (another 4-6 MB)


Source/UI/BillingForm/View/BillingFormTextFieldView.swift
```diff
@@ -89,6 +83,17 @@ class BillingFormTextFieldView: UIView {
         updateErrorView(style: style)
     }
 
+    func refreshLayoutComponents() {
+        let handlesPhoneNumber = isPhoneNumberType()
```
deepesh-vasthimal-cko, line 87: Question what does isPhoneNumberType() method do ?. Is this part of the phoneNumber kit framework.

iOS Example Frame/iOS Example Frame/Style/CustomStyle1.swift
```diff
@@ -201,7 +201,7 @@ private struct StyleOrganiser {
         var textAlignment: NSTextAlignment = .natural
         var text: String = ""
         var isSupportingNumericKeyboard: Bool = true
-        var height: Double = 30
+        var height: Double = 56
```
deepesh-vasthimal-cko, line 204: Any reason to change this ?

alex-ioja-yang-cko: Yes, there is a conflict in the current UI creation where it sometimes uses this constraint, but other times a default fallback value.

Should be raised as well at some point but thought I'd save you guys the horrors 😄

Source/UI/BillingForm/View/BillingFormTextFieldView.swift
deepesh-vasthimal-cko, line 87: Is this naming inconsistency ? we are using isPhoneNumberType to return Bool with is prefix and storing in handlesPhoneNumber without a is prefix. Should we keep them consistent across.

deepesh-vasthimal-cko (approved):
Looks good. just few minor comments.

iOS Example Frame/iOS Example Frame/Style/CustomStyle1.swift
deepesh-vasthimal-cko: Was this broken with 30 and now changing to 56 to follow UI designs ?

Source/UI/BillingForm/View/BillingFormTextFieldView.swift
alex-ioja-yang-cko: I don't want to confuse method and property. If `isPhoneNumberType` was also a property it would be ok, but as its a function I tried to name it differently and not change it. But happy to change as it might help make more sense.

iOS Example Frame/iOS Example Frame/Style/CustomStyle1.swift
alex-ioja-yang-cko: Sometimes the default from the fallback is applied, other times the style provided one is applied.

In practice, this leads to some of the debug warnings of unsatisfied constraints you see in Xcode when you run the app. This number specifically made the PhoneNumberInput become 30 pixels height on refresh (other components don't trigger that refresh so they wouldn't have shown...

alex-ioja-yang-cko: Unsatisfied constraints warnings, would be nice to fix at some point...

Screenshot 2022-09-07 at 16 31 39

ehab-al-cko:
Hi @alex-ioja-yang-cko ,  I have found another memory issue for PhoneKit with warning. please advice regarding this issue. should be solved in current PR or another PR? Thanks
Screenshot 2022-09-08 at 11 38 57 am

ehab-al-cko:
Also found another issue, after dismissing the billing form screen, `BillingFormPhoneNumberText` still not released with 4 memory locations.
Screenshot 2022-09-08 at 11 44 51 am

ehab-al-cko (approved):
I think, it improved the memory usage. I left some comments. Feel free to add new fix PRs.

PR #254, opened 2022-08-22
PIMOB-1381: Enable control over components required for payment input validation

alex-ioja-yang-cko:
[Jira task](https://checkout.atlassian.net/browse/PIMOB-1381)

Add business logic that enables the configuration to show fields in the payment form (cardholder name, billing) that can be compulsory or optional for enabling payment.
Additionally enable Security Code being shown / removed from payment form and enforce its validation if shown, whilst enabling it to also be missing and not block form.

PR #279, opened 2022-09-08
PIMOB-1381: Clean theme demo code

ehab-al-cko:
## Proposed changes

Clean Theme Demo code to be more readable, maintainable, scalable and understandable.
it's also required by swiftlint to break down function.
`swiftlint:disable function_body_length
`
## Types of changes

## Checklist

## Further comments


iOS Example Frame/iOS Example Frame/ThemeDemo.swift
```diff
@@ -9,68 +9,207 @@
 import UIKit
 import Frames
 
-// swiftlint:disable function_body_length
-enum ThemeDemo {
+struct ThemeDemo {
```
ehab-al-cko, line 12:
```suggestion
enum ThemeDemo {
```

Source/UI/PaymentForm/Factory/PaymentFormFactory.swift
```diff
@@ -7,7 +7,7 @@ public enum PaymentFormFactory {
   internal static var sessionCorrelationID = ""
 
   public static func buildViewController(configuration: PaymentFormConfiguration,
-                                         style: PaymentStyle,
+                                         style: FramesStyle,
```
ehab-al-cko, line 10: `PaymentStyle` is changed to `FramesStyle` because there's another `PaymentFormStyle` which will be confusing

iOS Example Frame/iOS Example Frame/Style/OtherStyle.swift
```diff
@@ -0,0 +1,274 @@
...
+//  Created by Alex Ioja-Yang on 18/08/2022.
+//  Copyright © 2022 Checkout. All rights reserved.
+//
+import UIKit
+import Frames
+
+enum OtherStyle {
+
+    // MARK: Matrix Style with one function
+    static func buildWithOneFunction() -> FramesStyle {
```
alex-ioja-yang-cko, line 14: I cannot agree with this PR as discussed previously. Happy to be overridden if need be, but previous ThemeDemo was 76 lines all in, with nearly 60 lines of linear code describing the components in sequence and "demoing the Theme". 

You're proposing replacing it with nearly 300 lines of complex unlinear code that negates the reason we created Theme (convenient simple way to declare the UI config).

Additionally, we should not put MARK on top of every function, it doesn't make sense. MARK exists to break down code into groups for easier navigation. If you break down into nearly every function, then the function itself was already the "mark".

iOS Example Frame/iOS Example Frame/Style/MatrixStyle.swift
```diff
@@ -199,6 +185,29 @@ private enum StyleOrganiser {
         var error: ElementErrorViewStyle? = ErrorViewStyle(text: "Insert a valid security code")
     }
 
+    struct PayButtonSection: ElementButtonStyle {
```
alex-ioja-yang-cko, line 188: This is actually a good change, but I don't think "Section makes sense when looking at the button.

User inputs are sections because they can be composed of:
- title
- subtitle
- optional mark
- input
- error

This is not the case for the button, which is just a button.

```suggestion
    struct PayButtonStyle: ElementButtonStyle {
```

iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift
```diff
@@ -11,7 +11,13 @@ import Frames
...
 
 #if UITEST
-extension Factory {
+enum UITestFactory {
+    // swiftlint:disable:next force_unwrapping
+    static let successURL = URL(string: "https://httpstat.us/200")!
+    // swiftlint:disable:next force_unwrapping
+    static let failureURL = URL(string: "https://httpstat.us/403")!
+    static let apiKey = "pk_test_6e40a700-d563-43cd-89d0-f9bb17d35e73"
+    static let environment: Frames.Environment = .sandbox
```
alex-ioja-yang-cko, line 20: What was gained by this besides adding UITest named objects inside the application source code?

iOS Example Frame/iOS Example Frame/Style/OtherStyle.swift
```diff
@@ -0,0 +1,274 @@
...
+//  OtherStyle.swift
+//  iOS Example Frame
+//
+//  Created by Alex Ioja-Yang on 18/08/2022.
+//  Copyright © 2022 Checkout. All rights reserved.
+//
+import UIKit
+import Frames
+
+enum OtherStyle {
```
alex-ioja-yang-cko, line 11: OtherStyle might make sense when looking at it from the application buttons, but it does not make sense when thinking "why it exists".

ThemeDemo was here to show how Theme can be used to achieve custom UI in a simple demonstrative way. Renaming it to OtherStyle doesn't help for its purpose as a way to demonstrate anything.

iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift
```diff
@@ -46,9 +52,9 @@ extension Factory {
     }
 }
 
-extension ThemeDemo {
+enum UITestThemeDemo {
```
alex-ioja-yang-cko, line 55: I do not agree with this change. Objects should open for extension and this was previously done in such a way that more extreme examples could be achieved for UI testing. Renaming to UITestThemeDemo brings objects named `UITest` into the application.

I strongly believe unit tests should complement existing code and not require new objects to be created for it! An extension was the complement to existing code, for UI Testing to occur with specific cases, complementing existing logic. This proposed change undoes that and creates new objects for testing!

alex-ioja-yang-cko (changes requested):
Only change I agree with is the need for the PayButton to be better included in the Style.

Otherwise, instead of renaming ThemeDemo to Other to please an UIButton, you could have renamed that button to "MinimalThemeDemo" and keep the **demo code** as it is, which would better help it achieve the purpose of **demoing Theme**

iOS Example Frame/iOS Example Frame/Factory/Factory+UITest.swift
```diff
@@ -11,7 +11,13 @@ import Frames
 import Checkout
 
 #if UITEST
-extension Factory {
+enum UITestFactory {
```
deepesh-vasthimal-cko, line 14: Note related to this PR : Just a question why are all these enums and not structs ? examples are `enum Factory`, `enum ThemeDemo` etc. My option for enum would be to have similar values grouped as cases and not add static functions. Just wanted to understand this pattern.

deepesh-vasthimal-cko:
@ehab-al-cko can you please help me understand the purpose of this PR, I can see that its been marked with PIMOB-1381 which is `"All components should have an isMandatory property. Security code and billing details are not mandatory"` which was fixed under the PR https://github.com/checkout/frames-ios/pull/254. Just need to understand the origin of this change.

Source/UI/PaymentForm/Factory/PaymentFormFactory.swift
deepesh-vasthimal-cko: My understanding is, buildViewController was returning back some sort of Payment Form View controller using a `PaymentFormConfiguration` and `PaymentStyle`. Removing the Payment prefix here would be more confusing as everything else will need same prefix like `FramesFormConfiguration`. 

My suggestion would be to change ```
buildViewController(configuration: PaymentFormConfiguration,
                                         style: PaymentStyle,
```
to 
```
buildPaymentFormViewController(configuration: FormConfiguration,
                                         style: FormStyle,
```

Please feel free to add your thoughts.

PR #281, opened 2022-09-09
PIMob-1058: Missing border line colors for UIStackView on iOS 12

ehab-al-cko:
[Jira issue](https://checkout.atlassian.net/browse/PIMOB-1058?focusedCommentId=186713)

## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/UI/CommonUI/View/Component/InputView.swift
+++ b/Source/UI/CommonUI/View/Component/InputView.swift
@@ -29,6 +29,10 @@ class InputView: UIView {
         LabelView().disabledAutoresizingIntoConstraints()
     }()
 
+    private lazy var textFieldContainerBorder: UIView = {
+        UIView().disabledAutoresizingIntoConstraints()
+    }()
+
     private(set) lazy var textFieldContainer: UIStackView = {
         let view = UIStackView().disabledAutoresizingIntoConstraints()
         view.axis = .horizontal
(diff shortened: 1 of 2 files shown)


Source/UI/CommonUI/View/Component/InputView.swift
```diff
@@ -29,6 +29,10 @@ class InputView: UIView {
         LabelView().disabledAutoresizingIntoConstraints()
     }()
 
+    private lazy var textFieldContainerBorder: UIView = {
+        UIView().disabledAutoresizingIntoConstraints()
+    }()
```
alex-ioja-yang-cko, line 34: Why do we need to add a container to the container for the border functionality? Could you please explain why / how this is supposed to work?

ehab-al-cko: Sure. The UIStackView is a nonrendering subclass of UIView., pre iOS 14. From iOS14+ it can render normally so it can not render borderlines.

alex-ioja-yang-cko: Cool, makes sense. Normally not a fan of making `if available` 's in code, but this may be a good opportunity to simplify in the future once the behaviour is standardly working without surprises?

I'm thinking of having logic like:
```swift
if #available(iOS 14, *) {
    // Modify or add stackview
} else {
    // Modify or add container of container 
}
```

I think this would impact all parts, but would ensure we don't deliver obsolete code once we are safe to strip it away.

And here we could use optionality & availability to also ensure it gets stripped:
```suggestion -2
    private lazy var textFieldContainerBorder: UIView? = {
        if #available(iOS 14.0, *) { return nil } 
        else { UIView().disabledAutoresizingIntoConstraints() }
    }()
```

alex-ioja-yang-cko: Oh and you'd also not have to "fix" unit tests, since we always run them on a "latest" device, so:
- no need to fix something that's not broken
- preserve memory footprint and layout for latest OSs
- help cleanup once no longer needed (becomes obsolete, in maybe 2 years)

ehab-al-cko: IMHO, The suggested solution `if #available(iOS 14.0, *) { return nil }` will have two implementations for borderline, one for iOS before iOS 14 and the other from iOS 14 and after. 
Pros:
- The current simple solution is for all starting from iOS 12+ with common code which will be easy to handle.
-  we don't have to add extra code per iOS version which I still vote for it. I think we will support min iOS 14 after on Oct 2025. this solution will work forever so no need to change it.
- Regarding memory, Native simple UIView will not preserve a lot like PhoneTextfield which uses logic from 3rd party framework.
- Regarding Clean up, We will need to clean up every time we upgrade min supported iOS. I can add comment regarding this to help us in the future.
- Regarding unit tests. it's small and simple. Also it must be done even with suggest solution.

alex-ioja-yang-cko (approved):

PR #312, opened 2022-09-22
PIMOB-1434: Fix Maestro card number formatting

alex-ioja-yang-cko:
[Jira task](https://checkout.atlassian.net/browse/PIMOB-1434)

Use Maestro Regex for eager matching by removing requirement to match full length.

Extend test case to cover for number flagged in task.


Checkout/Checkout/Test/Validation/CardNumberValidatorTests.swift
```diff
@@ -31,6 +31,7 @@ final class CardNumberValidatorTests: XCTestCase {
       "5297412542005689": .success(.mada),
       "6759649826438453": .success(.maestro(length: 16)),
       "6016607095058666": .success(.maestro(length: 16)),
+      "5000550000000029": .success(.maestro(length: 16)),
```
deepesh-vasthimal-cko, line 34: Do we not have a test case to check the number of spaces are added for new formatting ?

alex-ioja-yang-cko: We do have some separate tests to confirm formatting is accurate. The issue here was the eager regex was not matching the scheme, and without a scheme there's no formatting.

Once the scheme is identified correctly (proposed regex change), the formatting occurs as expected.

alex-ioja-yang-cko: Not sure if that made sense, so another way to look at it, the flow from number to formatting is:

Input
 -> Eager Matching
   -> Formatting
   
The formatting failure observed was a symptom of the Eager Matching failing. That's fixed and confirmed by unit tests. 

So the formatting was never the issue, hence no formatting unit test update :).

deepesh-vasthimal-cko (approved):

PR #315, opened 2022-10-06
PIMOB-1362(2): Create & Use paymentFormSubmitted logging event

alex-ioja-yang-cko:
## Proposed changes

Create new logging event and trigger on button press.

Found an inconsistency in some unit test where a helper method was named inaccurately.

### DIFF
--- a/Source/Core/Logging/FramesLogEvent.swift
+++ b/Source/Core/Logging/FramesLogEvent.swift
@@ -37,6 +37,7 @@ enum FramesLogEvent: Equatable, PropertyProviding {
 
     case paymentFormInitialised(environment: Environment)
     case paymentFormPresented
+    case paymentFormSubmitted
     case billingFormPresented
     case threeDSWebviewPresented
     case threeDSChallengeLoaded(success: Bool)
@@ -53,6 +54,8 @@ enum FramesLogEvent: Equatable, PropertyProviding {
             return "payment_form_initialised"
         case .paymentFormPresented:
             return "payment_form_presented"
+        case .paymentFormSubmitted:
+            return "payment_form_submitted"
         case .billingFormPresented:
             return "billing_form_presented"
         case .threeDSWebviewPresented:
@@ -70,6 +73,7 @@ enum FramesLogEvent: Equatable, PropertyProviding {
         switch self {
         case .paymentFormInitialised,
              .paymentFormPresented,
+             .paymentFormSubmitted,
              .billingFormPresented,
              .threeDSWebviewPresented:
             return .info
@@ -83,8 +87,9 @@ enum FramesLogEvent: Equatable, PropertyProviding {
 
     var properties: [Property: AnyCodable] {
         switch self {
-        case .billingFormPresented,
-             .threeDSWebviewPresented:
+        case .paymentFormSubmitted,
+                .billingFormPresented,
+                .threeDSWebviewPresented:
             return [:]
         case .paymentFormPresented:
             return [Property.locale: Locale.current.identifier].mapValues(AnyCodable.init(_:))
--- a/Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
+++ b/Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
@@ -154,14 +154,15 @@ extension DefaultPaymentViewModel: PaymentViewControllerDelegate {
     validateMandatoryInputProvided()
   }
 
-  func payButtonIsPressed() {
-    guard let card = cardDetails.getCard() else { return }
-    isLoading = true
-    checkoutAPIService.createToken(.card(card)) { [weak self] result in
-      self?.isLoading = false
-      self?.cardTokenRequested?(result)
+    func payButtonIsPressed() {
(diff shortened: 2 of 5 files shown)


Source/Core/Logging/FramesLogEvent.swift
```diff
@@ -53,6 +54,8 @@ enum FramesLogEvent: Equatable, PropertyProviding {
             return "payment_form_initialised"
         case .paymentFormPresented:
             return "payment_form_presented"
+        case .paymentFormSubmitted:
+            return "payment_form_submitted"
```
deepesh-vasthimal-cko, line 58: @maxim-nosov-cko @chintan-soni-cko @patrick-hoban-cko will have to create similar events for Android as well in the backlog once this PR is approved and merged.

Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
```diff
@@ -154,14 +154,15 @@ extension DefaultPaymentViewModel: PaymentViewControllerDelegate {
...
   }
 
-  func payButtonIsPressed() {
-    guard let card = cardDetails.getCard() else { return }
-    isLoading = true
-    checkoutAPIService.createToken(.card(card)) { [weak self] result in
-      self?.isLoading = false
-      self?.cardTokenRequested?(result)
+    func payButtonIsPressed() {
+        guard let card = cardDetails.getCard() else { return }
```
deepesh-vasthimal-cko, line 158: Unrelated to PR , Question : What happens if cardDetails.getCard() is nil, will this do a silent return. Or this will never happen.

Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
```diff
@@ -154,14 +154,15 @@ extension DefaultPaymentViewModel: PaymentViewControllerDelegate {
...
+        logger.log(.paymentFormSubmitted)
```
deepesh-vasthimal-cko, line 159: Should we move this line to ( line 158 instead ) first thing in the func payButtonIsPressed so it always gets logged when user taps on button ?

Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
alex-ioja-yang-cko: As long as the button enabling / disabling is synced correctly it should not happen. But if it did, it would just silently fail. This could actually be logged as a `warn`

Tests/UI/BillingForm/ViewModel/PaymentViewModelTests.swift
```diff
@@ -626,7 +660,7 @@ class PaymentViewModelTests: XCTestCase {
         BillingForm(name: "John", address: Address(addressLine1: "Kong Drive", addressLine2: "Blister", city: "Dreamland", state: nil, zip: "DR38ML", country: Country.allAvailable.first), phone: nil)
     }
     
-    private func makeBillingFormStyle(isCardholderMandatory: Bool = false,
+    private func makePaymentFormStyle(isCardholderMandatory: Bool = false,
```
deepesh-vasthimal-cko, line 663: What is the purpose of this renaming ? from makeBillingFormStyle to makePaymentFormStyle

deepesh-vasthimal-cko (approved):
Few Minor comments otherwise looks good.

Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
alex-ioja-yang-cko: Problem is if it fails the `guard`, **payment form hasn't been submitted**, so it would be an inconsistency in the log?

Tests/UI/BillingForm/ViewModel/PaymentViewModelTests.swift
alex-ioja-yang-cko: The output was a PaymentFormStyle, not a Billing. So the function name was not consistent with the object being created.

Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
alex-ioja-yang-cko: As long as the button enabling / disabling is synced correctly it should not happen. But if it did, it would just silently fail. 

This could actually be logged as a `warn`? I think the warn comes 2 PRs later, so if you're keen on reporting that, I'll open a follow up after all other events have their PRs open.

alex-ioja-yang-cko: As warn comes in next PR (#316 ), I've opened a PR for this after all the other logging work (#320 )

Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
alex-ioja-yang-cko: As mentioned on other comment thread, I opened PR #320 to trigger a warn event if the button press was to fail silently (from UI perspective)

PR #317, opened 2022-10-07
PIMOB-XXXX: Prevent multiple requests from UI

alex-ioja-yang-cko:
## Proposed changes

Disable button to prevent user from triggering multiple requests to tokenisation whilst one is already in progress. 

This adds a loading overlay built from the disabled button state, whilst making other content transparent, with a small smooth animation.

Have also added an update to the Token (got other tasks and not enough time to keep switching back & forth for that).

## Default UI
https://user-images.githubusercontent.com/102028170/194516272-f631dd56-b116-4520-a059-a912ad5ca244.mp4

## Matrix Style
https://user-images.githubusercontent.com/102028170/194516474-b5bd7b60-86a4-4ad9-83df-8463b440b260.mp4

## Theme Demo
https://user-images.githubusercontent.com/102028170/194516503-6b58bbee-74ed-4846-ba29-16ec63869877.mp4

## Edge case 
<sub>User makes a change to the input whilst the call is ongoing</sub>

https://user-images.githubusercontent.com/102028170/194516695-91682c63-524d-45f0-8d4c-9017d338bf20.mp4


Checkout/Checkout/Integration/CheckoutAPIServiceIntegrationTests.swift
```diff
@@ -66,9 +66,9 @@ final class CheckoutAPIServiceIntegrationTests: XCTestCase {
 
     // details associated with default apple pay token
     let expectedApplePayDetails = ApplePayDetails(
-      expiryDate: try! CardValidator(environment: .sandbox).validate(expiryMonth: 9, expiryYear: 22).get(),
-      bin: "520424",
-      last4: "6937"
```
alex-ioja-yang-cko, line 71: Unrelated to this PR, but blocker for pipeline. Too many tasks at the moment to separate.

Checkout/Checkout/Integration/StubProvider.swift
```diff
@@ -195,12 +195,12 @@ enum StubProvider {
   private static let tokenData = """
     {
         "version": "EC_v1",
-        "data": "cWjjsdADd0e6O2cba0cpF1458RnoR0DAxvP/lfwBPz/HfcfBQY8Wld/m+k8WjBpwn504khNVFuJ3pGdZwUHUGVv5PrujtiglkCH3B+uMDYSzalfCi6/wQf0zDmHJVoghZb0dDz4Xrh4NicTYqcmTBzPRSEcisDxxbyzH8nD2jsd/bBA2Q+jzCGhWcV/gKKjLY2XQOEc0RpHyVAKtanGSfCpCZQPQ6D/19nzvWQe9kuqfLPqmB+gBi6Z7eYMhgdLKUZpXd/m3TL3AjrOeZxvoUr++VEI+XqRqDJT5GH+cmKRlR4/ezM9y4fwdO6DJUmh8kx1iWh0CFO38KmPM+dEcL7vKS2UkIyw4FExt4LLtgGBF2P6xu7JWUyRGnLYXRfZxpd0VpgGqCqLcDRn1",
```
alex-ioja-yang-cko, line 198: Unrelated to this PR, but blocker for pipeline. Too many tasks at the moment to separate.

alex-ioja-yang-cko:
Alex Byrne (UI) has [approved](https://checkout.slack.com/archives/GS5KC6HU3/p1665136778983259?thread_ts=1664870400.500179&cid=GS5KC6HU3) the change from his side

Checkout/Checkout/Integration/StubProvider.swift
deepesh-vasthimal-cko: Was the data expired ? What was the reason to update this.

Source/UI/CommonUI/View/Core/ButtonView.swift
```diff
@@ -31,6 +37,8 @@ class ButtonView: UIView {
         LabelView().disabledAutoresizingIntoConstraints()
     }()
 
+    private var loadingOverlay: UIView?
```
deepesh-vasthimal-cko, line 40: I see there is a edge case in your videos where user tries to update card details when the pay button is pressed ? Are we not disabling the screen when pay button is pressed ?. Just want to understand what is the expected behaviour in this case.

deepesh-vasthimal-cko (approved):
Looks good, just one question on the edge case and what should be the expected behaviour ?

Source/UI/CommonUI/View/Core/ButtonView.swift
alex-ioja-yang-cko: No, we are not disabling the screen while pending the tokenisation request and we are not communicating anything to application owner either, so they could not block UI on their side.

There was an activity indicator placed in the navigation controller but it would not prevent multiple requests being triggered.

[Recent Slack discussion](https://checkout.slack.com/archives/GS5KC6HU3/p1664870400500179)

alex-ioja-yang-cko: @deepesh-vasthimal-cko I will merge this for now as it prevents unexpected behaviours from occurring, but happy to change again once @patrick-hoban-cko is back if you guys want a different solution.

I'd just like to limit possibilities that things go wrong for the user for now, and I think you agreed, it does achieve that

Checkout/Checkout/Integration/StubProvider.swift
alex-ioja-yang-cko: Yes, the data contained was of a test Apple Token, which comes with certain information, part of which is an expiry date. That makes it un-processable to our backend which then continues to fail on our "integration tests", blocking the pipelines

PR #318, opened 2022-10-07
PIMOB-1362(4): Create & Use paymentFormCanceled logging event

alex-ioja-yang-cko:
## Proposed changes

Create new logging event and trigger when the back arrow is pressed to dismiss the payment view controller. This is user initiated and would equate to a cancel action from the user.

### DIFF
--- a/Source/UI/PaymentForm/ViewModel/PaymentViewModel.swift
+++ b/Source/UI/PaymentForm/ViewModel/PaymentViewModel.swift
@@ -23,6 +23,7 @@ protocol PaymentViewModel {
   var cardTokenRequested: ((Result<TokenDetails, TokenisationError.TokenRequest>) -> Void)? { get set }
   func updateAll()
   func viewControllerWillAppear()
+  func viewControllerCancelled()
   mutating func preventDuplicateCardholderInput()
 }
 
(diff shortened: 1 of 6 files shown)


Source/UI/PaymentForm/ViewModel/PaymentViewModel.swift
```diff
@@ -23,6 +23,7 @@ protocol PaymentViewModel {
   var cardTokenRequested: ((Result<TokenDetails, TokenisationError.TokenRequest>) -> Void)? { get set }
   func updateAll()
   func viewControllerWillAppear()
+  func viewControllerCancelled()
```
deepesh-vasthimal-cko, line 26: The name of the method viewControllerCancelled and what it does not match. Should we rename this to `logPaymentFormDismissed` ?

deepesh-vasthimal-cko (approved):
just one comment on naming, Otherwise looks good. Amazing turn around in getting all the PR's for logging and clearing out the correlation ids.

Source/UI/PaymentForm/ViewModel/PaymentViewModel.swift
alex-ioja-yang-cko: I wanted to name dismiss, but looking at the code, the screen is also dismissed on completion of the journey (tokenisation completed). That's what was originally planned but can be misguiding when seen out of codebase?
Next best thing was cancelled, as the event gets triggered on user pressing "back arrow" from navigation bar, which is the closest thing to a cancel that the journey offers.

deepesh-vasthimal-cko (approved):

PR #320, opened 2022-10-07
PIMOB-1362(6): Create & Use paymentFormSubmitted logging event

alex-ioja-yang-cko:
## Proposed changes

As [raised in previous PR from chain](https://github.com/checkout/frames-ios/pull/315#discussion_r989928317), pressing the Done button if card details cannot be generated (for now this means missing any number input or expiry date) would result in a silent fail.

Changes proposed log this failure as its something we should not expect to happen, as the pay button getting enabled has more stringent conditions than the card object creation. If it does occur it would be of interest to fix & review processes into how it appeared.

### DIFF
--- a/Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
+++ b/Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
@@ -164,7 +164,10 @@ extension DefaultPaymentViewModel: PaymentViewControllerDelegate {
     }
 
     func payButtonIsPressed() {
-        guard let card = cardDetails.getCard() else { return }
+        guard let card = cardDetails.getCard() else {
+            logger.log(.warn(message: "Pay button pressed without all required fields input"))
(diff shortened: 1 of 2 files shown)


Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
```diff
@@ -164,7 +164,10 @@ extension DefaultPaymentViewModel: PaymentViewControllerDelegate {
     }
 
     func payButtonIsPressed() {
-        guard let card = cardDetails.getCard() else { return }
+        guard let card = cardDetails.getCard() else {
+            logger.log(.warn(message: "Pay button pressed without all required fields input"))
```
deepesh-vasthimal-cko, line 168: 👍

deepesh-vasthimal-cko (approved):

deepesh-vasthimal-cko (approved):

PR #324, opened 2022-10-17
PIMOB-XXXX: Improve automatic scroll on keyboard appearance

alex-ioja-yang-cko:
## Proposed changes

Replace usage of [keyboardFrameBeginUserInfoKey](https://developer.apple.com/documentation/uikit/uiresponder/1621616-keyboardframebeginuserinfokey) for the estimate of keyboard height with [keyboardFrameEndUserInfoKey](https://developer.apple.com/documentation/uikit/uiresponder/1621578-keyboardframeenduserinfokey).

This is an intended correction, as per their own documentations:
- `keyboardFrameEndUserInfoKey`: A user info key to retrieve the keyboard’s frame **at the end of its animation**
- `keyboardFrameBeginUserInfoKey`: A user info key to retrieve the keyboard’s frame **at the beginning of its animation**

As we need the size to calculate offset needed, the final size would be the most accurate value to use.

### DIFF
--- a/Source/UI/PaymentForm/ViewController/PaymentViewController.swift
+++ b/Source/UI/PaymentForm/ViewController/PaymentViewController.swift
@@ -143,7 +143,7 @@ final class PaymentViewController: UIViewController {
 
   @objc private func keyboardWillShow(notification: Notification) {
     guard let userInfo = notification.userInfo,
-      let keyboardFrameValue = userInfo[UIResponder.keyboardFrameBeginUserInfoKey] as? NSValue else { return }
+    let keyboardFrameValue = userInfo[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue else { return }
     var keyboardFrame = keyboardFrameValue.cgRectValue
     keyboardFrame = view.convert(keyboardFrame, from: nil)
     var contentInset: UIEdgeInsets = scrollView.contentInset


Source/UI/PaymentForm/ViewController/PaymentViewController.swift
```diff
@@ -143,7 +143,7 @@ final class PaymentViewController: UIViewController {
 
   @objc private func keyboardWillShow(notification: Notification) {
     guard let userInfo = notification.userInfo,
-      let keyboardFrameValue = userInfo[UIResponder.keyboardFrameBeginUserInfoKey] as? NSValue else { return }
+    let keyboardFrameValue = userInfo[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue else { return }
```
deepesh-vasthimal-cko, line 146: Screenshot 2022-10-17 at 09 10 48
Assume if fixes the above ?

deepesh-vasthimal-cko (approved):

Source/UI/PaymentForm/ViewController/PaymentViewController.swift
alex-ioja-yang-cko: Yes, that's the one. I'm not sure why, but sometimes the start frame height is 100 pixels shorter than end height. I'm surprised the start was not 0 since you'd expect the animation to start from nothing 🤷‍♂️

PR #328, opened 2022-10-20
PostRelease: Fix podspec 

alex-ioja-yang-cko:
## Proposed changes

The pod specs were not linting correctly for releases. This is due to structure changes and mostly misplaced unit tests.

As we are using SPM for our core development & pipeline tests, the tests are already being ran and validated so it doesn't make sense to worry about distributing the pods with unit tests?

I took the easy way and removed them from the podspec.

### DIFF
--- a/iOS Example Frame/Podfile.lock
+++ b/iOS Example Frame/Podfile.lock
@@ -1,13 +1,9 @@
 PODS:
-  - Checkout (3.5.3):
+  - Checkout (4.0.0):
     - CheckoutEventLoggerKit (= 1.2.0)
   - CheckoutEventLoggerKit (1.2.0)
-  - Frames (3.5.3):
-    - Checkout (= 3.5.3)
-    - CheckoutEventLoggerKit (= 1.2.0)
-    - PhoneNumberKit (= 3.3.3)
-  - Frames/Tests (3.5.3):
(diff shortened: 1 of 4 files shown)


iOS Example Frame/Podfile.lock
```diff
@@ -1,13 +1,9 @@
...
+  - Checkout (4.0.0):
     - CheckoutEventLoggerKit (= 1.2.0)
   - CheckoutEventLoggerKit (1.2.0)
-  - Frames (3.5.3):
-    - Checkout (= 3.5.3)
-    - CheckoutEventLoggerKit (= 1.2.0)
-    - PhoneNumberKit (= 3.3.3)
-  - Frames/Tests (3.5.3):
-    - Checkout (= 3.5.3)
+  - Frames (4.0.0):
```
deepesh-vasthimal-cko, line 5: Question if this change was never merged how was i able to run the pod install earlier today on main branch and get latest 4.0.0 ?. Am I missing something. 
Screenshot 2022-10-20 at 12 43 33

deepesh-vasthimal-cko: I noticed your are updating the iOS example frame so makes sense now.

deepesh-vasthimal-cko (approved):

PR #332, opened 2022-10-21
PIMOB-XXXX: Remove overhead from ThreeDSWebViewController initialiser

alex-ioja-yang-cko:
The VC was requesting a network client as part of the initialisation, but only consuming the logging part of the network client, discarding the rest.

Changed initialiser to only request an environment, and use that to create logger directly, reducing complexity and unnecessary overhead.

## Notice
This will require update of ReadMe & Public Docs, but that may be done as part of the next release, not this PR (since latest release is in sync with the current docs & change would not be available via the actual release)

### DIFF
--- a/Source/UI/3DS/ThreedsWebViewController.swift
+++ b/Source/UI/3DS/ThreedsWebViewController.swift
@@ -29,24 +29,11 @@ public class ThreedsWebViewController: UIViewController {
     // MARK: - Initialization
 
     /// Initializes a web view controller adapted to handle 3dsecure.
-    public convenience init(checkoutAPIService: CheckoutAPIService, successUrl: URL, failUrl: URL) {
-        self.init(
-            checkoutAPIProtocol: checkoutAPIService,
-            successUrl: successUrl,
-            failUrl: failUrl,
-            threeDSWKNavigationHelperFactory: ThreeDSWKNavigationHelperFactory()
-        )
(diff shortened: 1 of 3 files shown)


ehab-al-cko (approved):

Source/UI/3DS/ThreedsWebViewController.swift
```diff
@@ -29,24 +29,11 @@ public class ThreedsWebViewController: UIViewController {
...
-    public convenience init(checkoutAPIService: CheckoutAPIService, successUrl: URL, failUrl: URL) {
-        self.init(
-            checkoutAPIProtocol: checkoutAPIService,
-            successUrl: successUrl,
-            failUrl: failUrl,
-            threeDSWKNavigationHelperFactory: ThreeDSWKNavigationHelperFactory()
-        )
-    }
+    public convenience init(environment: Environment, successUrl: URL, failUrl: URL) {
+        let logger = FramesEventLogger(environment: environment, correlationID: UUID().uuidString)
```
deepesh-vasthimal-cko, line 33: Injecting a new correlation ID which means the ThreedsWebViewController and Token events will have different cor ids ?

alex-ioja-yang-cko: They didn't have the same correlation ID to begin with.

PR #337, opened 2022-10-27
PIMOB-1480:  Implementing editable borders for one component

ehab-al-cko:
[Jira ticket ](https://checkout.atlassian.net/browse/PIMOB-1480)

**Screenshot**

## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/Extensions/UIViewExtensions.swift
+++ b/Source/Extensions/UIViewExtensions.swift
@@ -68,3 +68,72 @@ extension UIView {
   }
 
 }
+
+extension UIView {
+
+    func addBorder(with list: [SideBorder], thickness: Double, color: UIColor) {
+        list.forEach {
+            addBorder(side: $0, thickness: thickness, color: color)
+        }
+    }
+
+    private func addBorder(side: SideBorder, thickness: Double, color: UIColor) {
+        var thickness = thickness
+        var edgeView = UIView().disabledAutoresizingIntoConstraints()
+                edgeView.backgroundColor = color
+
+        switch side {
+            case .left(let edge):
+                if let edge = edge {
+                    edgeView = edge
+                    thickness = edgeView.frame.width
+                }
+                addSubview(edgeView)
+                NSLayoutConstraint.activate([
+                    edgeView.topAnchor.constraint(equalTo: topAnchor),
+                    edgeView.leftAnchor.constraint(equalTo: leftAnchor),
+                    edgeView.bottomAnchor.constraint(equalTo: bottomAnchor),
+                    edgeView.widthAnchor.constraint(equalToConstant: thickness)
+                ])
+
+            case .right(let edge):
+                if let edge = edge {
+                    edgeView = edge
+                    thickness = edgeView.frame.width
+                }
+                addSubview(edgeView)
+                NSLayoutConstraint.activate([
+                    edgeView.topAnchor.constraint(equalTo: topAnchor),
+                    edgeView.rightAnchor.constraint(equalTo: rightAnchor),
+                    edgeView.bottomAnchor.constraint(equalTo: bottomAnchor),
+                    edgeView.widthAnchor.constraint(equalToConstant: thickness)
+                ])
+            case .top(let edge):
+                if let edge = edge {
+                    edgeView = edge
+                    thickness = edgeView.frame.height
+                }
+                addSubview(edgeView)
+                NSLayoutConstraint.activate([
+                    edgeView.topAnchor.constraint(equalTo: topAnchor),
+                    edgeView.rightAnchor.constraint(equalTo: rightAnchor),
+                    edgeView.leftAnchor.constraint(equalTo: leftAnchor),
+                    edgeView.heightAnchor.constraint(equalToConstant: thickness)
+                ])
+
+            case .bottom(let edge):
+                if let edge = edge {
+                    edgeView = edge
+                    thickness = edgeView.frame.height
+                }
+                addSubview(edgeView)
+                NSLayoutConstraint.activate([
+                    edgeView.bottomAnchor.constraint(equalTo: bottomAnchor),
+                    edgeView.rightAnchor.constraint(equalTo: rightAnchor),
+                    edgeView.leftAnchor.constraint(equalTo: leftAnchor),
+                    edgeView.heightAnchor.constraint(equalToConstant: thickness)
+                ])
+        }
+
+    }
+}
--- a/Source/UI/BillingForm/Default Implementation/Elements/TextField/DefaultTextField.swift
+++ b/Source/UI/BillingForm/Default Implementation/Elements/TextField/DefaultTextField.swift
@@ -1,6 +1,7 @@
 import UIKit
 
 public struct DefaultTextField: ElementTextFieldStyle {
+    public var sideBorders: [SideBorder]? = [.bottom(layer: nil)]
     public var textAlignment: NSTextAlignment = .natural
     public var isHidden = false
     public var isSupportingNumericKeyboard = false
--- a/Source/UI/BillingForm/View/BillingFormTextFieldView.swift
+++ b/Source/UI/BillingForm/View/BillingFormTextFieldView.swift
@@ -78,9 +78,9 @@ class BillingFormTextFieldView: UIView {
         mandatoryLabel.update(with: style.mandatory)
         hintLabel.update(with: style.hint)
 
-        updateTextFieldContainer(style: style)
         updateTextField(style: style, textFieldValue: textFieldValue, tag: tag)
         updateErrorView(style: style)
+        updateTextFieldContainer(style: style)
     }
 
     func refreshLayoutComponents() {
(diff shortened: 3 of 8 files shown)


Source/Extensions/UIViewExtensions.swift
```diff
@@ -68,3 +68,72 @@ extension UIView {
...
+        var edgeView = UIView().disabledAutoresizingIntoConstraints()
+                edgeView.backgroundColor = color
+
+        switch side {
+            case .left(let edge):
+                if let edge = edge {
+                    edgeView = edge
+                    thickness = edgeView.frame.width
+                }
+                addSubview(edgeView)
```
ehab-al-cko, line 91: This will add multi views every time updating the colors.

Source/Extensions/UIViewExtensions.swift
```diff
@@ -68,3 +68,72 @@ extension UIView {
...
+                    edgeView.bottomAnchor.constraint(equalTo: bottomAnchor),
+                    edgeView.widthAnchor.constraint(equalToConstant: thickness)
+                ])
+
+            case .right(let edge):
+                if let edge = edge {
+                    edgeView = edge
+                    thickness = edgeView.frame.width
+                }
+                addSubview(edgeView)
```
ehab-al-cko, line 104: same.

Source/UI/CommonUI/Protocols/Elements/ElementTextFieldStyle.swift
```diff
@@ -10,4 +10,5 @@ public protocol ElementTextFieldStyle: ElementStyle {
     var normalBorderColor: UIColor { get }
     var focusBorderColor: UIColor { get }
     var errorBorderColor: UIColor { get }
+    var sideBorders: [SideBorder]? { get }
```
alex-ioja-yang-cko, line 13: I think at this stage we can group all the border properties into a new object.

```swift
public struct BorderStyle {
    public var cornerRadius: CGFloat
    public var borderWidth: CGFloat
    public var normalBorderColor: UIColor        
    public var focusBorderColor: UIColor
    public var errorBorderColor: UIColor
}
```

This should help improve discoverability and fluid navigation of the properties within.

So after ☝️ we would not have
```swift
public protocol ElementTextFieldStyle: ElementStyle {
    var isSupportingNumericKeyboard: Bool { get set }
    var height: Double { get }
    var borderStyle: BorderStyle { get }
    var placeholder: String? { get }
    var tintColor: UIColor { get }
}
```

Source/Extensions/UIViewExtensions.swift
```diff
@@ -68,3 +68,72 @@ extension UIView {
   }
 
 }
+
+extension UIView {
+
+    func addBorder(with list: [SideBorder], thickness: Double, color: UIColor) {
+        list.forEach {
+            addBorder(side: $0, thickness: thickness, color: color)
```
alex-ioja-yang-cko, line 76: Code level I'm not sure this approach will work if you have right and bottom border with corner radius.

Source/Extensions/UIViewExtensions.swift
```diff
@@ -68,3 +68,72 @@ extension UIView {
...
+                if let edge = edge {
+                    edgeView = edge
+                    thickness = edgeView.frame.width
+                }
+                addSubview(edgeView)
+                NSLayoutConstraint.activate([
+                    edgeView.topAnchor.constraint(equalTo: topAnchor),
+                    edgeView.leftAnchor.constraint(equalTo: leftAnchor),
+                    edgeView.bottomAnchor.constraint(equalTo: bottomAnchor),
+                    edgeView.widthAnchor.constraint(equalToConstant: thickness)
```
alex-ioja-yang-cko, line 96: Just curious, but why left anchor and width, instead of left anchor and right anchor?

Source/UI/CommonUI/Protocols/Elements/SideBorder.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+
+import UIKit
+
+public enum SideBorder: CaseIterable {
+    public static var allCases: [SideBorder] = [.left(layer: nil), .right(layer: nil), .top(layer: nil), .bottom(layer: nil)]
+    case left(layer: UIView?)
+    case right(layer: UIView?)
+    case top(layer: UIView?)
+    case bottom(layer: UIView?)
```
alex-ioja-yang-cko, line 16: We can make use of [Default enum associated values](https://www.swiftbysundell.com/tips/default-enum-associated-values/) which has been available since Swift 5.3 (so should have no impact on our minimum Swift 5.4 requirement).

Why UIView and not CALayer as originally proposed?

alex-ioja-yang-cko (commented):
Overall looks good but I think we need to review a few aspects (UIView vs CALayer, corner radius, edge cases) on Slack.

Source/UI/BillingForm/Default Implementation/Elements/TextField/DefaultTextField.swift
```diff
@@ -1,21 +1,25 @@
...
     public var tintColor: UIColor = FramesUIStyle.Color.textPrimary
-    public var borderWidth: CGFloat = 1.0
-    public var cornerRadius: CGFloat = 10.0
     public var width: Double = Constants.Style.BillingForm.InputTextField.width.rawValue
     public var height: Double = Constants.Style.BillingForm.InputTextField.height.rawValue
     public var font = FramesUIStyle.Font.inputLabel
+    public var borderStyle: ElementBorderStyle = DefaultBorderStyle()
+}
+
+public struct DefaultBorderStyle: ElementBorderStyle {
```
alex-ioja-yang-cko, line 18: This will need its own file.

Source/Extensions/UIViewExtensions.swift
```diff
@@ -71,68 +71,52 @@ extension UIView {
 
 extension UIView {
 
-    func addBorder(with list: [SideBorder], thickness: Double, color: UIColor) {
+    func addBorder(with list: [UIRectEdge], thickness: Double, color: UIColor) {
```
alex-ioja-yang-cko, line 74: UIRectEdge is an `OptionSet`, which means you don't need to send it as an array

```suggestion
    func addBorder(with edges: UIRectEdge, thickness: Double, color: UIColor) {
```

You can try this in Xcode by pasting this code and seeing it compile :)
```swift
let edge1: UIRectEdge = [.left, .right]
let edge2: [UIRectEdge] = [.left, .right]
```

Furthermore, the fact this is an option set it will drastically impact our approach to add the borders. The whole approach should probably be changed to something like:

```swift
extension UIView {

  func applyBorder(_ style: ElementBorderStyle) {
    if style.edge.contains(.all) && style.corners.contains(.all) {
      // Apply standard border using relevant properties from provided style
      return
    }

    if style.edge.contains(.left) {
      // Add left edge line
    }

    // Perform this if check for each side and corner
  }

}
```

Source/UI/BillingForm/Default Implementation/Elements/TextField/DefaultTextField.swift
```diff
@@ -1,21 +1,25 @@
...
+    public var cornerRadius: CGFloat = 10
+    public var borderWidth: CGFloat = 1.0
+    public var normalBorderColor: UIColor = FramesUIStyle.Color.borderPrimary
+    public var focusBorderColor: UIColor = FramesUIStyle.Color.borderActive
+    public var errorBorderColor: UIColor = FramesUIStyle.Color.borderError
+    public var sideBorders: [UIRectEdge] = [.all]
```
alex-ioja-yang-cko, line 24: Parent object is not `BorderStyle`, so would remove the references from property names:

```swift
public var cornerRadius: CGFloat = 10
public var width: CGFloat = 1.0
public var normalColor: UIColor = FramesUIStyle.Color.borderPrimary
public var focusColor: UIColor = FramesUIStyle.Color.borderActive
public var errorColor: UIColor = FramesUIStyle.Color.borderError
public var edges: UIRectEdge = .all

// Include as well
public var corners: UIRectCorner = .all
```

Source/UI/BillingForm/View/BillingFormTextFieldView.swift
```diff
@@ -80,6 +80,10 @@ class BillingFormTextFieldView: UIView {
 
         updateTextField(style: style, textFieldValue: textFieldValue, tag: tag)
         updateErrorView(style: style)
+        layoutIfNeeded()
+        layoutSubviews()
+        textFieldContainer.layoutIfNeeded()
+        textFieldContainer.layoutSubviews()
```
alex-ioja-yang-cko, line 86: I wonder if we really need this code on update. 

The objects lifecycle may invoke on update, so would be curious if we can instead just use object's own lifecycle updates:

```swift
override func layoutSubviews() {
    super.layoutSubviews()

    textFieldContainer.layoutIfNeeded()
    textFieldContainer.layoutSubviews()
}
```

alex-ioja-yang-cko (commented):
I think we can optimise on the API consumption side and enable border customisation as well (we actually increase customisability quite a bit with this, whilst saving us from having to detect corners ourselves).

The change to border approach (grouping properties into object) will cause a breaking change for some, so @deepesh-vasthimal-cko this may require a follow up release of 4.1.0. imo its a big improvement and worth it but need to be clear that the UI declaration will break (for custom, theme & default if overriding border).

Source/Extensions/UIViewExtensions.swift
```diff
@@ -68,3 +68,56 @@ extension UIView {
   }
 
 }
+
```
alex-ioja-yang-cko, line 71: Another idea that actually should be best for performance & maintainability (both in code & in sync on UI) would be to try `UIBezierPath`. 

It can draw completely custom paths. Using `move`, `addLine`, `addCurve` we should be able to completely custom draw all edges & corners, inside a system of 1 x 1. 

We then assign this path to a CAShapeLayer that we assign to the view, also keeping the reference inside the object. This will then allow us to scale it with the view as needed:
```swift
override func layoutSubviews {
    super.layoutSubviews()

    customBorderLayer.frame = bounds;
}
```

alex-ioja-yang-cko (changes requested):
Please lets talk this ASAP as a way to get outcome quickly

PR #338, opened 2022-11-01
PIMOB-1479: Update Phone number validation logic

ehab-al-cko:
[Jira ticket ](https://checkout.atlassian.net/browse/PIMOB-1479)

## Proposed changes

## Types of changes

## Checklist

## Further comments


Source/UI/Validator/PhoneNumberValidator.swift
```diff
@@ -1,12 +1,13 @@
...
-    }
+    private let maxDigitLength = 25
+    private let minDigitLength = 6
 
-    private func isEmpty(text: String?) -> Bool {
-        text?.isEmpty ?? false
+    func validate(text: String?) -> Bool {
+        guard let text = text else { return true }
+        let number = text.components(separatedBy: CharacterSet.decimalDigits.inverted).joined()
+        return number.isEmpty || number.count < 6 || number.count > 25
```
deepesh-vasthimal-cko, line 10: If I read this right, validate will return true if `number is empty` or `number is less than 6` characters or` more than 25 characters`. However we should be returning false when conditions are not met and true for valid values ?

Tests/UI/BillingForm/Validator/PhoneNumberValidatorTests.swift
```diff
@@ -3,17 +3,38 @@ import XCTest
...
 class PhoneNumberValidatorTests: XCTestCase {
 
-    func testValidationWhenTextIsEmptyThenShouldBeTrue() {
+    func testValidationTextWithEmptyString() {
         let expectedType = BillingFormCell.phoneNumber(nil)
         let text = ""
-        let isValid = expectedType.validator.validate(text: text)
-        XCTAssertTrue(isValid)
+        let isInvalid = expectedType.validator.validate(text: text)
+        XCTAssertTrue(isInvalid)
```
deepesh-vasthimal-cko, line 10: In relation to my comment : https://github.com/checkout/frames-ios/pull/338/files#r1010412180. Here the text is empty string so the expectedType.validator.validate should return false Instead of returning true

deepesh-vasthimal-cko (changes requested):
see return value comments on validate function.

Source/UI/Validator/PhoneNumberValidator.swift
deepesh-vasthimal-cko, line 10: you can also leave a comment here refer : https://api-reference.checkout.com/#operation/requestAToken!path=0/phone&t=request

Source/UI/Validator/PhoneNumberValidator.swift
```diff
@@ -1,12 +1,13 @@
...
-    func validate(text: String?) -> Bool {
-        isEmpty(text: text)
-    }
+    private let maxDigitLength = 25
+    private let minDigitLength = 6
 
-    private func isEmpty(text: String?) -> Bool {
-        text?.isEmpty ?? false
+    func validate(text: String?) -> Bool {
+        guard let text = text else { return true }
```
deepesh-vasthimal-cko, line 8: if text is nil we should return false ?, then it would mean if validate returns true for a good data and validate returns false for wrong data. @alex-ioja-yang-cko @ehab-al-cko suggestions ?

ehab-al-cko: Fixed in new commit. naming was confusing.

Checkout/Checkout/Source/Model/Country.swift
```diff
@@ -277,6 +277,15 @@ public struct Country: Equatable {
     ]
   }
 
+    public init?(dialingCode: String) {
+        var dialingCode = dialingCode.components(separatedBy: .decimalDigits.inverted).joined()
+        if let country = Self.allAvailable.first(where: { $0.dialingCode == dialingCode }) {
+            self = country
+        } else {
+            return nil
+        }
```
alex-ioja-yang-cko, line 286:
```suggestion
        return Self.allAvailable.first(where: { $0.dialingCode == dialingCode })
```

Checkout/Checkout/Source/Model/Country.swift
```diff
@@ -277,6 +277,15 @@ public struct Country: Equatable {
     ]
   }
 
+    public init?(dialingCode: String) {
+        var dialingCode = dialingCode.components(separatedBy: .decimalDigits.inverted).joined()
```
alex-ioja-yang-cko, line 281: Why is this needed?

CheckoutTests/Network/RequestFactoryTests.swift
```diff
@@ -58,7 +58,7 @@ final class RequestFactoryTests: XCTestCase {
       XCTAssertEqual(requestParameters.url, URL(string: "https://www.checkout.com/tokens"))
       XCTAssertEqual(requestParameters.additionalHeaders, [
         "Authorization": "Bearer publicKey",
-        "User-Agent": "checkout-sdk-ios/0.1.0"
+        "User-Agent": "checkout-sdk-ios/4.0.0"
```
alex-ioja-yang-cko, line 61: As a rule of thumb I wouldn't change things that don't need changing. This is a test framework, having a version that clearly stands out may come in handy.

Source/Extensions/StringExtensions.swift
```diff
@@ -13,6 +13,10 @@ extension String {
     func standardize() -> String {
         return self.replacingOccurrences(of: "[^0-9]", with: "", options: .regularExpression)
     }
+
+    func decimalDigits() -> String {
+        components(separatedBy: .decimalDigits.inverted).joined()
+    }
```
alex-ioja-yang-cko, line 19: I am unsure what this is doing. Some unit tests would also be nice to clarify why it is needed.

Source/UI/BillingForm/View/BillingFormTextFieldView.swift
```diff
@@ -18,6 +19,20 @@ class BillingFormTextFieldView: UIView {
...
+        return view
+    }()
+
+    private lazy var headerStackView: UIStackView = {
+        let view = UIStackView().disabledAutoresizingIntoConstraints()
+        view.axis = .vertical
+        view.spacing = 6
+        return view
+    }()
+
```
alex-ioja-yang-cko, line 35: Given we're trying to add validation to the phone number, I am unsure why this is needed? I would expect new UI components if we were focussed on changing UI, but that is not what I thought we're doing?

Source/UI/Validator/PhoneNumberValidator.swift
```diff
@@ -1,12 +1,14 @@
 import Foundation
 
 class PhoneNumberValidator: Validator {
-    func validate(text: String?) -> Bool {
-        isEmpty(text: text)
-    }
+    //  https://api-reference.checkout.com/#operation/requestAToken!path=0/phone&t=request
+    private let maxDigitLength = 25
+    private let minDigitLength = 6
```
alex-ioja-yang-cko, line 6: I don't think we should just drop urls in code. If it really is needed, we may at least drop some context to the url.

Again, my impression was the source of truth for validator was supposed to be inside the Checkout module. Why are we adding these constraints here instead of using a single validator?

alex-ioja-yang-cko (changes requested):
Based on conversations we had on this task, I don't think we should have
❌  UI layout changes
❌  Frames module Phone validator logic

Checkout/Checkout/Source/Model/Country.swift
ehab-al-cko: I tried and i got this error message "'nil' is the only return value permitted in an initializer"

Checkout/Checkout/Source/Model/Country.swift
ehab-al-cko: For country code validation, we need to know `iso3166Alpha2` using `dialingCode`

Source/Extensions/StringExtensions.swift
ehab-al-cko: It's returning only digits without special characters. It's Native Apple API , no need for unit test

Source/UI/BillingForm/View/BillingFormTextFieldView.swift
ehab-al-cko: Every time there's an error for phone number, Sometimes, UI was not showing error view and sometime hide hint label.   Old UI was depending on constraints, I think, Stackviews is better for this solution.

Source/UI/Validator/PhoneNumberValidator.swift
ehab-al-cko: This was a comment from Deepesh to add it. I agree with Deeepsh, URL is a good reference for it.

Source/UI/BillingForm/ViewController/BillingFormViewController.swift
```diff
@@ -130,7 +131,7 @@ final class BillingFormViewController: UIViewController {
     private func refreshCell(at row: Int?) {
         guard let row = row else { return }
         let indexPath = IndexPath(row: row, section: 0)
-        tableView?.reloadRows(at: [indexPath], with: .automatic)
+        tableView?.reloadRows(at: [indexPath], with: .none)
```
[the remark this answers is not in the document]
ehab-al-cko: Animation was causing problems when multi fields has errors.

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -167,6 +169,19 @@ final class DefaultBillingFormViewModel: BillingFormViewModel {
...
+        guard let text = text, !text.isEmpty else { return true }
+        let phoneMaxLength = Checkout.Constants.Phone.phoneMaxLength
+        let region: String? = text.first == "+" ? nil : data?.phone?.country?.iso3166Alpha2
+        let phoneNumberKit = PhoneNumberKit()
+        let phoneNumber = try? phoneNumberKit.parse(text, withRegion: region ?? PhoneNumberKit.defaultRegionCode(), ignoreType: true)
+        guard let phoneNumber = phoneNumber else {
+            let decimalDigitsLength = text.decimalDigits().count
+            return decimalDigitsLength <= phoneMaxLength
+        }
+        return String(phoneNumber.nationalNumber).count <= phoneMaxLength
```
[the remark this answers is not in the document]
ehab-al-cko: Conversation was to use validator from Checkout after user finish editing. This code is to limit user to add more than max length.

ehab-al-cko:
> Based on conversations we had on this task, I don't think we should have
> ❌ UI layout changes
> ❌ Frames module Phone validator logic

- UI was mandatory to solve UI Error view issue which is related to phone validator. 
- Frames validator is doing some validation on UI and business logic

Source/Extensions/StringExtensions.swift
alex-ioja-yang-cko: `decimalDigits` is not Apple API though, its our own method that uses 2 Apple APIs together to create our own transformation.

Also why the name _decimal digits_ and not just _digits_?

Source/UI/Validator/Validator.swift
```diff
@@ -1,5 +1,5 @@
 import Foundation
 
 protocol Validator {
-    func validate(text: String?) -> Bool
+    func isInvalid(value: Any?) -> Bool
```
alex-ioja-yang-cko, line 4: Not sure this is an improvement.
- we should avoid negation (is invalid vs is valid)
- why use Any? The validators are probably looking at Strings, why introduce ambiguity?

Checkout/Checkout/Source/Validation/Constants.swift
```diff
@@ -7,26 +7,26 @@
...
-  enum Product {
-    static let version = "4.0.0"
-    static let name = "checkout-ios-sdk"
-    static let userAgent = "checkout-sdk-ios/\(version)"
-  }
+    public enum Product {
+        public static let version = "4.0.0"
+        public static let name = "checkout-ios-sdk"
+        public static let userAgent = "checkout-sdk-ios/\(version)"
+    }
```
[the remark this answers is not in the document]
deepesh-vasthimal-cko: Same question I don't see them being used elsewhere in PR.

Tests/Mocks/BillingFormTextFieldCellMockDelegate.swift
```diff
@@ -13,13 +13,17 @@ class BillingFormTextFieldCellMockDelegate: CellTextFieldDelegate {
     var textFieldDidChangeCharactersLastCalledWithReplacementString: String?
 
     var phoneNumberIsUpdatedCalledTimes = 0
-    var phoneNumberIsUpdatedLastCalledWithNumber: String?
+    var phoneNumberIsUpdatedLastCalledWithNumber: Frames.Phone?
```
deepesh-vasthimal-cko, line 16: Frames.Phone what is this referring to ? Is this a type alias

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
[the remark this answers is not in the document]
alex-ioja-yang-cko: Did you try that? From their ReadMe:

> The region code is automatically computed 

this is ambiguous now from our viewpoint, but I am still tempted to think computed = computed from the string being parsed?

Screenshot 2022-11-08 at 10 14 01

Source/UI/Validator/Validator.swift
alex-ioja-yang-cko: I don't agree with Any. We should avoid ambiguity as much as we can, and there's not much more ambiguity then `Any`.

`isInvalid` is not clear because it negates a boolean. 

is invalid = is NOT valid
```swift
if isInvalid 

vs 

if isValid
```
easier to mess up logical boolean comparisons when the variable is already inverted.

Source/UI/BillingForm/View/BillingFormPhoneNumberText.swift
```diff
@@ -28,7 +27,17 @@ final class BillingFormPhoneNumberText: PhoneNumberTextField, BillingFormTextFie
...
         super.textFieldDidEndEditing(textField, reason: reason)
-        guard let phoneNumber = text else { return }
-        phoneNumberTextDelegate?.phoneNumberIsUpdated(number: phoneNumber, tag: tag)
+        let country = Country(iso3166Alpha2: partialFormatter.currentRegion)
+        let phone = Phone(number: textField.text, country: country)
+        phoneNumberTextDelegate?.phoneNumberIsUpdated(number: phone, tag: tag)
     }
+
+    override func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+        guard string.isEmpty || string == "+" || Int(string) != nil else { return false }
```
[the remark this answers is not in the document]
alex-ioja-yang-cko: An unit test would confirm whether my observation is correct or not. But you make 3 assumptions to proceed in applying logic:
- string is empty `string = ""`
- string is equal to a plus `string = "+"
- string is a number `string = "077"

So now I wonder what happens if input is "+077".

CheckoutTests/Model/CountryTests.swift
```diff
@@ -268,4 +268,16 @@ final class CountryTests: XCTestCase {
     XCTAssertEqual(Country(iso3166Alpha2: "UZ"), Country(iso3166Alpha2: "UZ", dialingCode: "998"))
     XCTAssertEqual(Country(iso3166Alpha2: "WF"), Country(iso3166Alpha2: "WF", dialingCode: "681"))
   }
+
+    func testValidInitWithDialingCode() {
+        let dialingCode = "+44"
+        let country = Country(dialingCode: dialingCode)
+        XCTAssertEqual(country?.iso3166Alpha2, "GG")
```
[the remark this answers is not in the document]
alex-ioja-yang-cko: The point of a unit test is to find errors in code. If +44 results in a GG country code, do you think there is a bug that you spotted and we should fix?

Checkout/Checkout/Source/Validation/Constants.swift
```diff
@@ -7,22 +7,23 @@
...
+        public static let zipLength = 50
+        public static let countryLength = 2
+    }
+
+    public enum Phone {
+        public static let phoneMinLength = 6
+        public static let phoneMaxLength = 25
+        public static let countryCodeMinLength = 1
+        public static let countryCodeMaxLength = 7
+    }
```
alex-ioja-yang-cko, line 25: Do we need to make all of these properties public? Tried looking for some and they don't appear to be used?

Checkout/Checkout/Source/Validation/Validators/PhoneValidator.swift
```diff
@@ -7,24 +7,26 @@
...
-final class PhoneValidator: PhoneValidating {
-  func validate(_ phone: Phone) -> ValidationResult<ValidationError.Phone> {
-    if let number = phone.number,
-      number.count < Constants.Phone.phoneMinLength ||
-      number.count > Constants.Phone.phoneMaxLength {
-      return .failure(.numberIncorrectLength)
+public final class PhoneValidator: PhoneValidating {
+    public init() {}
+    
+    public func validate(_ phone: Phone) -> ValidationResult<ValidationError.Phone> {
```
alex-ioja-yang-cko, line 17:
```suggestion
public final class PhoneValidator: PhoneValidating {
    
    static public func validate(_ phone: Phone) -> ValidationResult<ValidationError.Phone> {
```

Source/UI/BillingForm/View/BillingFormPhoneNumberText.swift
alex-ioja-yang-cko, line 36: unit test?

Tests/UI/BillingForm/Validator/AddressLine2ValidatorTests.swift
```diff
@@ -6,14 +6,14 @@ class AddressLine2ValidatorTests: XCTestCase {
...
+        XCTAssertTrue(isInvalid)
     }
 
     func testValidationWhenTextIsNonEmptyThenShouldBeFalse() {
         let expectedType = BillingFormCell.addressLine2(nil)
         let text = "addressLine2"
-        let isValid = expectedType.validator.validate(text: text)
-        XCTAssertFalse(isValid)
+        let isInvalid = expectedType.validator.isInvalid(value: text)
+        XCTAssertFalse(isInvalid)
```
alex-ioja-yang-cko, line 17: This is an example of why I was saying your naming of `isInvalid` should not be negative. We have been using `isValid` before and that was more readable.

Now Assert isInvalid is False. See the double negation? The tests now read completely opposite which makes me wonder if they're still even working as expected?

Source/UI/Validator/AddressLine1Validator.swift
```diff
@@ -1,8 +1,8 @@
 import Foundation
 
 class AddressLine1Validator: Validator {
-    func validate(text: String?) -> Bool {
-        isEmpty(text: text)
+    func isInvalid(value: Any?) -> Bool {
+        isEmpty(text: value as? String)
```
alex-ioja-yang-cko, line 5: I cannot get behind this naming. It reads a lot cleaner `isValid`.

Additionally, we cannot have Any as parameter, it just strips away any type safety.  A different approach is needed.

alex-ioja-yang-cko (commented):
I cannot get behind the `isInvalid` and `Any` parameter. They need to be improved :/ .

Source/UI/BillingForm/ViewModel/DefaultBillingFormViewModel.swift
```diff
@@ -231,12 +245,22 @@ extension DefaultBillingFormViewModel: BillingFormTextFieldDelegate {
...
+    }
+
+    func phoneNumberIsUpdated(number: Phone, tag: Int) {
+        let index = BillingFormCell.phoneNumber(nil).index
+        textValueOfCellType[index] = number.number
+        let currentCellTypeIndex = style.cells[tag].index
+        errorFlagOfCellType[currentCellTypeIndex] = !PhoneNumberValidator().validate(value: number)
+        notifyContentChangeToDelegate()
+    }
+
```
alex-ioja-yang-cko, line 263: Why was the `textFieldDidEndEditing` introduced? It seems previously this was covered within `phoneNumberIsUpdated`?

Source/UI/Validator/PhoneNumberValidator.swift
```diff
@@ -1,12 +1,11 @@
...
-    func validate(text: String?) -> Bool {
-        isEmpty(text: text)
-    }
+    //  https://api-reference.checkout.com/#operation/requestAToken!path=0/phone&t=request
 
-    private func isEmpty(text: String?) -> Bool {
-        text?.isEmpty ?? false
+    func validate(value: Any?) -> Bool {
+        guard let phoneNumber = (value as? Phone) else { return false }
+        return PhoneValidator().validate(phoneNumber) == .success
```
alex-ioja-yang-cko, line 8: The url reference might make more sense if placed inside PhoneValidator. I assume its meant to justify why the current validation is the way it is, which would make sense to happen at the actual validation site.

Also, if we expect this to be called multiple times, we may as well allocate the PhoneValidator internally to save it from being created on each call.

```swift
class PhoneNumberValidator: Validator {
    private let validator = PhoneValidator()

    func validate(value: Any?) -> Bool {
         guard let phoneNumber = (value as? Phone) else { return false }
         return validator.validate(phoneNumber) == .success
    }
}
```

alex-ioja-yang-cko (commented):
few questions

alex-ioja-yang-cko (approved):

PR #346, opened 2022-11-14
PIMOB-1480: Implementing editable borders for one component using `CAShapeLayer`

ghost:
[Jira ticket](https://checkout.atlassian.net/browse/PIMOB-1480)

## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/Extensions/CAShapeLayerExtension.swift
+++ b/Source/Extensions/CAShapeLayerExtension.swift
@@ -11,7 +11,7 @@ extension CAShapeLayer {
 
     /// Add Edges and corners with style
     /// - Parameter style: Protocol that provide which edge or corner will be added with style
-    func updateEdgesAndCorners(with style: ElementBorderStyle) {
+    func createCustomBorder(with style: ElementBorderStyle) {
 
         // if path not nil then it's already added so no need to draw lines or curves again
         guard self.path == nil else { return }
@@ -23,7 +23,7 @@ extension CAShapeLayer {
         let rect = bounds.insetBy(dx: style.cornerRadius / 2,
                                   dy: style.cornerRadius / 2)
         let path = UIBezierPath()
-        updateBorders(with: path, rect: rect, style: style)
+        applyBorders(with: path, rect: rect, style: style)
         updateCorners(with: path, rect: rect, style: style)
         self.path = path.cgPath
     }
@@ -47,7 +47,7 @@ extension CAShapeLayer {
     ///   - path: UIBezierPath that will be added
     ///   - rect: CGRect of current view
     ///   - style: Protocol that provide which edge will be added
-    private func updateBorders(with path: UIBezierPath, rect: CGRect, style: ElementBorderStyle) {
+    private func applyBorders(with path: UIBezierPath, rect: CGRect, style: ElementBorderStyle) {
         guard let edges = style.edges else { return }
 
         //  Bottom Left corner Radius
@@ -57,7 +57,6 @@ extension CAShapeLayer {
         // Top Left corner Radius
         let isTopLeftCornerRequired = style.corners?.contains(.topLeft) ?? false
         let topLeftCornerRadius: CGFloat = isTopLeftCornerRequired ? style.cornerRadius : 0
-        print(topLeftCornerRadius)
 
         //  Top Right corner Radius
         let isTopRightCornerRequired = style.corners?.contains(.topRight) ?? false
@@ -105,7 +104,7 @@ extension CAShapeLayer {
     ///   - path: UIBezierPath
     ///   - rect: CGRect of current view
     ///   - style: Protocol that provide which corners will be added
-    func updateCorners(with path: UIBezierPath, rect: CGRect, style: ElementBorderStyle) {
+    private func updateCorners(with path: UIBezierPath, rect: CGRect, style: ElementBorderStyle) {
         guard let corners = style.corners else { return }
 
         // bottom left corner
--- a/Source/UI/BillingForm/View/BillingFormTextFieldView.swift
+++ b/Source/UI/BillingForm/View/BillingFormTextFieldView.swift
@@ -5,37 +5,6 @@ protocol PhoneNumberTextFieldDelegate: AnyObject {
     func phoneNumberIsUpdated(number: String, tag: Int)
 }
 
-class BorderView: UIView {
-
-    private var style: ElementBorderStyle?
-
-    /// shaper layer that draw edges and corners
-    var borderLayer = CAShapeLayer()
-
-    override init(frame: CGRect) {
-        super.init(frame: frame)
-        layer.addSublayer(borderLayer)
-    }
-
-    // Keep border layer size in sync with owning view
-    override func layoutSubviews() {
-        super.layoutSubviews()
-        guard let style = style else { return }
-        borderLayer.frame = bounds
-        borderLayer.updateEdgesAndCorners(with: style)
-    }
-
-    required init?(coder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
-    }
-
-    /// Update edges and corners style
-    /// - Parameter style: protocol with some parameters for styling border and corners
-    func update(with style: ElementBorderStyle) {
-        self.style = style
-    }
-}
-
 class BillingFormTextFieldView: UIView {
 
     // MARK: - Properties
@@ -127,10 +96,10 @@ class BillingFormTextFieldView: UIView {
 
     private func updateTextFieldContainer(style: CellTextFieldStyle) {
         let borderColor = !(style.error?.isHidden ?? true) ?
-        style.textfield.borderStyle.errorColor.cgColor :
-        style.textfield.borderStyle.normalColor.cgColor
+        style.textfield.borderStyle.errorColor :
+        style.textfield.borderStyle.normalColor
         textFieldContainer.update(with: style.textfield.borderStyle)
-        textFieldContainer.borderLayer.strokeColor = borderColor
+        textFieldContainer.updateBorderColor(to: borderColor)
         textFieldContainer.backgroundColor = style.textfield.backgroundColor
     }
 
--- a/Source/UI/CommonUI/View/Component/BorderView.swift
+++ b/Source/UI/CommonUI/View/Component/BorderView.swift
@@ -1,0 +1,36 @@
+import UIKit
+
+final class BorderView: UIView {
+
+    private var style: ElementBorderStyle?
+
+    /// shaper layer that draw edges and corners
+    private var borderLayer = CAShapeLayer()
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        layer.addSublayer(borderLayer)
+    }
+
+    // Keep border layer size in sync with owning view
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        borderLayer.frame = bounds
+        guard let style = style else { return }
+        borderLayer.createCustomBorder(with: style)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    /// Update edges and corners style
+    /// - Parameter style: protocol with some parameters for styling border and corners
+    func update(with style: ElementBorderStyle) {
+        self.style = style
+    }
+
+    func updateBorderColor(to newBorderColor: UIColor) {
+        borderLayer.strokeColor = newBorderColor.cgColor
+    }
+}
--- a/Source/UI/BillingForm/Default Implementation/Elements/DefaultBorderStyle.swift
+++ b/Source/UI/BillingForm/Default Implementation/Elements/DefaultBorderStyle.swift
@@ -2,7 +2,7 @@ import UIKit
 
 struct DefaultBorderStyle: ElementBorderStyle {
     var cornerRadius: CGFloat = 10
-    var borderWidth: CGFloat = 20
+    var borderWidth: CGFloat = 1
     var normalColor: UIColor = FramesUIStyle.Color.borderPrimary
     var focusColor: UIColor = FramesUIStyle.Color.borderActive
     var errorColor: UIColor = FramesUIStyle.Color.borderError
--- a/Tests/Core/Extensions/CAShapeLayerExtensionTests.swift
+++ b/Tests/Core/Extensions/CAShapeLayerExtensionTests.swift
@@ -1,0 +1,117 @@
+import XCTest
+@testable import Frames
+
+final class CAShapeLayerExtensionTests: XCTestCase {
+
+    func testCreateCustomBorderStyle() {
+        let layer = CAShapeLayer()
+        let style = DefaultBorderStyle()
+        layer.createCustomBorder(with: style)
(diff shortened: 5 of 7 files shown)


Source/UI/CommonUI/Protocols/Elements/ElementBorderStyle.swift
```diff
@@ -0,0 +1,11 @@
+import UIKit
+
+public protocol ElementBorderStyle {
+    var cornerRadius: CGFloat { get }
+    var borderWidth: CGFloat { get }
+    var normalColor: UIColor { get }
+    var focusColor: UIColor { get }
+    var errorColor: UIColor { get }
+    var edges: UIRectEdge? { get }
+    var corners: UIRectCorner? { get }
```
alex-ioja-yang-cko, line 10: The problem with UIRectCorner is it does not currently support `automatic`.

We have different options for setting up the corners:
- selected corners (topLeft, topRight, bottomLeft, bottomRight). This is supported by UIRectCorner
- allCorners. Also supported by UIRectCorner
- autodetect / automatic / auto. This would be matched by the edges requested and imo should be the default on all our implementations. For example if edges is top, right & bottom, this corner mode will add corners to topRight & bottomRight. I don't think this can be achieved as intended without the new option.

We can however create 
```swift
public extension UIRectCorner {
    static let automatic = Self(rawValue: 1 << 4)
}
```

The problem with doing so is that any Frames consumer will see this changed native interface and may wrongly assume this is native. Which is why personally I would have favoured us creating our own optionSet for this, showing a clear ownership over the change and not interfering with Apple library.

Source/Extensions/UIBezierPathExtension.swift
```diff
@@ -0,0 +1,26 @@
...
+extension UIBezierPath {
+
+    func drawLine(startX: CGFloat, startY: CGFloat, endX: CGFloat, endY: CGFloat) {
+        let startPoint = CGPoint(x: startX, y: startY)
+        let endPoint = CGPoint(x: endX, y: endY)
+        move(to: startPoint)
+        addLine(to: endPoint)
+    }
+
+    func drawCorner(startX: CGFloat, startY: CGFloat, endX: CGFloat, endY: CGFloat, controlX: CGFloat, controlY: CGFloat) {
```
alex-ioja-yang-cko, line 19: I think we should be able to test these methods? I haven't tried this code, just checking the APIs, so might need changes to make work. But appears possible?

```swift
let testStartX: CGFloat = 0
let testStartY: CGFloat = 12
let testEndX: CGFloat = 78
let testEndY: CGFloat = 576

// Use method to create path
let testPath = UIBezierPath()
testPath.drawLine(startX: testStartX, .... )
let testCGPath = testPath.cgPath

// Prepare standard behaviour
let expectedPath = UIBezierPath()
expectedPath.move(to: CGPoint(x: testStartX, y: testStartY))
expectedPath.addLine(to: CGPoint(x: testEndX, y: testEndY))
let expectedCGPath = expectedPath.cgPath

// Ensure tested method has same outcome
let diffPath = testCGPath.symmetricDifference(expectedCGPath)
XCTAssertTrue(diffPath.isEmpty)
```

Source/UI/BillingForm/View/BillingFormTextFieldView.swift
```diff
@@ -5,6 +5,37 @@ protocol PhoneNumberTextFieldDelegate: AnyObject {
     func phoneNumberIsUpdated(number: String, tag: Int)
 }
 
+class BorderView: UIView {
```
alex-ioja-yang-cko, line 8: Might need its own file in a more generic path. I'm thinking of `UI/CommonUI/View/Component`

Source/UI/BillingForm/View/BillingFormTextFieldView.swift
```diff
@@ -5,6 +5,37 @@ protocol PhoneNumberTextFieldDelegate: AnyObject {
...
+        super.init(frame: frame)
+        layer.addSublayer(borderLayer)
+    }
+
+    // Keep border layer size in sync with owning view
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        guard let style = style else { return }
+        borderLayer.frame = bounds
+        borderLayer.updateEdgesAndCorners(with: style)
```
alex-ioja-yang-cko, line 25: We could also use an optional layer to pass the responsibility of only doing it once to the view instead of the layer. Currently, the layer inside `updateEdgesAndCorners` doesn't actually update, it checks if it exists and creates if not.

We could make a few minor changes and move that reponsability to the view, where it may be better suited:

```swift
private var borderLayer: BorderLayer?

override func layoutSubviews() {
         super.layoutSubviews()
         guard let style = style else { return }
         if borderLayer == nil {
             createLayer()
         }
         borderLayer?.frame = bounds
}

private func createLayer() {
    let borderLayer = CAShapeLayer()
    borderLayer.updateEdgesAndCorners(with: style)
    layer.addSublayer(borderLayer)
    self.borderLayer = borderLayer
}
```

We would now no longer need an init method and the CAShapeLayer extension would no longer need this code:
~`guard self.path == nil else { return }`~

Source/UI/BillingForm/View/BillingFormTextFieldView.swift
```diff
@@ -5,6 +5,37 @@ protocol PhoneNumberTextFieldDelegate: AnyObject {
...
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    /// Update edges and corners style
+    /// - Parameter style: protocol with some parameters for styling border and corners
+    func update(with style: ElementBorderStyle) {
```
alex-ioja-yang-cko, line 34: We could add another method
```swift
func updateBorderColor(to newBorderColor: UIColor) {
  // update
}
```

which would allow us to keep the layer private.

alex-ioja-yang-cko: Also, this method behaves more like a `setup` that an update. As in, it will only ever accept first style provided.

We should either document this limitation, or try and support the behaviour as described:
```swift
if self.style != style {
    self. borderLayer = nil
    self.style = style
    self.layoutSubviews()
}
```

Source/UI/BillingForm/View/BillingFormTextFieldView.swift
```diff
@@ -96,12 +127,10 @@ class BillingFormTextFieldView: UIView {
...
-        style.textfield.errorBorderColor.cgColor :
-        style.textfield.normalBorderColor.cgColor
-
-        textFieldContainer.layer.borderColor = borderColor
-        textFieldContainer.layer.cornerRadius = style.textfield.cornerRadius
-        textFieldContainer.layer.borderWidth = style.textfield.borderWidth
+        style.textfield.borderStyle.errorColor.cgColor :
+        style.textfield.borderStyle.normalColor.cgColor
+        textFieldContainer.update(with: style.textfield.borderStyle)
+        textFieldContainer.borderLayer.strokeColor = borderColor
```
alex-ioja-yang-cko, line 133: I think this method should be moved inside the BorderView so we don't expose the `borderLayer` directly?

```suggestion
        textFieldContainer.updateBorderColor(to: borderColor)
```

Source/UI/BillingForm/View/BillingFormTextFieldView.swift
```diff
@@ -5,6 +5,37 @@ protocol PhoneNumberTextFieldDelegate: AnyObject {
     func phoneNumberIsUpdated(number: String, tag: Int)
 }
 
+class BorderView: UIView {
+
```
alex-ioja-yang-cko, line 9: We need to unit test `BorderView`.  Some ideas:
```swift
func testInitDoesNotAddLayer()
func testLayoutSubviewsDoesNotAddLayerIfNoStyle()

// These could be merged into 1 if we force the layer creation when style is provided
func testUpdateStyleDoesNotAddLayer()
func testLayoutSubviewsAddsLayerIfStylePresent()

// Suggested to update color internally and not expose directly the layer reference
func testBorderColorUpdate()
```

The layer can always be found via `view.layer.sublayers`, which we can use to test these behaviours.

Source/Extensions/CAShapeLayerExtension.swift
```diff
@@ -0,0 +1,151 @@
...
+        guard let edges = style.edges else { return }
+
+        //  Bottom Left corner Radius
+        let isBottomLeftCornerRequired = style.corners?.contains(.bottomLeft) ?? false
+        let bottomLeftCornerRadius: CGFloat = isBottomLeftCornerRequired ? style.cornerRadius : 0
+
+        // Top Left corner Radius
+        let isTopLeftCornerRequired = style.corners?.contains(.topLeft) ?? false
+        let topLeftCornerRadius: CGFloat = isTopLeftCornerRequired ? style.cornerRadius : 0
+        print(topLeftCornerRadius)
```
alex-ioja-yang-cko, line 60:
```suggestion 
```

Source/UI/CommonUI/Protocols/Elements/ElementBorderStyle.swift
alex-ioja-yang-cko, line 10: I don't think we need to have optional edges and corners? They should always be provided, and being option sets, even values as [] will work.

Source/Extensions/CAShapeLayerExtension.swift
```diff
@@ -0,0 +1,151 @@
...
+                          endY: rect.maxY)
+        }
+    }
+
+    /// Add corners with style
+    /// - Parameters:
+    ///   - path: UIBezierPath
+    ///   - rect: CGRect of current view
+    ///   - style: Protocol that provide which corners will be added
+    func updateCorners(with path: UIBezierPath, rect: CGRect, style: ElementBorderStyle) {
```
alex-ioja-yang-cko, line 108:
```suggestion
    private func applyCorners(with path: UIBezierPath, rect: CGRect, style: ElementBorderStyle) {
```

Source/Extensions/CAShapeLayerExtension.swift
```diff
@@ -0,0 +1,151 @@
...
+     maxX = 100
+     maxY = 30
+     */
+
+    /// Add corners with style
+    /// - Parameters:
+    ///   - path: UIBezierPath that will be added
+    ///   - rect: CGRect of current view
+    ///   - style: Protocol that provide which edge will be added
+    private func updateBorders(with path: UIBezierPath, rect: CGRect, style: ElementBorderStyle) {
```
alex-ioja-yang-cko, line 50:
```suggestion
    private func applyBorders(with path: UIBezierPath, rect: CGRect, style: ElementBorderStyle) {
```

Source/Extensions/CAShapeLayerExtension.swift
```diff
@@ -0,0 +1,151 @@
...
+//  Created by Ehab Alsharkawy on 14/11/2022.
+//
+
+import UIKit
+
+extension CAShapeLayer {
+
+    /// Add Edges and corners with style
+    /// - Parameter style: Protocol that provide which edge or corner will be added with style
+    func updateEdgesAndCorners(with style: ElementBorderStyle) {
```
alex-ioja-yang-cko, line 14:
```suggestion
    func createCustomBorder(with style: ElementBorderStyle) {
```

Source/Extensions/CAShapeLayerExtension.swift
```diff
@@ -0,0 +1,151 @@
...
+        let topRightCornerRadius = isTopRightCornerRequired ? style.cornerRadius : 0
+
+        // Bottom Right corner Radius
+        let isBottomRightCornerRequired = style.corners?.contains(.bottomRight) ?? false
+        let bottomRightCornerRadius = isBottomRightCornerRequired ? style.cornerRadius : 0
+
+        // left line
+        if edges.contains(.left) {
+            path.drawLine(startX: rect.minX,
+                          startY: rect.minY + topLeftCornerRadius,
```
alex-ioja-yang-cko, line 73: For these corner radius offsets, I really wonder whether they are full offset on both lines, or half of offset on both sides.

If you have a setup to easily check, could you try a standard border with a corner radius of something big, like 50, then try the same with this approach and check whether they look the same?

alex-ioja-yang-cko (commented):
Looks great. A few things to talk about and maybe small improvements, but overall 👍 💯

Source/UI/BillingForm/View/BillingFormTextFieldView.swift
alex-ioja-yang-cko: Could you describe why its not working? We should be able to achieve the same outcome while being clear and better enforcing responsibilities of involved objects, so if you say its not possible, I would like to understand what the challenge is.

Source/UI/CommonUI/View/Component/BorderView.swift
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    /// Update edges and corners style
+    /// - Parameter style: protocol with some parameters for styling border and corners
+    func update(with style: ElementBorderStyle) {
+        self.style = style
```
alex-ioja-yang-cko, line 30: I know its not currently supported, but as it stands if a style is overridden it will never materialise. We could fix by:
```suggestion
        self.style = style
        borderLayer.removeFromSuperlayer()
        borderLayer = CAShapeLayer()
        borderLayer.createCustomBorder(with: style)
        layer.addSublayer(borderLayer)
```

And in fairness once we do this, we no longer need to create the border layer on layout subviews, we create it when a style is provided. And we can now have the source of truth on whether to replace the layout in the style provided being different from last or not.
```swift
guard style != self.style else { return }
```

Tests/UI/BillingForm/View/BillingFormTextFieldViewTests.swift
```diff
@@ -24,11 +24,6 @@ class BillingFormTextFieldViewTests: XCTestCase {
 
     func testTextFieldContainerStyle() {
         XCTAssertEqual(view.textFieldContainer.backgroundColor, style.textfield.backgroundColor)
-        XCTAssertEqual(view.textFieldContainer.layer.borderColor, style.textfield.normalBorderColor.cgColor)
-        XCTAssertEqual(view.textFieldContainer.layer.cornerRadius, style.textfield.cornerRadius)
-        XCTAssertEqual(view.textFieldContainer.layer.borderWidth, style.textfield.borderWidth)
-        XCTAssertEqual(view.textFieldContainer.layer.cornerRadius, style.textfield.cornerRadius)
-        XCTAssertEqual(view.textFieldContainer.layer.borderWidth, style.textfield.borderWidth)
```
alex-ioja-yang-cko, line 31: Would these tests not work anymore, just having to update the `style` path to the new wrapper object for border?

alex-ioja-yang-cko (approved):
Didn't get to run, I trust you've done that and our testing may catch details too, but otherwise code looks good.

Minor suggestion left which I think would help behave more consistently (beyond our current needs) and question on an unit test checks being removed, whether we can keep them.

PR #347, opened 2022-11-15
PIMOB-1506: Can input CVVs with leading 0's

ghost:
[Jira ticket](https://checkout.atlassian.net/browse/PIMOB-1506)

## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/UI/PaymentForm/View/SecurityCodeViewModel.swift
+++ b/Source/UI/PaymentForm/View/SecurityCodeViewModel.swift
@@ -41,8 +41,8 @@ final class SecurityCodeViewModel {
             isInputValid = cardValidator.validate(cvv: cvv, cardScheme: scheme) == .success
         }
 
-        guard let cleanedInput = newInput?.filter({ !$0.isWhitespace }),
-              (Int(cleanedInput) ?? 0 > 0) || cleanedInput.isEmpty,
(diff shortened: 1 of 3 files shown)


Source/UI/PaymentForm/View/SecurityCodeViewModel.swift
```diff
@@ -42,7 +42,7 @@ final class SecurityCodeViewModel {
         }
 
         guard let cleanedInput = newInput?.filter({ !$0.isWhitespace }),
-              (Int(cleanedInput) ?? 0 > 0) || cleanedInput.isEmpty,
```
alex-ioja-yang-cko, line 45: Are you sure this is the problem? 

Looks to me like we should just improve the filter to not only remove whitespaces, but any non digit character. Then on the remaining string, check for input length. Why do we need to parse to Int now, or compare with 0?

alex-ioja-yang-cko: But logically, is that the right logical thing to do?

**Need**
I want to ensure only digits are entered.
I want to ensure the correct number of digits are entered.

At which stage do I need to check if the value of the number created from those digits is comparable to something?

alex-ioja-yang-cko (approved):

PR #351, opened 2022-11-18
CheeryPick commit: PIMOB-1480: edit borders for one component using `CAShapeLayer` (#346)

ghost:
commit: 33ac1019e8db76af9ec6884adc82736c15cc469d
## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/UI/Theme/Generic/Theme+TextField.swift
+++ b/Source/UI/Theme/Generic/Theme+TextField.swift
@@ -15,7 +15,7 @@ public extension Theme {
         public var text: String
         public var isSupportingNumericKeyboard = true
         public var textAlignment: NSTextAlignment = .natural
-        public var height: Double = 30
+        public var height: Double = 56
         public var placeholder: String?
         public var tintColor: UIColor
         public var isHidden = false


Source/Extensions/CAShapeLayerExtension.swift
```diff
@@ -0,0 +1,150 @@
...
+
+        // always fill is clear color
+        fillColor = UIColor.clear.cgColor
+
+        lineWidth = style.borderWidth
+        let rect = bounds.insetBy(dx: style.cornerRadius / 2,
+                                  dy: style.cornerRadius / 2)
+        let path = UIBezierPath()
+        applyBorders(with: path, rect: rect, style: style)
+        updateCorners(with: path, rect: rect, style: style)
```
alex-ioja-yang-cko, line 27: I think these 2 methods should read the same

```suggestion
        applyBorders(with: path, rect: rect, style: style)
        applyCorners(with: path, rect: rect, style: style)
```

Source/Extensions/CAShapeLayerExtension.swift
```diff
@@ -0,0 +1,150 @@
...
+            |           |
+            |           |
+     (0,30) c-----------c  (100,30)
+
+     origin = (0,0)
+     minx = 0
+     minY = 0
+     maxX = 100
+     maxY = 30
+     */
```
alex-ioja-yang-cko, line 43: Is there any value in having this? I think it makes sense to use when writing the code, but given the unit tests and all else, would this actually help someone working on the code in the future? imo the numbers may actually end up confusing them, so we could either use x and y instead of 100 and 30, or we may even just remove the comment altogether.

What do you think?

alex-ioja-yang-cko (commented):
Just a few comments, otherwise looks same as when it first got merged 😬.

Happy to approve once release is made (just to make sure we don't need to keep changing git history if we forget).

Source/UI/Theme/Generic/Theme+TextField.swift
```diff
@@ -15,18 +15,24 @@ public extension Theme {
         public var text: String
         public var isSupportingNumericKeyboard = true
         public var textAlignment: NSTextAlignment = .natural
-        public var height: Double = 56
-        public var cornerRadius: CGFloat
-        public var borderWidth: CGFloat
+        public var height: Double = 30
```
alex-ioja-yang-cko, line 18: This needs to be 56

alex-ioja-yang-cko (changes requested):
Please undo the Theme+TextField height change

alex-ioja-yang-cko (approved):
👍

PR #352, opened 2022-11-21
PIMOB-1511: edit borders for all UI components

ghost:
Please review this PR https://github.com/checkout/frames-ios/pull/351 firstly
[Jira ticket](https://checkout.atlassian.net/browse/PIMOB-1511)

## Proposed changes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/Source/UI/CommonUI/View/Component/InputView.swift
+++ b/Source/UI/CommonUI/View/Component/InputView.swift
@@ -221,12 +221,11 @@ extension InputView: TextFieldViewDelegate {
 
     func textFieldShouldEndEditing(textField: UITextField, replacementString: String) -> Bool {
         let shouldEndEditing = delegate?.textFieldShouldEndEditing(textField: textField, replacementString: replacementString) ?? true
-        if shouldEndEditing {
-            let normalColor = style?.textfield.borderStyle.normalColor
-            let errorColor = style?.textfield.borderStyle.errorColor
-            if let color = style?.error?.isHidden == true ? normalColor : errorColor {
-                textFieldContainerBorder.updateBorderColor(to: color)
-            }
+        if let style = style, shouldEndEditing {
+            let normalColor = style.textfield.borderStyle.normalColor
+            let errorColor = style.textfield.borderStyle.errorColor
+            let textfieldColor = style.error?.isHidden == true ? normalColor : errorColor
+            textFieldContainerBorder.updateBorderColor(to: textfieldColor)
         }
         return shouldEndEditing
     }


Source/Extensions/CAShapeLayerExtension.swift
```diff
@@ -13,18 +13,14 @@ extension CAShapeLayer {
     /// - Parameter style: Protocol that provide which edge or corner will be added with style
     func createCustomBorder(with style: ElementBorderStyle) {
 
-        // if path not nil then it's already added so no need to draw lines or curves again
-        guard self.path == nil else { return }
-
```
alex-ioja-yang-cko, line 18: You cannot remove this without making change to how `createCustomBorder` is consumed. This bit of code in `BorderView` will end up stacking border over border, especially if inside a dequeable cell that keeps reusing.

Screenshot 2022-11-29 at 09 30 59

As I have previously suggested, dropping that `guard let style = style` and the subsequent line in the `layoutSubviews` should be removed.

Source/Extensions/CAShapeLayerExtension.swift
```diff
@@ -13,18 +13,14 @@ extension CAShapeLayer {
...
         // always fill is clear color
         fillColor = UIColor.clear.cgColor
 
         lineWidth = style.borderWidth
-        let rect = bounds.insetBy(dx: style.cornerRadius / 2,
-                                  dy: style.cornerRadius / 2)
+
         let path = UIBezierPath()
-        applyBorders(with: path, rect: rect, style: style)
-        applyCorners(with: path, rect: rect, style: style)
```
alex-ioja-yang-cko, line 27: Would like to understand what changed here and why its better to remove the offset now?

Source/Core/Constants/Constants+Style.swift
```diff
@@ -13,6 +13,11 @@ extension Constants {
 
       static let animationLength: TimeInterval = 0.25
 
+        enum BorderStyle: Double {
+            case cornerRadius = 10
+            case borderWidth = 1
+        }
```
alex-ioja-yang-cko, line 19: Correct Constants use here would be:
```suggestion
        enum BorderStyle {
            static let cornerRadius: Double = 10
            static let borderWidth: Double = 1
        }
```

This will 
- enable the consumer to navigate object hierarchy in a logical manner
- prevent being able to have an initialiser
- remove need for `rawValue` at use site, offering a simpler and cleaner syntax

Source/UI/CommonUI/Protocols/Elements/ElementBorderStyle.swift
```diff
@@ -1,11 +1,11 @@
...
-}
+    public protocol ElementBorderStyle {
+        var cornerRadius: CGFloat { get }
+        var borderWidth: CGFloat { get }
+        var normalColor: UIColor { get }
+        var focusColor: UIColor { get }
+        var errorColor: UIColor { get }
+        var edges: UIRectEdge? { get }
+        var corners: UIRectCorner? { get }
+    }
```
alex-ioja-yang-cko, line 11: There doesn't appear to be anything else inside this file, so why the indentation?

Source/UI/CommonUI/View/Component/BorderView.swift
```diff
@@ -28,6 +28,7 @@ final class BorderView: UIView {
     /// - Parameter style: protocol with some parameters for styling border and corners
     func update(with style: ElementBorderStyle) {
         self.style = style
+        borderLayer.createCustomBorder(with: style)
     }
```
alex-ioja-yang-cko, line 32: We could be more efficient here and only createCustomBorder if style is different from previous style. We might also need to remove an existing style before adding another.

Source/UI/CommonUI/View/Component/InputView.swift
```diff
@@ -216,9 +222,12 @@ extension InputView: TextFieldViewDelegate {
...
         let shouldEndEditing = delegate?.textFieldShouldEndEditing(textField: textField, replacementString: replacementString) ?? true
         if shouldEndEditing {
-            textFieldContainerBorder.layer.borderColor = (style?.error?.isHidden ?? true) ?
-            style?.textfield.borderStyle.normalColor.cgColor :
-                style?.textfield.borderStyle.errorColor.cgColor
+            if let color = (style?.error?.isHidden ?? true) ?
+            style?.textfield.borderStyle.normalColor :
+                    style?.textfield.borderStyle.errorColor {
+                textFieldContainerBorder.updateBorderColor(to: color)
+            }
```
alex-ioja-yang-cko, line 229: I don't think we should use ternary operator in an if statement. It makes things very fuzzy.

ie:
```swift
if let variable = condition ? option1 : option2 {
```

Your concern seems to be strictly on the optionality of style, so if you just unwrap that, it should become a lot easier.

```swift
if shouldEndEditing,
    let style = style {
        let color = style.error?.isHidden == false : style.textfield.borderStyle.errorColor : style.textfield.borderStyle.normalColor
        textFieldContainerBorder.updateBorderColor(to: color)
}
```

Source/UI/CommonUI/View/Core/ButtonView.swift
```diff
@@ -9,6 +9,12 @@ class ButtonView: UIView {
     var style: ElementButtonStyle?
     lazy var constraintLeading: NSLayoutConstraint? = buttonTextLabel.leadingAnchor.constraint(equalTo: leadingAnchor, constant: 120.0)
 
+    private(set) lazy var borderView: BorderView = {
+        let view = BorderView().disabledAutoresizingIntoConstraints()
+        view.backgroundColor = .clear
```
alex-ioja-yang-cko, line 14: Could we just move this background color inside BorderView ? I don't see any reason why a view designed to provide border would ever have a default background color.

alex-ioja-yang-cko (commented):
partial review

Source/Extensions/CAShapeLayerExtension.swift
alex-ioja-yang-cko: Cool, haven't noticed that, was under the assumption we're adding layer on top of layer. I do think though that we would only ever need to createCustomBorder once, when the style is injected, then only update the frame as subviews are told to layout?

Source/UI/CommonUI/View/Component/BorderView.swift
```diff
@@ -0,0 +1,37 @@
+import UIKit
+
+final class BorderView: UIView {
```
alex-ioja-yang-cko, line 3: This file already exists.

alex-ioja-yang-cko (changes requested):
PR is duplicating another previously [merged PR](https://github.com/checkout/frames-ios/pull/351).

Please see what needs to be done to not duplicate logic & files

Source/UI/CommonUI/View/Core/ButtonView.swift
alex-ioja-yang-cko: > BorderView is responsable only for borders for buttons or textfield or any UI components

Exactly. So it should not be expected for consumer to write `view.backgroundColor = .clear`.

Source/UI/CommonUI/View/Component/BorderView.swift
alex-ioja-yang-cko: What I am saying is that if the same style is assigned, we can save some processing by not having to create a new path and assign it. I cannot see anything at code level being efficient on this, so currently if a style is provided, the path is created **every time**

Source/UI/CommonUI/View/Component/InputView.swift
```diff
@@ -122,6 +121,10 @@ class InputView: UIView {
         errorView.update(style: style.error)
         errorView.isHidden = style.error?.isHidden ?? false
     }
+
+    func updateBorderColor(with color: UIColor) {
+        textFieldContainerBorder.updateBorderColor(to: color)
+    }
 }
```
alex-ioja-yang-cko, line 128: I think code is easier to navigate if we organise our logic by consistent criteria. In this case I would expect access control to play a part in the grouping of methods, and this internal method to appear before any private logic.

Source/UI/PaymentForm/Default Implementation/DefaultPayButtonFormStyle.swift
```diff
@@ -17,15 +17,15 @@ public struct DefaultPayButtonFormStyle: ElementButtonStyle {
...
-  public var normalBorderColor: UIColor = .clear
-  public var focusBorderColor: UIColor = .clear
-  public var errorBorderColor: UIColor = .clear
   public var imageTintColor: UIColor = .clear
   public var isHidden = false
   public var isEnabled = false
   public var height: Double = 56
   public var width: Double = 0
-  public var cornerRadius: CGFloat = 10
   public var borderWidth: CGFloat = 0
```
alex-ioja-yang-cko, line 25:
```suggestion
```

borderWidth also moved

Source/UI/PaymentForm/ViewController/PaymentViewController.swift
```diff
@@ -236,6 +236,7 @@ extension PaymentViewController {
     viewModel.updateCardNumberView = { [weak self] in
       DispatchQueue.main.async {
         self?.updateCardNumber()
+        self?.view.layoutSubviews()
```
alex-ioja-yang-cko, line 239: This looks like we're connecting the card number input change with some form of subview resizing. I am not sure that should be needed, what is the problem we are solving with this line?

alex-ioja-yang-cko (commented):
Mostly looking good, just details to iron out 👍

Source/UI/CommonUI/View/Component/InputView.swift
```diff
@@ -216,9 +222,11 @@ extension InputView: TextFieldViewDelegate {
     func textFieldShouldEndEditing(textField: UITextField, replacementString: String) -> Bool {
         let shouldEndEditing = delegate?.textFieldShouldEndEditing(textField: textField, replacementString: replacementString) ?? true
         if shouldEndEditing {
-            textFieldContainerBorder.layer.borderColor = (style?.error?.isHidden ?? true) ?
-            style?.textfield.borderStyle.normalColor.cgColor :
-                style?.textfield.borderStyle.errorColor.cgColor
+            let normalColor = style?.textfield.borderStyle.normalColor
+            let errorColor = style?.textfield.borderStyle.errorColor
+            if let color = style?.error?.isHidden == true ? normalColor : errorColor {
```
alex-ioja-yang-cko, line 227: Syntax its simpler now by separating the color declarations, but you still have a ternary operator used inside an if unwrapping.
`if let optional ` is one thing, but
`if let value = condition ? optional1 : optional 2` is not easy to read imo.

I would still suggest that unwrapping the `style` inside the if will provide simplest writing for this.

```suggestion
    if shouldEndEditing,
        let style = style {
                let normalColor = style.textfield.borderStyle.normalColor
                let errorColor = style.textfield.borderStyle.errorColor
                let textfieldColor = style.error?.isHidden == true ? normalColor : errorColor
                textFieldContainerBorder.updateBorderColor(to: textfieldColor)
    }
```

alex-ioja-yang-cko (approved):

PR #357, opened 2022-12-01
Added public init to TokenDetails model in order to facilitate public…

Aecasorg:
… access.

## Proposed changes

Add public init to TokenDetails in order to facilitate access to struct from outside package. In this instance it would be used to aid mocking test results in main project where Frames package is used.

## Types of changes

Not really a bug per say...

## Checklist

## Further comments

No need to run tests for this as it is only adding a public init to a model.

### DIFF
--- a/Checkout/Checkout/Source/Model/TokenDetails.swift
+++ b/Checkout/Checkout/Source/Model/TokenDetails.swift
@@ -25,6 +25,25 @@ public struct TokenDetails: Equatable {
   public let billingAddress: Address?
   public let phone: Phone?
   public let name: String?
+    
+  public init(type: TokenType, token: String, expiresOn: String, expiryDate: ExpiryDate, scheme: Card.Scheme?, last4: String, bin: String, cardType: String?, cardCategory: String?, issuer: String?, issuerCountry: String?, productId: String?, productType: String?, billingAddress: Address?, phone: Phone?, name: String?) {
+      self.type = type
+      self.token = token
+      self.expiresOn = expiresOn
+      self.expiryDate = expiryDate
+      self.scheme = scheme
+      self.last4 = last4
+      self.bin = bin
+      self.cardType = cardType
+      self.cardCategory = cardCategory
+      self.issuer = issuer
+      self.issuerCountry = issuerCountry
+      self.productId = productId
+      self.productType = productType
+      self.billingAddress = billingAddress
+      self.phone = phone
+      self.name = name
+    }
 
   public enum TokenType: String, Codable, Equatable {
     case card


Checkout/Checkout/Source/Model/TokenDetails.swift
```diff
@@ -25,6 +25,25 @@ public struct TokenDetails: Equatable {
   public let billingAddress: Address?
   public let phone: Phone?
   public let name: String?
+    
+  public init(type: TokenType, token: String, expiresOn: String, expiryDate: ExpiryDate, scheme: Card.Scheme?, last4: String, bin: String, cardType: String?, cardCategory: String?, issuer: String?, issuerCountry: String?, productId: String?, productType: String?, billingAddress: Address?, phone: Phone?, name: String?) {
```
alex-ioja-yang-cko, line 29: Hi @Aecasorg , thank you for your involvement and attempt at making the SDK work better for our users. 

I see you explained the PR as 
> Add public init to TokenDetails in order to facilitate access to struct from outside package. In this instance it would be used to aid mocking test results in main project where Frames package is used.

We have previously touched on this internally and the issue is we are not comfortable with introducing the ability for consumers to create mock tokens outside of testing environments. This would make it too easy for implementations that reach production environments to create false outcomes, violating the reliability of the data our SDK outputs.

For unit testing you may be able to use specific Swift syntax to improve access control in a way that facilitates your requirements.

Aecasorg: Thank you for your response! 

I completely get your concern regarding mocking token outside testing environments. 

I've extended the `CheckoutAPIService` with a protocol in order to access the `createCardToken(card: Card)` function for mocking. Before Frames 4.0 I could to create a `CkoCardTokenResponse` from json, but that is less easy with `TokenDetails`. Did you have any "specific Swift syntax" in mind that would aid with this?

Any help with the above would be most appreciated.

alex-ioja-yang-cko: Hi @Aecasorg by "specific Swift syntax" I was thinking of `@testable import Frames` inside your unit test target. Having tried it now myself I found a side effect of our type alias definition for `TokenDetails`, which means you need to import a different module to access the initialiser.

I have been trying to see how we can help and ended up with this helper object and sample test to validate it compiles and builds:

## Helper Object inside your unit test module
`TokenCreator.swift`

```Swift
@testable import Checkout

enum TokenCreator {

    static func create(type: TokenDetails.TokenType,
                       token: String = "",
                       expiresOn: String = "",
                       expiryDate: ExpiryDate = ExpiryDate(month: 12, year: 2055),
                       scheme: Card.Scheme? = nil,
                       last4: String = "",
                       bin: String = "",
                       cardType: String? = nil,
                       cardCategory: String? = nil,
                       issuer: String? = nil,
                       issuerCountry: String? = nil,
                       productId: String? = nil,
                       productType: String? = nil,
                       billingAddress: Address? = nil,
                       phone: TokenDetails.Phone? = nil,
                       name: String? = nil) -> TokenDetails {
        TokenDetails(type: type,
                     token: token,
                     expiresOn: expiresOn,
                     expiryDate: expiryDate,
                     scheme: scheme,
                     last4: last4,
                     bin: bin,
                     cardType: cardType,
                     cardCategory: cardCategory,
                     issuer: issuer,
                     issuerCountry: issuerCountry,
                     productId: productId,
                     productType: productType,
                     billingAddress: billingAddress,
                     phone: phone,
                     name: name)
    }

}
```

## Sample unit test to validate it all compiles & works
```Swift
func testExample() {
    let testName = "testToken"
    let token = TokenCreator.create(type: .applePay, name: testName)
    XCTAssertEqual(token.name, testName)
}
```

Please let me know if this works for your requirements. I believe this should enable you to mock data for unit testing, whilst preserving the integrity of SDK output.

Do be warned in a future version we may be dropping the module `Checkout` altogether (nothing confirmed or planned right now), which may require to import Frames instead, but we would still aim to enable integration unit tests that rely on our SDKs output.

Aecasorg: Thanks for this! This looks helpful. Unfortunately I still get the "Module 'Checkout' was not compiled for testing" error so I can't test it. It looks like an enable testing flag is needed to be added to the SPM Package.swift file in order to enable any external testable access to the Checkout package. 

```
.target(
            name: "Checkout",
            dependencies: [
                "CheckoutEventLoggerKit",
            ],
            path: "Checkout/Checkout/Source",
            swiftSettings: [
                .unsafeFlags(["-enable-testing"])
            ]
        )
```

I have not been able to test and verify this as Xcode refuses to access my GitHub account for adding an SPM. Is there any way this could be tested?

alex-ioja-yang-cko:
In case other engineers encounter a similar issue, will document here why this appeared to be helpful.

## [Example problem case](https://antran.app/2021/custom_build_configuration_swift_package_tests/)
When Xcode imports Packages via SPM, it will compile them with testing support if compiling for Debug, or compile for release if "Non-Debug". Xcode uses `Configurations` to determine whether it will compile for Debug or Release. By default, Xcode projects come with "Debug" and "Release" configurations, so the mapping is very easy:
- Debug configuration = **yes** testing support for package
- Release configuration = **no** testing support for package

This becomes an issue when the project contains a custom Configuration (lets use Automation as our example). Since any non debug configuration will be treated by compiling for Release, this now means that:
- Automation configuration = **no** testing support for package


 


## Solutions
1. An ideal solution would appear to create some SPM improvement that enables it to ship with different configurations supported. This has already been [proposed](https://forums.swift.org/t/update-spm-to-support-custom-configuration-names/43075) but the fact this would create stronger links between SPM & Xcode makes it a difficult change to see as an improvement. After all, unlike Cocoapods, SPM is built to enable Swift consumption in different environments. Xcode is just 1 of these environments.

2. Looking at the discussion mentioned on point 1, there is an answer that solves the problem. This can also be observed if approaching this problem from scratch (vs in a preexisting complex project), so let's try to do this instead and work our way to the solution. 
In a standard Xcode project you start with 2 configurations (Debug & Release).
Creating a new configuration will require you to duplicate an existing one. If we duplicate Debug, we get `Debug copy`. Changing our unit tests to run in `Debug copy`, will actually enable the compilation of the code. We can then observe that packages are compiled with testing support when the Configuration has a format of `Debug {config_name}`. We can then solve our problem by renaming our custom configuration from ~Automation~ to **Debug Automation**. 

This was shown to work using Xcode 14.1 & 14.0 , but the behaviour is dependant on Xcode interaction with SPM (hence may not still be accurate at a later date).


 
 
 
## Disclaimer

I'm sure there must be a WWDC talk documenting this for a few seconds, but from our team we are not aware with it. We're also not familiar with any official docs from Apple documenting this. If you're aware of such a source please let us know and we'd be happy to refer this to the official support.

PR #361, opened 2022-12-05
Update main with latest release changes

alex-ioja-yang-cko:
Changes for 4.0.3 release were done in a branch and not included in `main`.

### DIFF
--- a/Frames.podspec
+++ b/Frames.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "Frames"
-  s.version      = "4.0.2"
+  s.version      = "4.0.3"
   s.summary      = "Checkout API Client, Payment Form UI and Utilities in Swift"
   s.description  = <<-DESC
   Checkout API Client and Payment Form Utilities in Swift.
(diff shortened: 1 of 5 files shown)


Frames.podspec
```diff
@@ -15,12 +15,10 @@ Pod::Spec.new do |s|
...
   s.exclude_files = "Classes/Exclude"
-  s.resource_bundles = {
-    'Frames' => ['Source/Resources/**/*']
-  }
+  s.resources = 'Source/Resources/**'
 
   s.dependency 'PhoneNumberKit', '3.3.3'
   s.dependency 'CheckoutEventLoggerKit', '1.2.0'
-  s.dependency 'Checkout', "#{s.version}"
+  s.dependency 'Checkout', '4.0.2'
```
deepesh-vasthimal-cko, line 22: Why are we hardcoding this now ? ,  "#{s.version}" would have been better right ?

deepesh-vasthimal-cko (changes requested):
Just one comment

deepesh-vasthimal-cko (approved):

Frames.podspec
alex-ioja-yang-cko: 4.0.3 did not require changes to Checkout pod. And looking at the roadmap, I don't see us needing to update Checkout pod for a while, so might be simpler to stop releasing Checkout for the sake of syncing numbers.

PR #362, opened 2022-12-05
PIMOB-1511: Minor theme related fixes

alex-ioja-yang-cko:
- using wrong property for border width
- removed theme properties that are not used (and don't really make sense to exist)
- indentation fixes where code changes occurred


Source/UI/Theme/Generic/Theme+Button.swift
```diff
@@ -53,12 +53,12 @@ public extension Theme {
     func buildNavigationButton(text: String) -> NavigationButtonStyle {
         NavigationButtonStyle(
             borderStyle: ThemeBorderStyle(cornerRadius: self.textInputBorderRadius,
-                                         borderWidth: self.textInputBorderRadius,
-                                         normalColor: self.textInputBorderColor,
-                                         focusColor: self.focussedTextInputBorderColor,
-                                         errorColor: self.errorBorderColor,
-                                         edges: .all,
-                                         corners: nil),
+                                          borderWidth: self.textInputBorderWidth,
```
deepesh-vasthimal-cko, line 56: Whats the reason to replace `textInputBorderRadius` with `textInputBorderWidth` ?

deepesh-vasthimal-cko (changes requested):
Just one comment.

deepesh-vasthimal-cko (approved):

Source/UI/Theme/Generic/Theme+Button.swift
alex-ioja-yang-cko: Previously we were wrongly using `textInputBorderRadius` for both `borderWidth` and `cornerRadius`.

Just linking correctly the source for the border width (which should be `textInputBorderWidth`, not `textInputBorderRadius`)

PR #364, opened 2022-12-21
PIMOB-XXXX: Surface additional backend provided property on token creation

alex-ioja-yang-cko:
## Proposed changes

Following [internal requirements](https://checkout.slack.com/archives/G017VRMD5K7/p1671545345323699?thread_ts=1671543589.742489&cid=G017VRMD5K7), enable the parsing of the missing token property and surface it to consumer of SDK.


CheckoutTests/Tokenisation/TokenDetailsFactoryTests.swift
```diff
@@ -8,20 +8,81 @@
...
-  }
+    
+    func testPropertiesMapped() {
+        let testType = TokenRequest.TokenType.card
+        let testToken = "12345676sdgdh"
+        let testExpiresOn = "tomorrow"
+        let testExpiryMonth = 6
+        let testExpiryYear = 3011
+        let testScheme = "visa"
+        let testSchemeLocal = "mada"
```
deepesh-vasthimal-cko, line 20: hardcoded string is fine, but if you want to actually write the proper value this should be visa or mastercard, see [mada test cards](https://www.checkout.com/docs/payments/payment-methods/mada#Mada_test_cards).

CheckoutTests/Tokenisation/TokenDetailsFactoryTests.swift
```diff
@@ -8,20 +8,81 @@
...
-    XCTAssertEqual(subject.create(tokenResponse: tokenResponse), expectedTokenDetails)
-  }
+    
+    func testPropertiesMapped() {
+        let testType = TokenRequest.TokenType.card
+        let testToken = "12345676sdgdh"
+        let testExpiresOn = "tomorrow"
+        let testExpiryMonth = 6
+        let testExpiryYear = 3011
+        let testScheme = "visa"
```
deepesh-vasthimal-cko, line 19: It can be anything, but this should be mada as well for proper readability of schemeLocal testing.

deepesh-vasthimal-cko (approved):

CheckoutTests/Tokenisation/TokenDetailsFactoryTests.swift
alex-ioja-yang-cko: The test is purely validating that TokenDetailsFactory is able to map correctly data from network object to SDK model. As network object actually handles String for these properties, there is an infinite number of permutations we could test for. That would defeat the purpose of testing that a given valid network object is parsed successfully to an SDK model.

CheckoutTests/Tokenisation/TokenDetailsFactoryTests.swift
alex-ioja-yang-cko: I don't understand your suggestion. Similarly to [other response](https://github.com/checkout/frames-ios/pull/364#discussion_r1059341654), the goal of the test is to ensure a valid network object will be transformed successfully into an SDK model.
Not to validate all possible permutations.

PR #367, opened 2022-12-22
PIMOB-XXXX: Pass user cancelled event in the completion handler

alex-ioja-yang-cko:
## Proposed changes

By passing the cancellation as an error of `userCancelled`, the consumer can successfully track the occurrence of this event.

Will require product review and internal discussion over possible alternatives that may fit better into the product roadmap.

### DIFF
--- a/Checkout/Checkout/Source/Tokenisation/Error/TokenisationError+TokenRequest.swift
+++ b/Checkout/Checkout/Source/Tokenisation/Error/TokenisationError+TokenRequest.swift
@@ -11,6 +11,7 @@ extension TokenisationError {
   public enum TokenRequest: CheckoutError {
     case couldNotBuildURLForRequest
     case applePayTokenInvalid
+    case userCancelled
     case cardValidationError(ValidationError.Card)
     case networkError(NetworkError)
     case serverError(ServerError)
@@ -21,6 +22,8 @@ extension TokenisationError {
         return 3001
       case .applePayTokenInvalid:
         return 1100
+      case .userCancelled:
+        return 0
       case .cardValidationError(let cardValidationError):
         return cardValidationError.code
       case .networkError(let networkError):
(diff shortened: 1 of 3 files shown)


Checkout/Checkout/Source/Tokenisation/Error/TokenisationError+TokenRequest.swift
```diff
@@ -21,6 +22,8 @@ extension TokenisationError {
         return 3001
       case .applePayTokenInvalid:
         return 1100
+      case .userCancelled:
+        return 0
```
deepesh-vasthimal-cko, line 26: is 0 a new error code ? Why not use 1000 series ?

alex-ioja-yang-cko: No specific reason than ensuring a clear distinction from other errors. User cancelling is not normally an error, but if we will deliver inside a Result outcome, then it will have to be an Error Type.

Given 0 is also a False Boolean value, it just felt like the best choice in my opinion.

deepesh-vasthimal-cko (approved):
Minor comment on the error code returned

Checkout/Checkout/Source/Tokenisation/Error/TokenisationError+TokenRequest.swift
deepesh-vasthimal-cko: The [current error codes](https://checkout.atlassian.net/wiki/spaces/FS/pages/4984537434/Checkout+-+Error+Codes) are XXXX format ( example 1000 and so on ) I feel we should create a new error code like 4000 as this is user enable action and follow similar convention for consistency.

alex-ioja-yang-cko: @ehab-al-cko raised the same point as me that User Cancelling is not really an error. But given our limitation of having to deliver this outcome in a Result format, we need to either include as a success (which it is not), or as a failure (which is attached to an error).

Having a distinct error code seems like a good compromise, delivering outcome whilst making it non-standard to the rest of the errors.

PR #368, opened 2022-12-22
Release 4.0.4: Pass UserCancelled in completion handler

alex-ioja-yang-cko:
## Proposed changes

By passing the cancellation as an error of `userCancelled`, the consumer can successfully track the occurrence of this event.

Will require product review and internal discussion over possible alternatives that may fit better into the product roadmap.

### DIFF
--- a/Checkout/Checkout/Source/Tokenisation/Error/TokenisationError+TokenRequest.swift
+++ b/Checkout/Checkout/Source/Tokenisation/Error/TokenisationError+TokenRequest.swift
@@ -11,6 +11,7 @@ extension TokenisationError {
   public enum TokenRequest: CheckoutError {
     case couldNotBuildURLForRequest
     case applePayTokenInvalid
+    case userCancelled
     case cardValidationError(ValidationError.Card)
     case networkError(NetworkError)
     case serverError(ServerError)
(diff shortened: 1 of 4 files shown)


Checkout/Checkout/Source/Tokenisation/Error/TokenisationError+TokenRequest.swift
```diff
@@ -21,6 +22,8 @@ extension TokenisationError {
         return 3001
       case .applePayTokenInvalid:
         return 1100
+      case .userCancelled:
+        return 0
```
deepesh-vasthimal-cko, line 26: Why 0 similar to other PR and not 1000 series ?

deepesh-vasthimal-cko (approved):
Minor comment on the error code.

Checkout/Checkout/Source/Tokenisation/Error/TokenisationError+TokenRequest.swift
alex-ioja-yang-cko: Discussed in other PR. As @ehab-al-cko similarly raised the point that userCancelled is not really an error, but we need to deliver it as outcome, I feel more confident in the current path (not an error, but delivered inside an error, so stands out from standard)

PR #369, opened 2022-12-29
Discussion: Simplify communication paths between Payment VC and its ViewModel

alex-ioja-yang-cko:
Most of the lines changed are within unit tests, where I have created new mock structures and supplemented existing testing!

**See following explanations to understand reasoning for PR**

 

## Problem description

Currently the ViewController holds a reference to the viewModel under 2 separate properties:

- `ViewModel` (direct clear reference)

- `Delegate` (used improperly)

The ViewModel then passes information back to the ViewController via `completion handlers` (/callbacks).

<br clear="left"/>
 

A few reasons why this is not great:
- architecturally the pattern is a mess. It uses multiple known elements and approaches in ways that are non standard
- 1 object consumes the child object under different variable names. This in itself causes many issues on scalability

 

 

## Solution

MVVM (Model - View - ViewModel) is a very straight forward pattern and to fix our current implementation we would be looking to drop the "delegate" from the current implementation.

This would leave our ViewController to own the ViewModel and the ViewModel to keep a way to communicate upwards in a child - parent relationship.

We now get clear object relationship that follows basic MVVM architecture.
<br clear="left"/>

 

 

## Improved Solution

The previous Solution is a natural step into fixing the shortcomings of current Problem, it also reduces fragility when considering any future work, but the flow can further be improved with a minor tweak.

Completion handlers create a 1:1 relationship, 1 property maps to 1 callback. This requires additional overhead (setup, references) and most importantly increases likelihood of memory related issues being introduced.

Delegate pattern fixes this and enables a 1:many relationship, 1 property maps to many possible callbacks (1 delegate can contain a number of functions), which reduces the need to have a complex setup where each property is assigned individually & enables scalability with no fear of new memory leaks being introduced (whilst scaling the particular relationship the delegate serves!).
<br clear="left"/>
 

 
 
 

## Changes
This PR:
- migrates from the current implementation (under Problem description) to Improved Solution
- removes some previously unused callbacks
- introduces `UIPresenter`. This allows the ViewModel to handle UI navigation without exposing UIKit (and with it UI library)
- fixes a minor unrelated SwiftLint warnings around `FramesBundle`
- migrates the billing summary formatting to a separate object which enables it being tested directly & fully (vs the current minor testing as part of the ViewModel)
- replaces the concrete use of `CardValidator` inside the ViewModel with its interface `CardValidating`
- updates unit tests in line to architectural changes & adds additional ones


alex-ioja-yang-cko:
No time to maintain this branch and lack of internal agreement to do anything with it. 👋

Source/UI/PaymentForm/ViewController/PaymentViewController.swift
```diff
@@ -167,158 +166,54 @@ final class PaymentViewController: UIViewController {
...
-            DispatchQueue.main.async {
-                self?.payButtonView.isLoading = self?.viewModel.isLoading ?? false
-            }
+    func loadingStateChanged() {
+        DispatchQueue.main.async { [weak self] in
+            self?.payButtonView.isLoading = self?.viewModel.isLoading ?? false
         }
     }
 
-  private func setupHeaderViewClosure() {
```
deepesh-vasthimal-cko, line 195: 👍

Source/UI/PaymentForm/ViewController/PaymentViewController.swift
```diff
@@ -167,158 +166,54 @@ final class PaymentViewController: UIViewController {
...
-    viewModel.shouldEnablePayButton = { [weak self] isEnabled in
-      self?.payButtonView.isEnabled = isEnabled
+    func refreshPayButtonState(isEnabled: Bool) {
+        DispatchQueue.main.async { [weak self] in
+            self?.payButtonView.isEnabled = isEnabled
+        }
     }
-  }
-
-  private func updateBackgroundViews() {
```
deepesh-vasthimal-cko, line 269: 👍

Source/UI/PaymentForm/ViewModel/DefaultPaymentViewModel.swift
```diff
@@ -57,46 +50,18 @@ class DefaultPaymentViewModel: PaymentViewModel {
         logger.log(.paymentFormCanceled)
     }
 
-    func updateAll() {
```
deepesh-vasthimal-cko, line 60: 👍

Source/UI/PaymentForm/ViewModel/PaymentViewModel.swift
```diff
@@ -1,30 +1,30 @@
...
-  var billingFormData: BillingForm? { get set }
+  var delegate: PaymentViewModelDelegate? { get }
   var paymentFormStyle: PaymentFormStyle? { get set }
   var billingFormStyle: BillingFormStyle? { get set }
   var supportedSchemes: [Card.Scheme] { get set }
-  var cardValidator: CardValidator { get set }
-  var logger: FramesEventLogging { get }
+  var cardValidator: CardValidating { get set }
   var isLoading: Bool { get set }
-  var updateLoading: (() -> Void)? { get set }
```
deepesh-vasthimal-cko, line 12: 👍

deepesh-vasthimal-cko:
> V-VM communication is depending on Notification. Notification mechanism for MVVM notifies without considering whatever objects receive notifications. Observables, NotificationCenter, closures, KVO satisfy above conditions. If delegation pattern is used in MVVM then it will depend on view then it will be like MVP. MVVM should not depend on View. Closures can be implemented outside without depending on owner that's one of the core implementation of MVVM. It will make sense for me if you rename to MVP. Screenshot 2023-01-04 at 4 20 17 pm

I had a review of the PR on a high level and looks like what @alex-ioja-yang-cko has managed to do is simpler version of MVVM using a delegate, This is still valid MVVM as we are aware we can do MVVM using Observers, Notifications for binding ViewModel to VC etc But we can also use delegate to perform the ViewBinding. I am happy to discuss the pros and cons in long terms as I am aware that more complex or scalable MVVM may have complex bindings ( example is using RxSwift ) using for different reason which may not applicable for our business use case now.

PR #373, opened 2023-01-05
Release 4.0.4: Bring latest release changes to main

alex-ioja-yang-cko:
## Proposed changes

Bring the missing changes from 4.0.4 release branch back into main:
- Changelog
- version update
- API key changes
- using String instead of enum for scheme returns


deepesh-vasthimal-cko (approved):

deepesh-vasthimal-cko:
Going forward we should try to release from main so we can may be avoid the housekeeping from other branches.

PR #384, opened 2023-05-09
PIMOB-1889: Phone Number Validation improvements

alex-ioja-yang-cko:
[Jira task](https://checkout.atlassian.net/browse/PIMOB-1889)

### Create singleton state for PhoneNumberValidator
- Performance improvement over recreating a validator on every use. As the form is presented & interacted with, the validator is being called many more times then when simply inputting text
- Enables the countryCode to be persisted in validator instance

 

### Connected Billing Form Country change update with Phone Validator
- Enabled actual feature to occur. 
    - When Country is modified, the phone validator will automatically use selected Country prefix & format if an international prefix is not already provided

 

### Housekeeping
- Removed Package resolved file from tracking which should reduce noise in commits & PRs. Xcode will still resolve SPM dependencies locally

 

⚠️ 
Failed to create unit tests. The current setup relies on UI components to be passed around and without compromising on core access control, they are unpractical to implement...
Have looked at improving this and decoupling validation from UI but unpractical to do in this PR 😞


Source/Extensions/PhoneExtensions.swift
```diff
@@ -22,9 +22,9 @@ extension Phone {
 
     func displayFormatted() -> String {
         if number?.starts(with: "+") == true {
-            return PhoneNumberValidator().formatForDisplay(text: number ?? "")
+            return PhoneNumberValidator.shared.formatForDisplay(text: number ?? "")
```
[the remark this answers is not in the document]
alex-ioja-yang-cko: That is a very good observation @ehab-al-cko . In that version it was not getting deallocated, but I have now updated it to ensure once the tokenisation is completed it does get removed.

Please have another look.

Tests/Core/Extensions/UIColor+PropertyProvidingTests.swift
```diff
@@ -19,7 +19,7 @@ final class UIColorPropertyProvidingTests: XCTestCase {
     func test_properties() {
         XCTAssertEqual(UIColor.systemBlue.properties, [.alpha: 1.0,
                                                        .blue: 1.0,
-                                                       .red: 0.0,
-                                                       .green: AnyCodable(122.0 / 255.0)])
+                                                       .red: AnyCodable(10.0 / 255.0),
+                                                       .green: AnyCodable(132.0 / 255.0)])
```
alex-ioja-yang-cko, line 23: We haven't had pipelines run for a while and these values ran behind to latest iOS version

Tests/UI/3DS/ThreedsWebViewControllerTests.swift
```diff
@@ -122,11 +122,6 @@ class ThreedsWebViewControllerTests: XCTestCase {
         _ = ThreedsWebViewController(nibName: nil, bundle: nil)
     }
 
-    func testInitializationCoder() {
-        let coder = NSKeyedUnarchiver(forReadingWith: Data())
-        _ = ThreedsWebViewController(coder: coder)
-    }
```
alex-ioja-yang-cko, line 128: Syntax changed for `NSKeyedUnarchiver` and its behaviour no longer allows it to create with empty data, throwing an error.

bitrise.yml
```diff
@@ -206,5 +206,5 @@ app:
     BITRISE_DISTRIBUTION_METHOD: development
 meta:
   bitrise.io:
-    stack: osx-xcode-13.4.x
-    machine_type_id: g2.4core
+    stack: osx-xcode-14.3.x-ventura
+    machine_type_id: g2-m1.4core
```
alex-ioja-yang-cko, line 210: had to update pipelines to require Xcode 14

Tests/UI/BillingForm/Validator/PhoneNumberValidatorTests.swift
```diff
@@ -212,4 +214,49 @@ class PhoneNumberValidatorTests: XCTestCase {
...
         XCTAssertEqual(validator.formatForDisplay(text: testNumber), "+86 21 8051 0459")
     }
+    
+    func testSingletonBehaviour() {
+        let testCountryCode = "GB"
+        
+        // GIVEN I need the validator singleton, I will receive a valid singleton to use
+        weak var originalFormatter = PhoneNumberValidator.shared
+        originalFormatter?.countryCode = testCountryCode
+        XCTAssertEqual(originalFormatter?.countryCode, testCountryCode)
```
okhan-okbay-cko, line 224: I think this assertion is redundant and will always pass

iOS Example Frame/iOS Example FrameUITests/FrameUITests.swift
```diff
@@ -4,7 +4,7 @@
 //
 //  Created by Alex Ioja-Yang on 22/08/2022.
 //  Copyright © 2022 Checkout. All rights reserved.
-//
+//
```
okhan-okbay-cko, line 7: this can be put out of this diff I guess

Tests/UI/BillingForm/Validator/PhoneNumberValidatorTests.swift
```diff
@@ -174,7 +176,7 @@ class PhoneNumberValidatorTests: XCTestCase {
         
         XCTAssertTrue(validator.shouldAccept(text: testNumber))
         XCTAssertTrue(validator.isValid(text: testNumber))
-        XCTAssertEqual(validator.formatForDisplay(text: testNumber), "+64 9-888 8625")
+        XCTAssertEqual(validator.formatForDisplay(text: testNumber), "+64 9 888 8625")
```
okhan-okbay-cko, line 179: is this intentional?

Tests/UI/3DS/ThreedsWebViewControllerTests.swift
okhan-okbay-cko: just an uneducated guess: does it accept if we create a dummy data from a dummy string?

alex-ioja-yang-cko: Honestly I'm not sure. Don't know why it was tested either, as init with coder as far as I know is triggered when loading via Storyboard 🤔, so the test didn't really make sense to begin with 🤷‍♂️.

Source/UI/Validator/PhoneNumberValidator.swift
```diff
@@ -29,11 +35,32 @@ class PhoneNumberValidator: Validator {
...
+// MARK: Singleton behaviour
+extension PhoneNumberValidator {
+
+    static var shared: PhoneNumberValidator {
+        if let instance {
+            return instance
+        }
+        let newValidator = PhoneNumberValidator()
+        instance = newValidator
+        return newValidator
```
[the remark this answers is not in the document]
alex-ioja-yang-cko: That would be nice, but `instance` is optional so it can be deallocated. 

Even though we just assign it the line before, compiler wouldn't know its been freshly created, it knows it may be nil. So unless we force unwrap and trigger the wrath of SwiftLint whilst risking future changes that will leave a bang operator hanging around, your suggestion would not work 😉

Source/UI/Validator/PhoneNumberValidator.swift
```diff
@@ -29,11 +35,32 @@ class PhoneNumberValidator: Validator {
...
+    }
+    static private var instance: PhoneNumberValidator?
+
+    /// Will deallocate the singleton source
+    static func removeSingleton() {
+        Self.instance = nil
```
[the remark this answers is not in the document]
alex-ioja-yang-cko: Fair call, it's not really needed as they are type level. At first glance I'd generally be explicit when calling type properties & methods, but happy to also remove `Self` if you insist 😛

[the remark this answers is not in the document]
okhan-okbay-cko: no I don't insist :)

Tests/UI/3DS/ThreedsWebViewControllerTests.swift
okhan-okbay-cko: I'd personally hesitate to remove a test because there might be a scenario that's hard to catch and that might've been handling that for a while but still don't mind it's getting deleted if you're confident :)

Source/UI/Validator/PhoneNumberValidator.swift
```diff
@@ -29,11 +35,32 @@ class PhoneNumberValidator: Validator {
...
+
+    static var shared: PhoneNumberValidator {
+        if let instance {
+            return instance
+        }
+        let newValidator = PhoneNumberValidator()
+        instance = newValidator
+        return newValidator
+    }
+    static private var instance: PhoneNumberValidator?
```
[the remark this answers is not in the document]
alex-ioja-yang-cko: I accept that computed properties can be messy and look less like a variable declaration and more like a function, but in this case they are strictly organised by access level and role:
- internal property
- private property
- internal method

I'd rather be consistent with formatting, so other alternative may be to make `shared` a function, but then we'll invoke it as `PhoneNumberValidator.shared().___` everywhere, which will trigger flashbacks of consuming Obj-C singletons in Swift.

We may also break apart the logic from the dynamic getter into a private function, making it look cleaner:
```Swift

    static var shared: PhoneNumberValidator {
        Self.getShared()
    }
    static private var instance: PhoneNumberValidator?

    /// Will deallocate the singleton source
    static func removeSingleton() {
        instance = nil
    }

    private static getShared() -> PhoneNumberValidator {
        if let instance {
            return instance
        }
        let newValidator = PhoneNumberValidator()
        instance = newValidator
        return newValidator
    }
```
I would however argue that this last version needlessly splits logic into "another function" and just adds noise.

So personally I'd suggest keeping it as is, since it continues to structure files consistently with project, but happy to discuss further if you feel another path may be beneficial.

Tests/UI/BillingForm/Validator/PhoneNumberValidatorTests.swift
alex-ioja-yang-cko: It prevents hidden passing for the final assertion.

If the object was not correctly deallocated and simply "appeared to not be available", the final part
```Swift
let newFormatter = PhoneNumberValidator.shared
XCTAssertNil(originalFormatter)
XCTAssertNotEqual(newFormatter.countryCode, testCountryCode)
``` 
may simply fail because the country code was never assigned correctly, whilst the object being identical to the first one.

So ensuring the property assignment behaves as expected, allows us to conclude in the final test that if that last assertion is failing, is only due to the original object being brought back to life instead of a new one (which never had the code assigned to begin with).

iOS Example Frame/iOS Example FrameUITests/FrameUITests.swift
alex-ioja-yang-cko: GitHub was unable to load commits today after it had the issues, so had to push something. Will undo 👍

Tests/UI/BillingForm/Validator/PhoneNumberValidatorTests.swift
alex-ioja-yang-cko: This test is dependent on the formatter, which is dependent on PhoneNumberKit. Since we have unblocked the strict linking to version 3.3.3, it would appear the later versions have changed the string representation of this country's phone number, which in turn invalidated our unit tests.

So it is intentional as I want the unit tests to pass, but I did not intentionally make changes to cause it (beyond unblocking PhoneNumberKit dependency).

alex-ioja-yang-cko:
Cannot be merged into `main` due to work done since last release. As we haven't decided how to support that work, this work will need to be handled in a release branch instead. As there are some conflicts, will close this for now and reopen

PR #386, opened 2023-05-10
Housekeeping Release: Update configurations

alex-ioja-yang-cko:
Syncing [Housekeeping change PR](https://github.com/checkout/frames-ios/pull/385) from `main` into the release branch used where the next work is taking place.


Checkout.podspec
```diff
@@ -12,11 +12,11 @@ TODO: Add long description of the pod here.
   s.author = { "Checkout.com Integration" => "integration@checkout.com" }
   s.source       = { :git => "https://github.com/checkout/frames-ios.git", :tag => s.version }
 
-  s.ios.deployment_target = '10.0'
-  s.swift_version = "5.0"
+  s.ios.deployment_target = '11.0'
```
okhan-okbay-cko, line 15: shall we do the same here as discussed in the other PR?

PR #388, opened 2023-05-10
PIMOB-1888: Create Accessibility Identifiers and use in UITests

alex-ioja-yang-cko:
[Jira task](https://checkout.atlassian.net/browse/PIMOB-1888)

`AccessibilityIdentifiers` are public, allowing an easier and more reliable integration inside consumers UITests of the SDK behaviour.
UITests are reenabled as this improvement, alongside the phone number validation, should make our own UITests more reliable.


Source/UI/BillingForm/BillingFormCell.swift
```diff
@@ -51,6 +51,19 @@ import Foundation
...
+        case .fullName: return AccessibilityIdentifiers.BillingForm.cardholder
+        case .addressLine1: return AccessibilityIdentifiers.BillingForm.addressLine1
+        case .addressLine2: return AccessibilityIdentifiers.BillingForm.addressLine2
+        case .city: return AccessibilityIdentifiers.BillingForm.city
+        case .state: return AccessibilityIdentifiers.BillingForm.state
+        case .postcode: return AccessibilityIdentifiers.BillingForm.postcode
+        case .country: return ""
+        case .phoneNumber: return AccessibilityIdentifiers.BillingForm.phoneNumber
+        }
+    }
```
okhan-okbay-cko, line 65: I think a typealias can come in handy here

okhan-okbay-cko (approved):
🦖

Source/UI/BillingForm/BillingFormCell.swift
alex-ioja-yang-cko: I don't understand, can you please provide an example? 
And where would it be used, just within `BillingFormCell`, or around the `AccessibilityIdentifiers` enum?

okhan-okbay-cko: I mean like `typealias Identifier = AccessibilityIdentifiers.BillingForm`

PR #390, opened 2023-05-11
PIMOB-1888: Use Accessibility Identifiers on UITests

alex-ioja-yang-cko:
[Jira task](https://checkout.atlassian.net/browse/PIMOB-1888)

Declare and assign public AccessibilityIdentifiers. These can be used for more reliable and readable UITests.

Minor updates:
- phone number validation being eager when country was being set. If the field has no value, it should not attempt to revalidate it
- iOS 16 keyboard support to find delete button


Source/Core/Constants/AccessibilityIdentifiers.swift
```diff
@@ -0,0 +1,29 @@
...
+public enum AccessibilityIdentifiers {
+
+    public enum PaymentForm {
+        static public let cardholder = "CardholderInput"
+        static public let cardNumber = "CardNumberInput"
+        static public let cardExpiry = "ExpiryDateInput"
+        static public let cardSecurityCode = "CardSecurityCodeInput"
+    }
+
+    enum BillingForm {
```
deepesh-vasthimal-cko, line 19: Is this not meant to be Public ?

alex-ioja-yang-cko: 🤦‍♂️ Good catch 👍

deepesh-vasthimal-cko (approved):

PR #393, opened 2023-05-17
PIMOB-1883: Modify approach to formatting the expiry date

alex-ioja-yang-cko:
[Jira task](https://checkout.atlassian.net/browse/PIMOB-1883)

---

- Create `ExpiryDateFormatter` to:
    - format user inputted strings into formatted dates & raise errors if needed
    - when formatting string, do not expect the modification to be done via the range and parse whatever state string is in at the formatting time
    - parse input (expected already formatted, but not necessarily) into ExpiryDate object
- Modify ExpiryDateView to consume the new formatter instead

---

Validation of correct impact on UITests:
- [Intel runner pipeline](https://app.bitrise.io/build/f541f48b-2560-4900-bd7d-5d66c8d15adc) 👌 
- [Apple silicon runner pipeline](https://app.bitrise.io/build/6855387c-c228-4b8a-a07d-ef7e1f51be0d) 👌


Source/UI/PaymentForm/View/ExpiryDate/ExpiryDateFormatter.swift
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    /// Propose a display formatted expiry date that may be shown to the user
+    /// If the cardScheme is `unknown`, this validates that the cvv is conforming to internal generic standards
+    /// - Parameters:
+    ///   - input: The input as received from the user facing UI component
+    /// - Returns: Tuple made of a formatted input for displaying to the user and an error if presented input is not valid.
+    /// If displayString is nil, the text in the UI component should be rejected at last valid input should be maintained!
+    func formatForDisplay(input: String) -> (displayString: String?, validationError: ValidationError.ExpiryDate?) {
```
deepesh-vasthimal-cko, line 45: @alex-ioja-yang-cko Has all of this logic been reused or added newly? I'm just asking so we can perform testing as required before release.

Source/UI/PaymentForm/View/ExpiryDate/ExpiryDateFormatter.swift
```diff
@@ -0,0 +1,114 @@
...
+import Checkout
+
+struct ExpiryDateFormatter {
+
+    let separator: String
+    let dateFormatTextCount: Int
+    private let cardValidator: CardValidating
+
+    init(componentSeparator: String = "/",
+         cardValidator: CardValidating = CardValidator(environment: .sandbox)) {
```
deepesh-vasthimal-cko, line 17: Can you remind me why we hardcod CardValidator to use .sandbox ? also would be worth adding a comment.

Source/UI/PaymentForm/View/ExpiryDateView.swift
```diff
@@ -1,201 +0,0 @@
-import UIKit
```
deepesh-vasthimal-cko, line 1: @alex-ioja-yang-cko Do you know which line in the older logic was causing the issue on Intel machines? would be interested to know if this is a swift API issue and how it behaves differently in different machines as the new approach still uses all the TextFieldViewDelegate methods which I assume is required.

alex-ioja-yang-cko: [This](https://github.com/checkout/frames-ios/pull/393/files#diff-5a7fe3d7bbb6ad57fd1dd9577704228acf52cfae0bc0ed52e0747328a99e606cL97-L105) (might not open correctly, so after you open link in a new tab, make sure to again expand the deleted file).

***

It was assuming a linear input occurring, where we always only modify the last character in the String (appending or removing). This was already prone to invalid behaviour if cutting/pasting or moving the cursor to somewhere else in the String. But the difference in behaviour seems to be from the values of `Range` and the way the code was triggered 2 times for every input on the runners we could reproduce. 

***
The flow for that code on a normal behaviour was:
- input change => modification reported with a Range (index of change, length 0)

The flow when the error was encountered (could only reproduce on Intel runners, hence the assumption it may be related 🤷‍♂️) was:
- input change
        => modification reported with a Range (index of change, length 0) - as before
        => modification reported with a Range (index of change, length 1) - unexpected ❗ . Here, our code was assuming any length present in the range was a deletion of previous content, and since it expected incremental updates, it would run into a "character deletion routine" where the `/` was removed. And then following it by simply adding the next character. This resulted in the index skipping the number where it should add separator.

Source/UI/PaymentForm/View/ExpiryDate/ExpiryDateFormatter.swift
alex-ioja-yang-cko: I've actually followed TDD whilst building this component, with tests being written before I even wrote the logic for how to solve it. On completion of the original version, as I integrated it and eventually manually tested, found less obvious edge cases which have also had additional unit tests added. So overall I am pretty confident it's a solid helper object for our UI. Full commit history is [on this branch](https://github.com/checkout/frames-ios/compare/release/4.0.5...alex/pimob-1883-refactor-uitest), but due to the noise in refactoring to even enable understanding the problem, I have copied the final version from those objects (hence fewer and cleaner commit history here - @okhan-okbay-cko should be happy with this 😬)

alex-ioja-yang-cko: But yes, it would be good to do some basic regression testing before an actual release. I am just reassured by the fact my keyboard has tried inputting more varied content and copied/pasted in formats that I do not know if you're even able to reproduce when smoke testing 😂

Source/UI/PaymentForm/View/ExpiryDate/ExpiryDateFormatter.swift
alex-ioja-yang-cko: We don't actually hardcode the CardValidator, this is just the default parameter to avoid **having to inject it**. In practice, we only create this object once, in ExpiryDateView and I don't envision another place to consume it.

Inside `ExpiryDateView` it gets created as:
```Swift
  init(cardValidator: CardValidator) {
    self.dateFormatter = ExpiryDateFormatter(cardValidator: cardValidator)
```

deepesh-vasthimal-cko: I meant why have you hardcoded `.sandbox` inside CardValidator. Is that used for logging? if then what happens when our SDK is used in production?

deepesh-vasthimal-cko (approved):
Looks good to me, Did a quick test on the simulator.

Source/UI/PaymentForm/View/ExpiryDate/ExpiryDateFormatter.swift
alex-ioja-yang-cko: That is the default argument to the initialiser. If you create an object `ExpiryDateFormatter` and do not provide a parameter for `cardValidator`, it will fallback to this one.
In practice, we only use this object in 1 place (ExpiryDateView) and we do not rely on this default parameter, we inject a parameter.

The reason I provided the default parameter has been the unit testing, where having extensive tests that don't all require injecting a parameter, has allowed to save some space and hopefully keep those tests more concise.

So we do not "hardcode" the environment into the CardValidator. As to why that is injected in the first place, I have now looked into it and it does appear to be for logging. But this Validator has been inherited from 3.x versions and seems to interact with a singleton logger. That is however not the same logger as the one we use for frames, it's an obsolete one. We may add something to backlog to remove that logger altogether if that is not something we need, our codebase would be grateful for it.

Source/Extensions/UITextFieldExtension.swift
```diff
@@ -2,6 +2,15 @@ import UIKit
 
 extension UITextField {
 
+    /// String generated after the range and the modification proposed are applied to current text
+    func newTextAfter(applyingString newString: String, at range: NSRange) -> String? {
```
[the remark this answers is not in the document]
alex-ioja-yang-cko: Interesting suggestion. I have renamed it to resemble the underlying String API. Please have another look

PR #395, opened 2023-05-19
PIMOB-XXXX: Update Package.swift dependencies

alex-ioja-yang-cko:
Enforce SPM requirements to consume valid dependencies for the functionality


deepesh-vasthimal-cko (approved):
Thanks for fixing this.

PR #405, opened 2023-06-09
PIMOB-1972: Add new error type for missing API Key

alex-ioja-yang-cko:
[PIMOB-1972](https://checkout.atlassian.net/browse/PIMOB-1972)

Create new error and return when tokenisation is attempted and an API Key does not exist.

❌ Change **does not** perform validation on the key! It simply validates that a value is provided
✔️ New error will get reported within the current analytics flows

*** 

🔴 This is an API breaking change, but is also auto corrected by using Xcode. The break only appears if consumer application is using a switch statement to iterate via errors returned by SDK. Xcode can however autoupdate the syntax to include the new `case`, making it a minor hurdle to overcome.

🟡 We may reduce impact of the breaking change if we move the error within `NetworkError`, where it will only break if the consumer application uses a switch statement on the `TokenRequestError.NetworkError` (1 layer inside the current proposed layer). 
This would however place error inside the wrong layer, as the validation is done locally and it is not technically a NetworkError. Also, it will still break the API in the exact same way, but for a smaller subset.

[PIMOB-1972]: https://checkout.atlassian.net/browse/PIMOB-1972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

*** 

Sample application example

Simulator Screenshot - iPhone 14 Pro Max - 2023-06-09 at 10 13 20

### DIFF
--- a/iOS Example Frame/iOS Example Frame/ViewController/HomeViewController.swift
+++ b/iOS Example Frame/iOS Example Frame/ViewController/HomeViewController.swift
@@ -137,6 +137,8 @@ class HomeViewController: UIViewController {
                 showAlert(with: "Error code: \(serverError.code)", title: "Server Error")
             case .couldNotBuildURLForRequest:
                 showAlert(with: "Error code: \(failure.code)", title: "Could Not Build URL")
+            case .missingAPIKey:
+                showAlert(with: "You need to make sure an API key is present", title: "Missing API Key")
             }
         case .success(let tokenDetails):
             showAlert(with: tokenDetails.token, title: "Success")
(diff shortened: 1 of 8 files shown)


iOS Example Frame/iOS Example Frame/ViewController/HomeViewController.swift
```diff
@@ -137,6 +137,8 @@ class HomeViewController: UIViewController {
                 showAlert(with: "Error code: \(serverError.code)", title: "Server Error")
             case .couldNotBuildURLForRequest:
                 showAlert(with: "Error code: \(failure.code)", title: "Could Not Build URL")
+            case .missingAPIKey:
+                showAlert(with: "You need to make sure an API key is present", title: "Missing API Key")
```
alex-ioja-yang-cko, line 141: This is the kind of API Breaking change consumers will face. Our `switch failure {` will not compile as a new case is introduced. Xcode can however auto-fix the code, leave a placeholder for the new error to be handled (ie: consumer will add something like `showAlert` to handle error)

deepesh-vasthimal-cko (approved):

PR #408, opened 2023-06-22
Apply changes to reduce the code change

okhan-okbay-cko:
Suggestion to edit the code to keep the change to the minimum


okhan-okbay-cko:
This PR would remove some code we had before and could have created a breaking change. Hence closing.

PR #409, opened 2023-06-23
PIMOB-1987: Add new style to test app

alex-ioja-yang-cko:
[Jira task](https://checkout.atlassian.net/browse/PIMOB-1987)

Add new style using Theme to showcase the border design. Changes required:
- `DefaultBorderStyle` public initialiser and make public. This is in line with other Default Style elements, all of which are public and is required in order to support customisation
- update `corners` default value in Theme related initialisers. This is required in order to ensure cornerRadius is applied as expected and previous implementation would have resulted in buggy behaviour
- update to Storyboard
- new `ThemeDemo+Border` focussing on the new theme style being added


Source/UI/BillingForm/Default Implementation/Elements/DefaultBorderStyle.swift
```diff
@@ -1,11 +1,27 @@
...
-    var edges: UIRectEdge? = .all
-    var corners: UIRectCorner? = .allCorners
+public struct DefaultBorderStyle: ElementBorderStyle {
+    public var cornerRadius: CGFloat = Constants.Style.BorderStyle.cornerRadius
+    public var borderWidth: CGFloat = Constants.Style.BorderStyle.borderWidth
+    public var normalColor: UIColor = FramesUIStyle.Color.borderPrimary
+    public var focusColor: UIColor = FramesUIStyle.Color.borderActive
+    public var errorColor: UIColor = FramesUIStyle.Color.borderError
+    public var edges: UIRectEdge? = .all
+    public var corners: UIRectCorner? = .allCorners
```
deepesh-vasthimal-cko, line 10: Do we need to update any UI tests after adding the support for borders?

deepesh-vasthimal-cko (changes requested):
Just one comment on UI tests. otherwise looks good.

Source/UI/BillingForm/Default Implementation/Elements/DefaultBorderStyle.swift
alex-ioja-yang-cko: UI tests validate that a component exists and is accessible, as if you were to tap it.

I think you are referring to snapshot tests, which we have not implemented and are currently not using in our SDK or sample application.

PR #421, opened 2023-07-10
Update/test checkout

okhan-okbay-cko:
This is a huge PR but it was necessary for this to go all in at once and most of the changes are folder path updates. 

- Removes some unused files
- Removes an additional cascaded layer of `Checkout/Checkout` into `Checkout` only
- Runs tests under `Checkout/Source` as part of the Bitrise workflows
- Makes all the sample projects work and the most importantly, all the samples now mimic the merchant behaviour with no local drag and dropping or referencing of the SDKs (i.e. they fetch the code from remote sources as our merchants do)
- `Package.swift` now denotes the package structure and to be used as a base for developments from now on


deepesh-vasthimal-cko:
The Bitrise YML file might need an update as the build is failing with below error
`Checkout/Samples/CocoapodsSample/CheckoutCocoapodsSample.xcworkspace is not exist`

Checkout/Checkout.podspec
```diff
@@ -0,0 +1,17 @@
+Pod::Spec.new do |s|
+  s.name         = "Checkout"
+  s.summary      = "Checkout Core Features that can only be used with the right level of PCI Compliance. Please contact us for the details."
```
deepesh-vasthimal-cko, line 3: @patrick-hoban-cko Can you review this text? if already done please ignore.

screenshots/Package.swift
```diff
@@ -0,0 +1,7 @@
+// swift-tools-version:5.5
+
+import PackageDescription
```
deepesh-vasthimal-cko, line 3: why do the screenshots folder have `Package.swift`?

scripts/Package.swift
```diff
@@ -0,0 +1,7 @@
+// swift-tools-version:5.5
+
+import PackageDescription
```
deepesh-vasthimal-cko, line 3: why does the scripts folder have Package.swift?

screenshots/Package.swift
okhan-okbay-cko: As mentioned in the PR description, it's to slim down and tidy up the distribution package. So, whenever an empty `Package.swift` file gets added to a folder like this, it excludes that folder from the SPM distribution. 

Merchants don't need to download our sample apps or screenshots or scripts to use our SDK. This is the only working way currently.

scripts/Package.swift
okhan-okbay-cko: same as above

screenshots/Package.swift
okhan-okbay-cko: I removed these to avoid future confusion from our merchants' side

Checkout/Checkout.podspec
patrick-hoban-cko: Checkout Core functionality can only be used with the right level of PCI Compliance. Please contact us for more details.

deepesh-vasthimal-cko (approved):

Checkout/Checkout.podspec
```diff
@@ -0,0 +1,17 @@
...
+  s.name         = "Checkout"
+  s.summary      = "Checkout Core functionality can only be used with the right level of PCI Compliance. Please contact us for more details."
+  s.version      = "4.1.0"
+  s.homepage     = "https://github.com/checkout/frames-ios"
+  s.swift_version = "5.0"
+  s.author       = { "Checkout.com Integration" => "integration@checkout.com" }
+  s.platform     = :ios, "12.0"
+  s.source       = { :path => "./Source" }
+  s.license      = "MIT"
+
```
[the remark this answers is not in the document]
okhan-okbay-cko: it's not deprecated but we'll continue releasing new versions of it. We can revisit to update this file when we need a new version. Probably we'd get some error/warnings in Terminal and just quickly update it

PR #429, opened 2023-08-01
PIMOB 2058 - expiry date validation

okhan-okbay-cko:
`CardValidator` was solid and was returning the error but `ExpiryDateFormatter` was not using that error information and also was overriding any kinds of errors that was being propagated through. 

This PR uses a central point of verification for expiry date validation and depends on CardValidator only.

### DIFF
--- a/Source/UI/PaymentForm/View/ExpiryDate/ExpiryDateView.swift
+++ b/Source/UI/PaymentForm/View/ExpiryDate/ExpiryDateView.swift
@@ -1,12 +1,8 @@
 import UIKit
 import Checkout
 
-enum ExpiryDateError: Error {
-  case invalidCode
-}
-
 protocol ExpiryDateViewDelegate: AnyObject {
-  func update(result: Result<ExpiryDate, ExpiryDateError>)
+  func update(result: Result<ExpiryDate, ValidationError.ExpiryDate>)
 }
 
 final class ExpiryDateView: UIView {
@@ -59,7 +55,7 @@ extension ExpiryDateView: TextFieldViewDelegate {
     guard textField.text?.count ?? 0 != dateFormatter.dateFormatTextCount else {
       return true
     }
-    let (displayValue, _) = dateFormatter.formatForDisplay(input: textField.text ?? "")
+    let displayValue = dateFormatter.formatForDisplay(input: textField.text ?? "")
     updateErrorViewStyle(isHidden: false,
                          textfieldText: displayValue,
                          error: .incompleteYear)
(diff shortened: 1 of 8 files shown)


Source/UI/PaymentForm/View/ExpiryDate/ExpiryDateView.swift
```diff
@@ -1,12 +1,8 @@
 import UIKit
 import Checkout
 
-enum ExpiryDateError: Error {
-  case invalidCode
-}
-
 protocol ExpiryDateViewDelegate: AnyObject {
-  func update(result: Result<ExpiryDate, ExpiryDateError>)
+  func update(result: Result<ExpiryDate, ValidationError.ExpiryDate>)
```
okhan-okbay-cko, line 5: Would this cause a breaking change? @ehab-al-cko @deepesh-vasthimal-cko

okhan-okbay-cko: I think it won't since it's not public. Anyway, double checking.

Tests/UI/BillingForm/View/ExpiryDateFormatterTests.swift
```diff
@@ -29,184 +29,166 @@ final class ExpiryDateFormatterTests: XCTestCase {
...
         let formatter = ExpiryDateFormatter()
-        let output = formatter.formatForDisplay(input: testInput)
+        let displayString = formatter.formatForDisplay(input: testInput)
 
-        XCTAssertEqual(output.displayString, "11/88")
-        XCTAssertNil(output.validationError)
+        XCTAssertEqual(displayString, "11/88")
     }
 
     func testIncompleteMonthAndFullYearInput() {
```
deepesh-vasthimal-cko, line 133: Is there an updated test case to test the Current Year + Previous Months of the Current year? If so can you please point me to that.

deepesh-vasthimal-cko (approved):
Looks good to me just one comment. @ehab-al-cko This may require some testing from your side so we have not accidentally broken anything which was previously working.

Tests/UI/BillingForm/View/ExpiryDateFormatterTests.swift
okhan-okbay-cko: There is no necessity for such a test case, since this is the formatter and no longer does the date validation. That test will reside in the next PR (`CardValidatorTests`) and it's been working from the beginning. So, there is no change in this PR to make the expiry date validation work, but, only to make the code flow through the required code pieces. 

Which means, `ExpiryDateFormatter` was doing the validation check, that it shouldn't have been doing from the beginning. And the expiry date validation logic was broken because of that. Now, after it stopped validating it but only kept formatting it, there must be no case that makes difference in its test cases. I mean, `11/88` and `05/23` makes no difference since those two inputs are in the same format. Adding a test case for it will be redundant. @deepesh-vasthimal-cko

PR #431, opened 2023-08-01
Add expiry date snapshot tests

okhan-okbay-cko:
First example of snapshot tests. More will be added.

It uses [this tool](https://github.com/pointfreeco/swift-snapshot-testing). 

**Must be run on iPhone 14 Pro**


iOS Example Frame SPM/iOS Example Frame Regression Tests/CardValidationSnapshotTests.swift
```diff
@@ -0,0 +1,35 @@
...
+
+    func testExpiryDateFormat_Failure() {
+        let app = XCUIApplication()
+        app.launchFrames()
+
+        let expiryTextField = app.otherElements[AccessibilityIdentifiers.PaymentForm.cardExpiry]
+        app.enterText("521", into: expiryTextField)
+        app.tapDoneButton()
+
+        assertSnapshot(matching: expiryTextField.screenshot().image, as: .image, record: false)
```
deepesh-vasthimal-cko, line 33: Do we not get back the "Expiry Date in the past error" at the code level to add an automation test without using a snapshot ? . Just trying to understand what was the purpose to use a snapshot in this case looking at the screen shot below.

okhan-okbay-cko: We do get it, but it's only a verification from all the `staticTexts` available on that screen. 

The success snapshot in this PR tests the expiry date formatting `MM/YY` and the failure case tests if we're showing the right error and if the error is shown below the expiry date field.

deepesh-vasthimal-cko (dismissed):

deepesh-vasthimal-cko:
I ran this locally. I can see a few test cases failing. Also when running the test it opens up 3 copy of simulators, Is that expected? Please review my SS. 
Screenshot 2023-08-07 at 15 21 43
Screenshot 2023-08-07 at 15 22 05

okhan-okbay-cko:
my bad, can you please run all the regression tests on iPhone 14 Pro or they'll fail @deepesh-vasthimal-cko

deepesh-vasthimal-cko:
> my bad, can you please run all the regression tests on iPhone 14 Pro or they'll fail @deepesh-vasthimal-cko

Still fails both on iPhone 14 Pro  and iPhone 13 pro locally for me. 
Screenshot 2023-08-07 at 15 22 05
Screenshot 2023-08-07 at 17 04 06
Screenshot 2023-08-07 at 17 04 30

iOS Example Frame SPM/iOS Example Frame SPM.xcodeproj/xcshareddata/xcschemes/Regression Tests.xcscheme
```diff
@@ -14,8 +14,7 @@
       shouldAutocreateTestPlan = "YES">
       <Testables>
          <TestableReference
-            skipped = "NO"
-            parallelizable = "YES">
+            skipped = "NO">
```
deepesh-vasthimal-cko, line 17: Was this causing multiple simulators to fire up ?. If we disable it the test might take too long to run. Is that what we want ?

okhan-okbay-cko: just testing if that was one of the causes of failure since Bitrise was failing as well

PR #434, opened 2023-08-03
Add validations for different language error texts

okhan-okbay-cko:
Adding snapshots to cover error messages' correctness in different languages.

**Must be run on iPhone 14 Pro**


deepesh-vasthimal-cko (approved):
LGTM these are really powerful snapshot tests for testing localisation. This would definitely save lot of manual effort. 

The only thing to consider in future PRs is we make sure we have unit tests, UI tests and Snapshot tests as part of the future PRs otherwise we will have a tech backlog created for ourselves.

PR #436, opened 2023-08-03
Exclude sample projects from package distribution

okhan-okbay-cko:
Screenshot 2023-08-03 at 21 25 04

Merchants don't need to download our sample projects into their application bundle. Especially after the addition of many snapshots. This PR removes them. The only working way of achieving this is adding empty `Package.swift` files as the other ways don't work.

### DIFF
--- a/iOS Example Frame/Package.swift
+++ b/iOS Example Frame/Package.swift
@@ -1,0 +1,5 @@
+// swift-tools-version:5.6
+// THIS IS TO EXCLUDE SAMPLE PROJECTS FROM PACKAGE DISTRIBUTION
+import PackageDescription
+let package = Package()
+// THIS IS TO EXCLUDE SAMPLE PROJECTS FROM PACKAGE DISTRIBUTION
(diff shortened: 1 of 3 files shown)


iOS Example Frame/Package.swift
```diff
@@ -0,0 +1,5 @@
+// swift-tools-version:5.6
+// THIS IS TO EXCLUDE SAMPLE PROJECTS FROM PACKAGE DISTRIBUTION
+import PackageDescription
+let package = Package()
```
deepesh-vasthimal-cko, line 4: How does just creating a Package instance EXCLUDE SAMPLE PROJECTS FROM PACKAGE DISTRIBUTION ?

okhan-okbay-cko: Seems like this is the only solution. I don't know the `how` part 🙂  @deepesh-vasthimal-cko

okhan-okbay-cko: Whatever the folder this empty package manifest is added, it gets excluded from the distribution. Was this your question?

deepesh-vasthimal-cko: My question was how does declare just an instance of the package and not using it help us in EXCLUDE SAMPLE PROJECTS FROM PACKAGE DISTRIBUTION

`let package = Package()`

okhan-okbay-cko: I think SPM excludes it when we do it this way but can't know the inner logic. Tested and it works.

deepesh-vasthimal-cko (approved):

PR #437, opened 2023-08-08
Fix bitrise yml

okhan-okbay-cko:
We had renamed integration tests to regression tests, therefore, we needed to update the name of the workflow on main branch since Bitrise gets the config from main.

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -170,20 +170,20 @@ workflows:
     before_run:
     - _frames-set-up
 
-  _run-integration-tests:
+  _run-regression-tests:
     steps:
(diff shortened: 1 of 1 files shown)


bitrise.yml
```diff
@@ -170,20 +170,20 @@ workflows:
     before_run:
     - _frames-set-up
 
-  _run-integration-tests:
+  _run-regression-tests:
```
deepesh-vasthimal-cko, line 173: Is this configured to run each commit ?

deepesh-vasthimal-cko (approved):

PR #442, opened 2023-08-10
Add expiry date edge case tests

okhan-okbay-cko:
Covering expiry date edge cases.


iOS Example Frame SPM/iOS Example Frame Regression Tests/Helpers/StaticTexts.swift
```diff
@@ -10,5 +10,7 @@ import Foundation
 
 enum StaticTexts {
     static let cardNumberNotValidError = "Please enter a valid card number"
+    static let expiryDateIsInThePastError = "Expiry date is in the past"
```
deepesh-vasthimal-cko, line 13: Are these static text not part of the SDK itself?

deepesh-vasthimal-cko (approved):

PR #460, opened 2023-09-01
Update README.md

ehab-al-cko:
Add checkout.com logo to README

[PREVIEW HERE](https://github.com/checkout/frames-ios/blob/454ab5c7d3d97d8a6ac542d1dd9ac6f294ee7b9e/README.md)


okhan-okbay-cko (commented):
I think we can assume this as a `hotfix`, so, it can directly go into `main`. Alongside that, the image URL has 3DS SDK reference in it. I believe it'd be better if we don't do that and source it from another place.

okhan-okbay-cko (approved):

okhan-okbay-cko (approved):

okhan-okbay-cko (dismissed):

okhan-okbay-cko (approved):

PR #469, opened 2023-09-08
Add SwiftLint as a blocker pre-check for all the other checks

okhan-okbay-cko:
- SwiftLint comes pre-installed on runner images of GitHub.
- SwiftLint will run before all the other checks. 
- If SwiftLint fails, then all the other checks don't run. This is made possible by the `needs:` mark.
- SwiftLint only runs for the new files to avoid a big PR
- It also parses errors and warnings for the new files and shows them on GitHub UI: (even though we won't get any warnings and will convert them to errors for new files. This is made possible by `--strict` flag. Parsing is made by using `sed -E`)

Screenshot 2023-09-11 at 12 45 17

- If there are no changes in `.swift` files, then it skips linting and marks the job as a success.

Screenshot 2023-09-12 at 16 28 39

- If there are changes in `.swift` files, then SwiftLint only lints them: 

Screenshot 2023-09-12 at 16 31 37

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,11 +1,16 @@
 included:
-    - Source
-    - iOS Example Frame
-    - iOS Example Frame Carthage
-    - iOS Example Frame SPM
+    - "Checkout"
+    - "Source"
+    - "iOS Example Frame"
+    - "iOS Example Frame SPM"
     
 excluded:
-    - iOS Example Frame/Pods
+    - "iOS Example Frame/Pods"
+    - "Checkout/Samples/CocoapodsSample/Pods"
+    - "CheckoutTests"
+    - "Tests"
+    - "iOS Example Frame SPM/iOS Example Frame Regression Tests"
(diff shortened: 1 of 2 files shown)


.swiftlint.yml
```diff
@@ -1,11 +1,16 @@
...
+    - "iOS Example Frame SPM"
     
 excluded:
-    - iOS Example Frame/Pods
+    - "iOS Example Frame/Pods"
+    - "Checkout/Samples/CocoapodsSample/Pods"
+    - "CheckoutTests"
+    - "Tests"
+    - "iOS Example Frame SPM/iOS Example Frame Regression Tests"
+    - "iOS Example Frame/iOS Example FrameUITests"
```
okhan-okbay-cko, line 13: These are added for local checks only. Even if they are not updated, the behaviour of showing the linter errors on GitHub UI would be the same since we only get the diff from the HEAD branch and the main branch and run the lint on those files

PR #470, opened 2023-09-08
Update README in favour of Wiki

okhan-okbay-cko:
This will make everything tidier and way much easier to find! 

Benefits:
- Developers won't land on a colossal page with all kind of irrelevant info and they won't get quickly overwhelmed by it
- Developers will easily find whatever it's they are looking for
- PRs to update README will not be needed anymore and we'll have quick updates whenever we need.
- Whenever there's a new version, we'll only need to update 1 or 2 pages, for example, [Integration](https://github.com/checkout/frames-ios/wiki/Integration) if we have a change on the integration methods or if we just released the border work then we only need to change [Make It Your Own](https://github.com/checkout/frames-ios/wiki/Make-It-Your-Own)
- Easy to browse through it on the sidebar. Rather than an endless scrolling on a huge README file.

Cons: 
- We won't have a frozen version of our README that's bound to our released versions and we'll need to be strict about naming our pages with the relevant Frames version number? There is also history available: https://github.com/checkout/frames-ios/wiki/Integration/_history

As a solution to this, we can keep the `Documentation` section under README. And then, we can have different pages that has the version number as a suffix like `Integration [4.2.0]` and link it. They will stay in the revision history of the Wiki pages and can be reviewed later. This is for a case where a merhant uses an old version and at a later time they come and want to see the old documentation. By doing this, we create a `frozen README` for every version out there.

**Investigation:**

Wiki pages are being created by the title string. Which means the path to them includes the string in the title. So, even if we see the history, when we change the title, the old version returns 404.

If we can’t solve this, we might hold separate .md  files for separate sections of README.md and link them under the Documentation section while the main README again just holding the Documentation section.

or, we can keep the same title, being careful about it, and then write commit messages as Update for 4.2.1 like this https://github.com/adam-p/markdown-here/wiki/Tips-and-Tricks/_history. even if you accidentally change the title and create a new file, when you change it back to the old name, it stands still there with all the revision history


okhan-okbay-cko:
Closing this because:
- Wiki don't get versioned under git
- Wiki can lose history if the page title is changed
- Wiki gets updates directly without reviewers and this makes it risky to maintain

PR #483, opened 2023-09-20
Update the PR template

okhan-okbay-cko:
Use new PR template. And here is an autolink: JIRA-2160


jheng-hao-lin-cko (approved):
Nice! that's a lot's check for a changes on `md` file

okhan-okbay-cko:
yeah good shout. I think we need to do a `git diff` before running some workflows @jheng-hao-lin-cko

PR #512, opened 2024-02-07
Skip verify PR flow for README updates

aashna-narula-cko:
Currently, verify PR workflow on Frames runs a variety of checks even if the change is in [README.md](http://readme.md/) which does not affect the code. Add conditional checks in the check-pr yaml script to make an early return if the change is only in [README.md](http://readme.md/) file.

### DIFF
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -36,11 +36,6 @@ jobs:
       - name: Analyze if only Readme.md is changed
         run: |
           CHANGED_FILES="${{ steps.get_files_changed.outputs.files_changed }}"
-          if [ -z "$CHANGED_FILES" ]; then
-            echo "No changed files found."
-            echo "::set-output name=files_changed_without_readme::true"
-            exit 0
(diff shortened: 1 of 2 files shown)


.github/workflows/check-pr.yml
```diff
@@ -20,13 +20,43 @@ concurrency:
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Determine Changed Files
+        id: get_files_changed
+        run: echo "::set-output name=files_changed::${{ toJson(github.event.pull_request.files.*.filename) }}"
+
+      - name: Analyze if only Readme.md is changed
```
okhan-okbay-cko, line 35: can we extend this to all `.md` suffixed files?

aashna-narula-cko:
This PR seems to do the needful but will look at the

okhan-okbay-cko:
When we did the trade-off of implementation effort of this workflow vs the added value that'll come from it, we decided to drop it for now.

PR #517, opened 2024-02-16
Analyze CodeQL scanned files

aashna-narula-cko:


.github/workflows/codeql-analysis.yml
```diff
@@ -54,6 +54,16 @@ jobs:
...
 
+    - name: Build iOS Example Frame
+      run: |
+        cd iOS\ Example\ Frame\ SPM
+        xcodebuild build -scheme  iOS\ Example\ Frame -destination "platform=iOS Simulator,name=iPhone 14 Pro,OS=latest"
+
+    - name: Build UITest
+      run: |
+        cd iOS\ Example\ Frame\ SPM
+        xcodebuild -scheme UITest -destination "platform=iOS Simulator,name=iPhone 14 Pro,OS=latest" test
```
okhan-okbay-cko, line 65: This will probably run all the UITests and will take 10 minutes. We might consider not including it maybe? or, we can add a build capability to the scheme

aashna-narula-cko: Umm, yes, it is taking a lot of time to build. Better to remove it?

okhan-okbay-cko: can we just compile that target instead?

aashna-narula-cko: Can you help with an example?

okhan-okbay-cko: I thought of selecting the `UITest` scheme on Xcode, then ticking the `Run` checkbox for the UITest scheme. After that, it should become a target that's runnable as well as testable. So that we can remove the `test` wording from the end inside the `yml` file

okhan-okbay-cko: actually, even without doing anything, just removing the `test` should do the trick

okhan-okbay-cko:
Shall we add CocoaPods example? They may be using the same files to some extent though.

.github/workflows/codeql-analysis.yml
aashna-narula-cko: @okhan-okbay-cko I had added Regression Tests but they are taking more than 15 mins to build alone. So, removed them. Do you suggest to add them?

okhan-okbay-cko: that's a bit unexpected. Which `xcodebuild` command was used please?

aashna-narula-cko: I remember using:
xcodebuild -scheme Regression\ Tests -destination "platform=iOS Simulator,name=iPhone 14 Pro,OS=latest" test

okhan-okbay-cko: I believe we can remove `test` command from the end, after making that scheme runnable. Then it shouldn't take that long

aashna-narula-cko: Without test command, I am getting below error, Do you have any workaround for this?

Screenshot 2024-02-19 at 11 43 47 am

aashna-narula-cko:
Created a JIRA ticket to include remaining files to be scanned by CodeQL.

okhan-okbay-cko (approved):

PR #550, opened 2025-05-27
Update SPM dependencies

ghost:
## Issue
[PIMOB-3566
](https://checkout.atlassian.net/browse/PIMOB-3566)
## Proposed changes

- Upgrade PhoneNumberKit to "version" : "4.0.2"
- Update Risk SDK min version  "3.0.2"
- Update min iOS version to be iOS 13.0
- Update GHA to macos15, xcode 16 and iphone 16 pro
- Support secure to use  uses: actions/checkout@85e6279cec87321a52edac9c87bce653a07cf6c2` instead of `@v3 `or `@v4`
- added [this](https://github.com/checkout/frames-ios/pull/549) solution because it's failing on MainActor


ali-farhadi-cko (approved):

Package.swift
```diff
@@ -17,10 +17,10 @@ let package = Package(
     dependencies: [
         .package(
             url: "https://github.com/marmelroy/PhoneNumberKit.git",
-            exact: "3.5.9"),
+            exact: "4.0.2"),
```
okhan-okbay-cko: would it be better to have this `from` but not `exact?`

iOS Example Frame SPM/iOS Example Frame SPM.xcodeproj/project.pbxproj
```diff
@@ -531,7 +533,8 @@
 			name = "iOS Example Frame";
 			packageProductDependencies = (
 				1629C08B2AF90379001BD3D9 /* Frames */,
-				16BA563E2B8FA6CE000775F9 /* Frames */,
+				958E51042DE74B3E0003F4B0 /* Frames */,
+				950F88602DE74DAB0041D2E7 /* Frames */,
```
okhan-okbay-cko: I believe this is linked twice unintentionally

Package.swift
ghost: They are breaking API changes on major releases so it's more secured to keep it 4.0.2.

okhan-okbay-cko: won't we be prone to have the same error in the future? The merchants will continue to have dependency conflicts when they use `PhoneNumberKit` separate than our SDK. 

when we do `from: "4.0.2"`, it'll go up to anything but not beyond `4.x.x` so it'll never be automatically `5.x.x`. I believe we should be safe in that case?

iOS Example Frame/Podfile
```diff
@@ -1,12 +1,13 @@
...
-platform :ios, '12.0'
+platform :ios, '13.0'
 
 target 'iOS Example Frame' do
   # Comment the next line if you don't want to use dynamic frameworks
   use_frameworks!
 
   # Pods for iOS Example Custom
-  pod 'Frames', '4.3.7'
+  pod 'PhoneNumberKit', '~> 4.0'
```
okhan-okbay-cko: How did it work before without this?

ghost: I have just pointed to current branch to run all UI tests on this branch. I will change it back in another PR to new release branch 

```swift
pod 'Frames', :git => 'https://github.com/checkout/frames-ios.git', :branch => 'update/spm_dependacies'
```

okhan-okbay-cko: my comment was rather pointing to the addition of `PhoneNumberKit` dependency. How come we didn't have it as a dependency here but everything still worked? Or, why did we need it now?

ghost: I was testing the issue for merchant about min version for `PhoneNumberKit`. I have removed it. Previously, and Now, `PhoneNumberKit` is picked up from Frames pod spec file

okhan-okbay-cko (dismissed):

.github/workflows/validate-for-app-store.yml
```diff
@@ -1,8 +1,8 @@
 name: Validate For App Store
 
-on:
-  pull_request:
-    types: [ labeled ]
+#on:
+#  pull_request:
+#    types: [ labeled ]
```
ghost: Error on using certificate. We don]t have admin access to update it

okhan-okbay-cko (approved):

